PR #77, opened 2019-07-24
76 load simulation

msevestre:


msevestre:
@PavelBal @Yuri05 SLowly but surely, we need to start

This is a WIP: I do not know how best to organize and copy dlls. At the moment, they are under lib. This probably will not be how we should go about it


My goal is to get something working and start the discussion

msevestre:
Question nr1: This package requries rClr to run.  
See https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r#L231

How can I install rCLR ? It is required to run the tests

msevestre:
@PavelBal @Yuri05 Success!
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r/build/artifacts

@PavelBal Can you try to installt the package using a the zip file (provided you have installed also rClr) 
and run the following code

```R
library(ospsuite)

ospsuite::initPackage()
```

Does that work?

msevestre:
@PavelBal I should add: Try first with the zip file (contains binary file as opposed to source code) and then with the tar file (which contains the R code)..
Which makes me realize that the way we created the MoB R package before was probably not ideal :)

PavelBal:
- Installing` .zip` under Windows works, running the code snippet too.

```
OSPSuite R API Initializing...
OSPSuite R API Initialized.
NULL
```

- Installing `.zip `under Linux fails 
`Error in install.packages : cannot install Windows binary packages on this platform`

Installing `.tar` under Windows works, but with warnings of type
```
Warning in file.copy(instdir, dirname(final_instdir), recursive = TRUE,  :
  problem copying D:\Documents\R\win-library\3.6\00LOCK-ospsuite\00new\ospsuite\lib\Castle.Core.dll to D:\Documents\R\win-library\3.6\ospsuite\lib\Castle.Core.dll: Permission denied
```

After installation, the directory `D:\Documents\R\win-library\3.6\00LOCK-ospsuite\00new\ospsuite\lib\` does not exist.

```
ospsuite::initPackage()
NULL
```
**Installing from .tar produces different output than from .zip!**

- Installing `.tar` in Linux works, but 

```
> ospsuite::initPackage()
Assembly '/home/pavel/R/x86_64-pc-linux-gnu-library/3.5/rClr/libs/ClrFacade.dll' doesn't have an entry point.
OSPSuite R API Initializing...
Error in rClr::clrCallStatic("OSPSuite.R.Api", "InitializeOnce", apiConfig$ref) : 
  Type:    System.ArgumentException
Message: The 'ExeConfigFilename' argument cannot be null.
Method:  Void CheckFileMap(System.Configuration.ConfigurationUserLevel, System.Configuration.ExeConfigurationFileMap)
Stack trace:
  at System.Configuration.ExeConfigurationHost.CheckFileMap (System.Configuration.ConfigurationUserLevel level, System.Configuration.ExeConfigurationFileMap map) [0x0001b] in <36be543d7b5c457999b9c47aaf95abd1>:0 
  at System.Configuration.ExeConfigurationHost.InitForConfiguration (System.String& locationSubPath, System.String& configPath, System.String& locationConfigPath, System.Configuration.Internal.IInternalConfigRoot root, System.Object[] hostInitConfigurationParams) [0x00030] in <36be543d7b5c457999b9c47aaf95abd1>:0 
  at System.Configuration.InternalConfigurationSystem.InitForConfiguration (System.String& locationConfigPath, System.String& parentConfigPath, System.String& parentLocationConfigPath) [0x00000] in <36be543d7b5c457999b9c47aaf95abd1>:0 
```

So obviously, when installing from .tar, something does not work.

PavelBal:
Can we see the source of `OSPSuite.R.dll`?

msevestre:
It definitely cannot work on Linux. The funcparser is the native for Windows so all Linux stuff can be forgotten for now. 

The source is on GitHub ospsuite.core in the .net standard branch

msevestre:
@PavelBal https://github.com/r-lib/remotes/issues/368#issuecomment-493030597

I think the warning with the tar under windows is because you had a session open somehow using the file and it could not be replaced (access denied). Fair enough

PavelBal:
> @PavelBal [r-lib/remotes#368 (comment)](https://github.com/r-lib/remotes/issues/368#issuecomment-493030597)
> 
> I think the warning with the tar under windows is because you had a session open somehow using the file and it could not be replaced (access denied). Fair enough

Yes, I first installed zip, tried it out, and then installed tar without closing the R session. That was the problem.

msevestre:
The question is still. If you remove the package and install it with TAR on windows, does it work or not?

msevestre:
Ok I am going to move ahead and implement the actual load simulation as we have it now in the proof of concept. Then I'll migrate the getParameters function. From their, we can start working on features

PavelBal:
> The question is still. If you remove the package and install it with TAR on windows, does it work or not?

Yes, it does :)

msevestre:
ok merging this code for now. Lot's of question remaining. 
I have  a unit test running loading a simulation and another one throwing an exception if this is not valid

PR #82, opened 2019-07-25
Fixes #81 implement get parameters

### LINKED ISSUE #81: implement getParameters
As described here https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/19

msevestre:


R/utilities-parameter.R
```diff
@@ -1,3 +1,43 @@
+
+#' Retrieve all parameters of a container (simulation or container instance) matching the given path criteria
```
msevestre: @PavelBal Do you know of any good reference to create Roxygen docs? Somehow I am a bit struggling with the keywords

This is what I have at the mometn
image


If I do not use the don't run key, the examples are being run when the package is built which does not work. Anyways I'll create an issue for that

tests/testthat.R
```diff
@@ -1,4 +1,9 @@
+# TODO. Not sure why this is required
+library(rClr)
```
msevestre: I don't understand WY I need to load the rCLR since it's a dependencies from the package. Any idea?

tests/testthat/test-utilities-parameter.R
```diff
@@ -0,0 +1,33 @@
...
+  expect_equal(length(parameters), 1)
+})
+
+test_that("It can retrieve parameters with generic path path", {
+  parameters <- getParameters(c("Organism", "Liv*", "Intracellu*", "Vol*"), sim)
+  expect_equal(length(parameters), 1)
+})
+
+test_that("It can retrieve all parameters matching a given criteria with wild card entry", {
+  parameters <- getParameters(c("Organism", "Liver", "*", "Volume"), sim)
```
msevestre: The code is really in .NET. Just want to document some behavior that's all

msevestre:
Merging. @PavelBal Great work at adding an issue for each point you see. That's how we are going to make it great :)

PR #87, opened 2019-07-26
Fixes #85 and Fixes #86

### LINKED ISSUE #85: Do not export functions/classes that should not be accessed directly
I suggest that users should not have access to the classes defined in `domain.R`. All manipulations should be performed through calling public functions.

Therefore, remove the `@export`-tag from the documentation of the classes.

### LINKED ISSUE #86: Call initPackage() from loadSimulation and create hidden env
We will probably have to deal with some global entities now and then. As discussed in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/64, I propose to use a hidden environment `.ospsuiteEnv` for this purpose that will be created upon initialization of the package.

Also, I don't like that the user has to explicitly call `initPackage()`. I would call `initPackage()` as part of `loadSimulation` or other potential functions that will be called first. `initPackage` will only be called once:

```
loadSimulation <- function(pkmlSimulationFile) {
  if (!exists(".ospsuiteEnv") || (.ospsuiteEnv$isInitialized == FALSE)){
    .ospsuiteEnv <<- new.env()
    .ospsuiteEnv$isInitialized <- TRUE
    initPackage()
  }
...
```

msevestre:


DESCRIPTION
```diff
@@ -1,15 +1,3 @@
-BinaryFiles:
```
msevestre: I had added that here to try to remove the warnings but this did not do anything

NAMESPACE
```diff
@@ -1,14 +1,7 @@
 # Generated by roxygen2: do not edit by hand
 
-export(ApiConfig)
```
msevestre: Removing all of that for now. Looks cleaner indeed :)

R/utilities-parameter.R
```diff
@@ -30,9 +29,10 @@ getParameters <- function(path, container) {
 #'
 #' @return [Parameter] The parameter with the given path or null if not found
 #' @examples
-#'\dontrun{
-#' getParameter(c("Organism", "Liver", "Volume"), sim))
-#' }
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' param <- getParameter(c("Organism", "Liver", "Volume"), sim)
```
msevestre: Pretty neat that those examples are being run in fact. You are sure that the example works as expected

R/zzz.R
```diff
@@ -0,0 +1,11 @@
+.onLoad <- function(...) {
+
+  # Should only be called once
+  initPackage()
+
+
+  # to do
+#  ggplot_global$theme_current <- theme_gray()
```
msevestre: I just copied that here. We ned to work on env stuff

tests/testthat.R
```diff
@@ -1,9 +1,5 @@
-# TODO. Not sure why this is required
-library(rClr)
```
msevestre: well not required anymore

tests/testthat.R
```diff
@@ -1,9 +1,5 @@
-# TODO. Not sure why this is required
-library(rClr)
 
 library(testthat)
 library(ospsuite)
 
-initPackage()
```
msevestre: and not required anymore :)

R/zzz.R
```diff
@@ -0,0 +1,17 @@
+.onLoad <- function(...) {
+
+  # Only for x64 bits packages. This is required to avoid error when package is being checked on CI for x86
+  is64 <- (.Machine$sizeof.pointer==8)
```
msevestre: @PavelBal I had to actually do that because package creation would fail. 
It tried to install the package on x86 and loading fails obviously. There might be a better solution. I tried to remove the check from R CMD CHECK but I was getting other erros

PR #94, opened 2019-07-27
59 get containers

msevestre:


R/enum.R
```diff
@@ -0,0 +1,35 @@
...
+  enumNames <- names(myEnum)
+  if (is.null(enumNames)) {
+    names(myEnum) <- myEnum
+  } else if ("" %in% enumNames) {
+    stop("The inputList has some but not all names assigned. They must be all assigned or none assigned")
+  }
+  return(myEnum)
+}
+
+ContainerType <- enum(c(
```
msevestre: @PavelBal This is a quick try to implement enum like struct based on some code I found
It works great but I have one issue. The enum I create, such as `ContainerType` need to be in the same file as the enum function otherwise it is not found 

Any idea why?

This probably due to some namespace issue. In short here, I want to declare a global variable to the package called `ContainerType` that everyone can access by doing for example `ContainerType$Organ`

Let's think about it a bit

R/api-config.R
```diff
@@ -0,0 +1,15 @@
+#' @title ApiConfig
+#' @docType class
+#' @description  Global configuration for the OSPSuite .NET API
+ApiConfig <- R6::R6Class(
+  "ApiConfig",
+  inherit = DotNetWrapper,
+  active = list(
+    pkParametersFilePath = function(value) {
+      private$wrapProperties("PKParametersFilePath", value)
```
msevestre: private is a keyword allowing me to create methods in base class that are not public available

R/dot-net-wrapper.R
```diff
@@ -0,0 +1,37 @@
...
+#' @importFrom R6 R6Class
+DotNetWrapper <- R6::R6Class(
+  "DotNetWrapper",
+  public = list(
+    ref = NULL,
+    initialize = function(ref) {
+      self$ref <- ref
+    }
+  ),
+  private = list(
```
msevestre: I have defined them private in the DotNetWrapper class

R/entity.R
```diff
@@ -0,0 +1,37 @@
...
+#' @docType class
+#' @description  Abstract wrapper for an OSPSuite.Core Entity class
+#'
+#' @field path The path of the entity in the container hiearchy. (read-only)
+Entity <- R6Class("Entity",
+  inherit = ObjectBase,
+  active = list(
+    path = function(value) {
+      if (missing(value)) {
+        rClr::clrCallStatic(typename = "OSPSuite.Core.Domain.EntityExtensions", methodName = "EntityPath", self$ref)
```
msevestre: @PavelBal @Yuri05 THis is how one can call an extension methods BTW. They are nothing else than static properties on a static class

R/enum.R
```diff
@@ -0,0 +1,25 @@
...
+#' @examples
+#'
+#' # Without predefined values
+#' Color <- enum(c("Red", "Blue", "Green"))
+#' myColor <- Color$Red
+#'
+#' # With predefined values
+#' Symbol <- enum(c(Diamond=1, Triangle=2, Circle=2))
+#' mySymbole <- Symbol$Diamond
+enum <- function(enumValues) {
```
msevestre: We will need to experiment with this but this is working quite nicely so far .

R/utilities-container.R
```diff
@@ -0,0 +1,50 @@
...
+  containers <- getContainers(path, container)
+  stopifnot(length(containers) <= 1)
+  if (length(containers) == 0) {
+    return(NULL)
+  }
+
+  containers[[1]]
+}
+
+toContainers <- function(netContainers) {
```
msevestre: @PavelBal Similar to the `toParameters` method, we should decide if we want to optimize here or not.

R/zzz.R
```diff
@@ -1,17 +1,18 @@
...
   }
-#
+  #
   # Should only be called once
   initPackage()
 
+  ContainerType <- containerType();
 
+  # Ensures that the variable ContainerType is available in the package as ospsuite::ContainerType
+  assign("ContainerType", ContainerType, envir = parent.env(environment()))
```
msevestre: For some reason, doing ContainerType <<- containerType does not allow to retrieve the variable outside of package using ospsuite::ContainerType

Like this, it works

@PavelBal @Yuri05 Not sure if this is the best way to define enum but this works nicely and from a caller, you can have something like

image

R/zzz.R
```diff
@@ -1,17 +1,18 @@
...
   # to do
-#  ggplot_global$theme_current <- theme_gray()
-
-
+  #  ggplot_global$theme_current <- theme_gray()
```
msevestre: This is more of a note for later :)

R/zzz.R
PavelBal: `ContainerType <<- containerType`

creates `ContainerType` in the `Global Environment`. The way you did it ensures that it is created in the package environment - good!

tests/testthat/test-container.R
```diff
@@ -0,0 +1,39 @@
+context("Container")
+
+dataPath <- file.path(getwd(), "..", "data", fsep = .Platform$file.sep)
+simFile <- file.path(dataPath, "S1.pkml", fsep = .Platform$file.sep)
+sim <- loadSimulation(simFile)
+liverPath <- c("Organism", "Liver")
+
+test_that("It can update the name of a container", {
```
PavelBal: Retrieve, not update?

tests/testthat/test-container.R
```diff
@@ -0,0 +1,39 @@
...
+liverPath <- c("Organism", "Liver")
+
+test_that("It can update the name of a container", {
+  container <- getContainer(liverPath, sim)
+  expect_equal(container$name, "Liver")
+})
+
+test_that("It can retrieve the id of a container", {
+  container <- getContainer(liverPath, sim)
+  expect_true(container$id != "")
```
PavelBal: Should check for `NULL`

```
> container$id
[1] "-z8KDLQX30SwvXuqREQ0-w"
> container$ids
NULL
```

tests/testthat/test-container.R
```diff
@@ -0,0 +1,39 @@
...
+test_that("It throwns an error when trying to set the path of a container", {
+  container <- getContainer(liverPath, sim)
+  expect_that(container$path <- "TOTO", throws_error())
+})
+
+
+test_that("It throwns an error when trying to set the id of a container", {
+  container <- getContainer(liverPath, sim)
+  expect_that(container$id <- "id", throws_error())
+})
```
PavelBal: Also add check for ttrying to change container name (error thrown).

tests/testthat/test-utilities-container.R
```diff
@@ -0,0 +1,39 @@
...
+  containers <- getContainers(c("Organism", "*", "Intracellu*"), sim)
+  expect_equal(length(containers), 15)
+})
+
+test_that("It can retrieve all containers matching a given criteria with wild card entry", {
+  containers <- getContainers(c("Organism", "Liver", "Pericentral", "*"), sim)
+  expect_equal(length(containers), 5) # 4 sub compartments in liver pericentral + drug
+})
+
+test_that("It can retrieve all parameters matching a given criteria with generic path entry", {
```
PavelBal: Parameters = containers.

tests/testthat/test-utilities-container.R
```diff
@@ -0,0 +1,39 @@
...
+
+test_that("It can retrieve a single container by path if it exists", {
+  container <- getContainer(c("Organism", "Liver", "Intracellular"), sim)
+  expect_equal(container$name, "Intracellular")
+})
+
+test_that("It returns null if the  container by path does not exist", {
+  container <- getContainer(c("Organism", "Liver", "TOTO", "Length"), sim)
+  expect_null(container)
+})
```
PavelBal: Test case for `getContainer()` called for path  that returns multiple containers.

R/enum.R
msevestre: This is not without names, this is without values
I could create enum(c(Diamond, Triangle, Circle)) and this would work

R/container.R
```diff
@@ -0,0 +1,17 @@
...
+#' \describe{
+#'   \item{containerType}{Type of container}
+#'   }
+#'
+Container <- R6Class("Container",
+  inherit = Entity,
+  active = list(
+    containerType = function(value) {
+      type <- private$wrapReadOnlyProperties("ContainerType", value)
+      ContainerType[[which(ContainerType == type)]]
```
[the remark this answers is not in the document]
msevestre: Donet enum are int underneath the hood. If we want to use strings, we'll have to cast back and forth. While I agree that this is not ideal to duplicate the stuff, it is probably the easiest for now.
Even if we where to be able to use the string, they would have to match exactly. While string may change, value will not. If we want to use an enum, I do not know of a better way to do it

tests/testthat/test-container.R
msevestre: yep

R/container.R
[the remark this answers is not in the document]
msevestre: in the code was nonesense indeed. I was hoping to return an actual entry in the enum but this is nothng more than a number indeed

tests/testthat/test-container.R
msevestre: done

msevestre:
All comments addressed. I'll merged once PR is ok

PR #107, opened 2019-08-01
Fixes #104 cache container task. Also fixes #101

### LINKED ISSUE #101: Use cat instead of print in the $print() method of objects
If using `print()`, the returned value is a list:

```
  public = list(
    print = function(...) {
      print(paste0("Parameter '", self$path, "' has a value of ", self$value))
      invisible(self)
    }
  )
```

```
> param
[[1]]
[1] "Parameter 'S1|Organism|Weight' has a value of 73.0000002396407"

```

Instead, using `cat` produces imo a more tidy output:

```
  public = list(
    print = function(...) {
      cat(paste0("Parameter '", self$path, "' has a value of ", self$value))
      invisible(self)
    }
  )
```

```
> param
[1] "Parameter 'S1|Organism|Weight' has a value of 73.0000002396407"

```

### LINKED ISSUE #104: Create  containerTask in init.package()
To reduce overhead of creating new instance of `ContainerTask`.

msevestre:


R/init-package.R
```diff
@@ -16,4 +16,7 @@ initPackage <- function() {
   apiConfig$dimensionFilePath <- filePathFor("OSPSuite.Dimensions.xml")
 
   rClr::clrCallStatic("OSPSuite.R.Api", "InitializeOnce", apiConfig$ref)
+
+  # initialize global variables (mostly usef for performance optimization)
+  ospsuiteEnv$containerTask <- rClr::clrCallStatic("OSPSuite.R.Api", "GetContainerTask")
```
msevestre: `ospsuiteEnv` is our global environment for global stuff. it won't be exposed by the package.
Does this sound ok to you?

R/quantity.R
```diff
@@ -0,0 +1,34 @@
...
+#'
+#' @field value The value of the quantity
+Quantity <- R6Class(
+  "Quantity",
+  inherit = Entity,
+  active = list(
+    value = function(value) {
+      private$wrapProperties("Value", value)
+    },
+    displayUnit = function(value) {
```
msevestre: Big need to add documentation to this but I want to discuss the API before continuing
A quantity in OSPSUite.Core (molecule, amount, parameter etc..) has  a value, dimension and baseunit
Then it also has a display unit and a value in display unit

I have exposed all those methods . Maybe we should only work with baseunit an value in base unit. But it's not as nice for display

R/quantity.R
```diff
@@ -0,0 +1,34 @@
...
+    },
+    valueInDisplayUnit = function(value) {
+      private$wrapProperties("ValueInDisplayUnit", value)
+    }
+  ),
+  private = list(
+    printQuantity = function() {
+      private$printClass()
+      private$printLine("Path", self$path)
+      private$printLine("Value", paste0(format(self$value, digits = 5, nsmall = 2), " [", self$unit, "]"))
```
msevestre: In theory this could be display unit

I am not sure about digigts etc.. Just wanted to have something out there.
Probably we need a format double method somewhere

R/utilities-container.R
```diff
@@ -17,9 +17,7 @@
 #' containers <- getAllContainersMatching(c("Organism", "**", "Intracellular"), sim)
 #' @export
 getAllContainersMatching <- function(path, container) {
-  containerTask <- rClr::clrCallStatic("OSPSuite.R.Api", "GetContainerTask")
-
-  toContainers(rClr::clrCall(containerTask, "AllContainersMatching", container$ref, path))
+  toContainers(rClr::clrCall(ospsuiteEnv$containerTask, "AllContainersMatching", container$ref, path))
```
msevestre: using the container task

R/quantity.R
msevestre: this is how it looks now
```
Parameter : 
  Path: S1|Organism|Liver|Volume
  Value: 2.3765 [l]
```

appveyor.yml
```diff
@@ -34,7 +34,7 @@ build_script:
 
 
 on_success:
-  - Rscript -e "covr::codecov()"
+  - Rscript -e "covr::codecov(function_exclusions = c('print\\.', '\\.onLoad'))"
```
msevestre: Remove those function from coverage. Does not make sense as we cannot test them (and onLoad is called no matter what)

tests/testthat/test-parameter.R
```diff
@@ -37,9 +38,29 @@ test_that("It throwns an error when trying to set the id of a parameter", {
...
+})
+
+test_that("It can retrieve the unit of a parameter", {
+  par <- getParameter(volumePath, sim)
+  expect_equal(par$unit, "l")
+})
+
+test_that("It can retrieve the display unit of a parameter", {
+  par <- getParameter(heightPath, sim)
+  expect_equal(par$displayUnit, "cm")
```
msevestre: example of different base unit and display unit

tests/testthat/test-utilities-container.R
```diff
@@ -21,9 +21,9 @@ test_that("It can retrieve all containers matching a given criteria with wild ca
   expect_equal(length(containers), 5) # 4 sub compartments in liver pericentral + drug
 })
 
-test_that("It can retrieve all parameters matching a given criteria with generic path entry", {
+test_that("It can retrieve all containers matching a given criteria with generic path entry", {
```
msevestre: Fixed a bug in core which resulted in the count being different

R/init-package.R
PavelBal: Sounds good.

R/ospsuite-env.R
```diff
@@ -0,0 +1,3 @@
+# Environment that holds various global variables and settings for the ospsuite,
+# It is not exported and should not be directly manipulated by other packages.
+ospsuiteEnv <- new.env(parent = emptyenv())
```
PavelBal: Why having a separate file, and not just initialize in `init.package.R`?

R/quantity.R
PavelBal: Wow, this greatly confuses me! I would only have the `unit` as base unit and `value` in base unit.

We could add a function to convert to base units (e.g. `convertToBaseUnit(dimension, unit, value)`)

PavelBal: And also `convertFromBaseUnit(dimension, targetUnit, valueInBaseUnit)`

R/quantity.R
PavelBal: I like this output.

R/ospsuite-env.R
msevestre: This file can document all the stuff we are going to put in there.  This can become quite long and that's why I followed ggplot structure and created a separate file

PavelBal: OK.

R/quantity.R
msevestre: I have removed the methods. They make sense if you think in OOP however

PR #115, opened 2019-08-03
114 add simulation runner infrastructure

msevestre:


R/init-package.R
```diff
@@ -18,5 +18,5 @@ initPackage <- function() {
   rClr::clrCallStatic("OSPSuite.R.Api", "InitializeOnce", apiConfig$ref)
 
   # initialize global variables (mostly usef for performance optimization)
-  ospsuiteEnv$containerTask <- rClr::clrCallStatic("OSPSuite.R.Api", "GetContainerTask")
```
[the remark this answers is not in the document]
msevestre: @PavelBal I just realized looking at the build that this call was not allowed here at all. Tests are running etc.. but package creation produces warning. I had to move that out of this method. not sure why (error with rClr). Maybe you can check it out if you want. 
see https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r/builds/26442216#L335

R/enum.R
```diff
@@ -1,18 +1,19 @@
...
 #'
 #' # Without predefined values
 #' Color <- enum(c("Red", "Blue", "Green"))
 #' myColor <- Color$Red
 #'
 #' # With predefined values
-#' Symbol <- enum(c(Diamond=1, Triangle=2, Circle=2))
+#' Symbol <- enum(c(Diamond = 1, Triangle = 2, Circle = 2))
 #' mySymbol <- Symbol$Diamond
+#' @export
```
msevestre: Require to export a function to be able to run the examples. So many fucking rules

R/get-net-task.R
```diff
@@ -0,0 +1,18 @@
...
+#'
+#' @return An instance of the Task
+#'
+#' @details
+#' simulationLoader <- getNetTask("SimulationLoader")
+getNetTask <-function(taskName) {
+  rClr::clrCallStatic("OSPSuite.R.Api", paste0("Get", taskName))
+}
+
+getContainerTask <- function() {
```
msevestre: This is where I am doing the cash. As a caller, we have to know if we need to cache stuff or not
I would argue this should not be done too often. For example, SimulationLoader. We need one new instance for each to ensure that all dependencies are cleanup properly. Too much optimziation can lead to brutal bug.

I am still not sure to be honest that we need to save a containerTask reference but we have the concept now so we can play with it

R/load-simulation.R
```diff
@@ -4,7 +4,7 @@
 #'
 #' @export
 loadSimulation <- function(pkmlSimulationFile) {
-  simulationLoader <- rClr::clrNew("OSPSuite.R.Services.SimulationLoader")
+  simulationLoader <- getNetTask("SimulationLoader")
```
PavelBal: Just out of curiosity - does it have to be a separate SimulationLoader per simulation?

R/quantity.R
```diff
@@ -1,8 +1,11 @@
 #' @title Quantity
 #' @docType class
 #' @description  A quantity of the model (with unit, value) such as a Parameter or an Amount
-#'
-#' @field value The value of the quantity
+#' @field value The value of the quantity in unit
+#' @field unit The base unit in which the quantity is defined (Read-Only)
+#' @field dimension The dimension in which the quantity is defined  (Read-Only)
+#' @field persitable A boolean indicating whether a value will be defined in the simulated results for this quantity
```
PavelBal: This is confusing for me.... Maybe

"A boolean indicating whether the ouput for this quantity will be available in simulated results"?

R/load-simulation.R
msevestre: It does not. Maintaining instances of services in memory may lead to unwanted leaks if said service is not implemented properly. In general, stateless objects can be discarded right away. No need to keep them in memory

R/quantity.R
msevestre: sure. Updating

R/load-simulation.R
PavelBal: So maybe
1) Ether cash it like `containerTask`
2) or destroy it after loading the simulation `rm(simulationLoader)`?

msevestre: I would not cache it as it may lead to leaks. This is not often that such an object is created. We should only cache stuff when really really necessary

2. I am assuming this is destroyed automatically (e.g garbage collected) when the `loadSimulation` is completed no? What does rm do ? 
I certainly hope that local variable are destroyed when the function exit otherwise we have  bigger problems :)

PavelBal: Yes, of course, got somehow confused... Yes it should get destroyed after existing the function.

PR #120, opened 2019-08-12
80 implement set parameters

PavelBal:


PavelBal:
And again:

Why does the build fail with
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r/builds/26636787#L362


```
> #' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
363> sim <- loadSimulation(simPath)
364Error in rClr::clrCall(simulationLoader, "LoadSimulation", pkmlSimulationFile) : 
365  object 'simPath' not found
```

though I've just copied the path from the examlpe of the functions above?

R/utilities-parameter.R
```diff
@@ -46,3 +63,66 @@ toParameters <- function(netParams) {
...
+#' @param parameters A single object of type 'Parameter' or a list of such objects
+#' The objects are created by calling 'getParameter' or 'getAllParametersMatching'
+#'
+#' @param values A numeric value that should be assigned to the parameter or a vector
+#' of numeric values, if the value of more than one parameter should be changed. Must have the same
+#' length as 'parameters'
+#'
+#' @examples
+#'
+#' #' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
```
msevestre: two comments here!!

msevestre:
TWO COMMENTS

#' #' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")

should be
#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")

R/utilities-parameter.R
```diff
@@ -46,3 +63,66 @@ toParameters <- function(netParams) {
...
+#' setParametersValues(param, 1)
+#' params <- getAllParametersMatching(c("Organism", "Liver", "*", "Volume"), sim)
+#' setParametersValues(params, c(1:6))
+#'
+#' @export
+setParametersValues <- function(parameters, values){
+  singleParameter <- FALSE
+
+  #Test if 'parameter' is an instance of the Parameter-class.
+  if (inherits(parameters, "Parameter") && inherits(parameters, "ObjectBase")){
```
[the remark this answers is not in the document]
PavelBal: Yes, my initial thought was to double-check as "Parameter" might be also some class other than ours (I could image other packages use such classes), but I think it's OK to just check for Parameter or Simulation...

R/utilities-parameter.R
```diff
@@ -46,3 +63,66 @@ toParameters <- function(netParams) {
...
+    if (!all(
+      sapply(parameters,
+             fun <- function(x){c(inherits(x, "Parameter"), inherits(x, "ObjectBase"))}))){
+      stop(paste0("setParametersValues: argument 'parameters' is not valid!
+                  Use 'getParameter()' or 'getAllParametersMatching()' to create parameter objects."))
+    }
+
+    #Test if parameters and values are of same length
+    if (length(parameters) != length(values)){
+      stop(paste0("setParametersValues: 'parameters' and 'Values' must be of the same length!"))
```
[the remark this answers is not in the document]
PavelBal: Yes, I wanted to approach this in the next step. Also collect all checks for the correct type somewhere.

R/utilities-parameter.R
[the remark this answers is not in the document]
msevestre: That's ok for me if people are using objects that are not part of the package and the app crashes. 

If we try to check for that, 95% of our code will be to prevent those types of mistakes which is just a waste of time on our part. Let's assume people are using the lib in an expected way.

R/utilities-parameter.R
```diff
@@ -46,3 +63,66 @@ toParameters <- function(netParams) {
...
+  }
+
+  #Test if all values are numeric
+  if (!is.numeric(values)){
+    stop(paste0("setParametersValues: argument 'values' is not valid!
+                Must be a vector of numeric entries"))
+  }
+
+  if (singleParameter){
+    parameters$value <- values
```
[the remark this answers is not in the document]
msevestre: yes. In fact you can simpy do something like `parameters <- c(parameters) and value <-c(values)`
It will convert to an array of one if there is only one element after that. You are good to go and you do not need to check if there is one or more values. The code is the same no matter what

R/utilities-parameter.R
PavelBal: Fixed.

R/utilities-parameter.R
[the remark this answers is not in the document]
PavelBal: Only check for the higher class now.

R/utilities-parameter.R
```diff
@@ -46,3 +63,66 @@ toParameters <- function(netParams) {
...
+    #Test if parameters and values are of same length
+    if (length(values) != 1){
+      stop(paste0("setParametersValues: 'parameters' and 'Values' must be of the same length!"))
+    }
+  }
+  else{
+    #Test if all parameters are valid R6 objects
+    if (!all(
+      sapply(parameters,
+             fun <- function(x){c(inherits(x, "Parameter"), inherits(x, "ObjectBase"))}))){
```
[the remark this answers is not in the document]
PavelBal: All checks are implemented in error-checks.R

R/error-checks.R
```diff
@@ -0,0 +1,27 @@
+#' Check if the provided object is of certain type
```
PavelBal: We can put all the checks that can lead to errors here.

R/messages.R
```diff
@@ -0,0 +1,62 @@
+errorWrongType <- function(object, optionalMessage = NULL) {
```
PavelBal: We can put all message strings here. We can generate message strings dynamically in functions and then call these functions e.g. when throwing an error.

R/messages.R
```diff
@@ -0,0 +1,62 @@
...
+  argnames <- sys.call()
+  arguments <- paste(lapply(argnames[-1], as.character), collapse = ", ")
+
+  message <- paste0(callingFunction, ": Arguments '", arguments,
+                    "' must have the same length, but they don't!", optionalMessage)
+
+  return(message)
+}
+
+getContainerPath <- function(container) {
```
PavelBal: Because `simulation` can also be a container, but does not have a `path`, we have to differentiate here.

R/utilities-container.R
```diff
@@ -17,13 +18,20 @@
...
+  }
+
   toContainers(rClr::clrCall(getContainerTask(), "AllContainersMatching", container$ref, path))
 }
 
 #' Retrieve a single container by path under the given container
 #'
-#' @param path Path of the container
-#' @param container The parent container used to find the container
+#' @inherit getAllContainersMatching
```
PavelBal: We can inherit the documentation and just change the parts that are specific in this function.

R/utilities-container.R
```diff
@@ -34,7 +42,10 @@ getAllContainersMatching <- function(path, container) {
 #' @export
 getContainer <- function(path, container) {
   containers <- getAllContainersMatching(path, container)
-  stopifnot(length(containers) <= 1)
+  if (length(containers) > 1) {
```
PavelBal: I prefer a proper error message here.

R/utilities-parameter.R
```diff
@@ -34,15 +42,58 @@ getAllParametersMatching <- function(path, container) {
...
+#' sim <- loadSimulation(simPath)
+#' param <- getParameter(c("Organism", "Liver", "Volume"), sim)
+#' setParametersValues(param, 1)
+#' params <- getAllParametersMatching(c("Organism", "**", "Volume"), sim)
+#' setParametersValues(params, c(2, 3))
+#' @export
+setParametersValues <- function(parameters, values) {
+  # Must turn the input into a list so we can iterate through even when only
+  # one parameter is passed
+  parameters <- c(parameters)
```
PavelBal: Yepp, works.

R/error-checks.R
```diff
@@ -0,0 +1,27 @@
...
+    sapply(
+      object,
+      function(x) inherits(x, type)
+    )
+  ))
+}
+
+isSameLength <- function(...) {
+  args <- list(...)
+  return(length(
```
msevestre: This indentation is so hard to read. especially with the ==1 at the end. 
I am not sure why the styling allows this.  Can you make it one line?

R/error-checks.R
```diff
@@ -0,0 +1,27 @@
...
+#' @param type String representation of the type that should be checked for
+#'
+#' @return TRUE if the object or all objects inside the list are of the given type.
+#' Only the first level of the given list is considered.
+#'
+#'
+isOfType <- function(object, type) {
+  object <- c(object)
+
+  return(all(
```
msevestre: In many places, you create a local variable and then return the variable. Wouldn't that be easier to read maybe? Or bypass the return completely. For obvious function, this is clear what is being returned

R/messages.R
```diff
@@ -0,0 +1,62 @@
...
+#' Error message when single output is expected but multiple entities produced.
+#'
+#' @param path A vector of strings relative to the \code{container}
+#' @param container A Container or Simulation used to find the entities
+#' @param optionalMessage A string that will be appended to the end of the message
+#'
+#' @return A string representation of the error message
+errorGetEntityMultipleOutputs <- function(path, container, optionalMessage = NULL) {
+  # Name of the calling function
+  callingFunction <- sys.call(-2)[[1]]
```
msevestre: Can you explain where is the -2 etc. is coming from. This will be very useful to understand the code in a few weeks from now

R/messages.R
```diff
@@ -0,0 +1,62 @@
...
+
+  message <- paste0(callingFunction, ": Arguments '", arguments,
+                    "' must have the same length, but they don't!", optionalMessage)
+
+  return(message)
+}
+
+getContainerPath <- function(container) {
+  # Need to check if container is of type Container and Simulation
+  if (isOfType(container, "Simulation")) {
```
msevestre: I think you could create a path method in the Simulation so that you do not have to be so defensive

R/utilities-container.R
msevestre: cool

R/utilities-container.R
msevestre: Agreed

R/utilities-parameter.R
```diff
@@ -34,15 +42,58 @@ getAllParametersMatching <- function(path, container) {
...
+
+  # Test for correct inputs
+  if (!isOfType(parameters, "Parameter")) {
```
msevestre: What about we add some other validation methods 
such as validateIsType(), validateIsNumeric, validateIsSamelLength
That way you only call this methods 

validateIsNumeric would be 
```R
validateIsNumeric <-function(values) {
 if (!is.numeric(values)) {
    stop(errorWrongType(values))
  }
}
```

That way you don't repeat yourself. This something we will often probably
Thoughts?

R/error-checks.R
PavelBal: OK.

R/messages.R
PavelBal: It actually should read 

```
  callingFunction <- sys.call(-length(callingFunctions) + 1)[[1]]
```

The negative number makes to traverse the call stack backwards. +1 was more of a trial-and-error to figure out how to dynamically get the right name.

`callingFunction <- sys.call(-2)[[1]]`
was from stackoverflow and I do not understand completely why -2 so far...

R/messages.R
PavelBal: Agree

R/utilities-parameter.R
PavelBal: Yes this makes sense. However, this will lead to difficulties in generating error messages:
```

  # Name of the calling function
  callingFunction <- sys.call(-length(callingFunctions) + 1)[[1]]
```

Because sometimes the error message function can be called from `validateIs`, and sometimes directly from the function we are interested in, so the the depth of the function we want to retrieve the name from (or the names of the arguments) will vary.

msevestre: oh i see. Ok let's forget about it for now if it's too complicated. 

Maybe there is something to be learned from https://adv-r.hadley.nz/conditions.html#call-stacks

PavelBal: Yes, maybe I'll revisit it later.

PR #127, opened 2019-08-13
123 path as string

msevestre:


R/messages.R
```diff
@@ -1,62 +1,64 @@
...
-}
-
-getContainerPath <- function(container) {
-  # Need to check if container is of type Container and Simulation
-  if (isOfType(container, "Simulation")) {
-    return(container$root$path)
-  }
-  return(container$path)
-}
+errorWrongType <- function(object, optionalMessage = NULL) {
```
msevestre: This files have changed because I ran the style package method from styler.

R/ospsuite-env.R
```diff
@@ -4,4 +4,4 @@ ospsuiteEnv <- new.env(parent = emptyenv())
 
 
 # Reference to container task for optimization purposes only
-ospsuiteEnv$ContainerTask <- NULL
+ospsuiteEnv$containerTask <- NULL
```
msevestre: this was wrong

R/utilities-path.R
```diff
@@ -0,0 +1,32 @@
...
+#'
+#' @param path A string representation of a path
+#'
+#' @return An array containing one element for each path entry
+#' @examples
+#'
+#' array <- toPathArray("Organism|Organ|Liver")
+#' @export
+toPathArray <- function(path) {
+  if (!is.character(path)) {
```
msevestre: @PavelBal As you can see. The same code is repeated over and over again
Havign a one liner validateIsString or something like that would be beneficial

msevestre:
@PavelBal This is ready to merge. Assuming now path as string

PR #132, opened 2019-08-15
Implement $print for Simulation and Container

PavelBal:
- Implement $print for Simulation and Container
- Add formatNumerics()

### DIFF
--- a/R/container.R
+++ b/R/container.R
@@ -9,8 +9,19 @@
 Container <- R6Class("Container",
   inherit = Entity,
   active = list(
-    containerType = function(value) {
+    containerTypeInt = function(value) {
       private$wrapReadOnlyProperties("ContainerType", value)
+    },
+    containerTypeStr = function(value){
+      names(which(ContainerType == self$containerTypeInt))
+    }
+  ),
+  public = list(
+    print = function(...) {
+      private$printClass()
+      private$printLine("Container type", self$containerTypeStr)
+      private$printLine("Path", self$path)
+      invisible(self)
     }
   )
 )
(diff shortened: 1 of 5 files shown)


R/container.R
```diff
@@ -9,8 +9,19 @@
 Container <- R6Class("Container",
   inherit = Entity,
   active = list(
-    containerType = function(value) {
+    containerTypeInt = function(value) {
```
PavelBal: @msevestre Does it make sense for you to have two different function for returning int (e.g. for flow-control in user code) or the string (for readability)?

R/messages.R
```diff
@@ -49,3 +49,7 @@ messages <- list(
 
   errorEnumNotAllNames = "The enumValues has some but not all names assigned. They must be all assigned or none assigned"
 )
+
+formatNumerics <- function(numerics, digits = 5, nsmall = 2) {
```
PavelBal: @msevestre I think this is what you meatn... I put it in the messages.R as it is used for formatting messages, right?

R/container.R
PavelBal: OK, implemented the second one.

R/messages.R
```diff
@@ -49,3 +49,8 @@ messages <- list(
 
   errorEnumNotAllNames = "The enumValues has some but not all names assigned. They must be all assigned or none assigned"
 )
+
+formatNumerics <- function(numerics, digits = ospsuiteEnv$formatNumericsDigits,
```
PavelBal: Defaults are stored in ospsuiteEnv

R/container.R
PavelBal: @msevestre Please review the solution.

Issue #131, opened 2019-08-15
Save path of the pkml file in the simulation


PavelBal:
I think it is a good idea to store the path to the pkml file as a property of the "Simulation" class.

Question - can we get it from .NET object, or should it be implemented as part of the R6 class?

If accepted, the `$print()` method should also return the path to the file.

msevestre:
This is not available in the .net class.
If you think this is required, we can define in the class. This is not required to implement the cache however as the path will be the key of the entry.  O reason to save it in the object

PavelBal:
No it is not required for caching. For me as a user it would be helpful to be able to see where the simulation comes from.

msevestre:
Ok let's put it in the simulation object.
Let's not call in path however;-) otherwise this will be highly confusing ;-)
Suggestion?

Yuri05:
>Suggestion?

`FileSource`. Empty if a simulation created NOT from file, otherwise full path to the pkml

msevestre:
SourceFile maybe?
I don't think we will have the case of a simulation created not from file. Or how can it happen?

msevestre:
this has been fixed in #142

PR #142, opened 2019-08-27
Fixes 111 cache simulation by path

PavelBal:


R/load-simulation.R
```diff
@@ -1,10 +1,52 @@
...
+#' # parameter3 belong to different simulation object than parameter1 and parameter2
+#' parameter3 <- getParameter(toPathString(c("Organism", "Liver", "Volume")), sim3)
+#' setParametersValues(parameters = parameter3, values = 1)
+#' parameter2$value == parameter3$value # FALSE#'
 #' @export
-loadSimulation <- function(pkmlSimulationFile) {
+loadSimulation <- function(pkmlSimulationFile, loadFromCache = TRUE) {
+  validateIsOfType(loadFromCache, "logical")
+
+  if (loadFromCache) {
```
PavelBal: This is where we can cache loaded simulation objects

R/load-simulation.R
```diff
@@ -1,10 +1,52 @@
...
+  # If the simulation has not been loaded so far, or loadFromCache == FALSE,
+  # new simulation object will be created
   simulationLoader <- getNetTask("SimulationLoader")
-  netSim <- clrCall(simulationLoader, "LoadSimulation", pkmlSimulationFile)
-  Simulation$new(netSim)
+  netSim <- rClr::clrCall(simulationLoader, "LoadSimulation", pkmlSimulationFile)
+  simulation <- Simulation$new(netSim)
+
+  # Add the simulation to the cache of loaded simulations
+  ospsuiteEnv$loadedSimulations[[pkmlSimulationFile]] <- simulation
```
PavelBal: It will always overwrite (if existing) the value assigned to a path when a simulation is loaded not from cache.

R/ospsuite-env.R
```diff
@@ -8,3 +8,7 @@ ospsuiteEnv$containerTask <- NULL
 
 # Separator defined in OSPSuite.Core.
 ospsuiteEnv$pathSeparator <- "|"
+
+
+# Cache of the so far loaded simulations. The keys are the paths to the pkml file.
+ospsuiteEnv$loadedSimulations <- new.env(parent = emptyenv())
```
PavelBal: Using an environment for caching.

tests/testthat/test-load-simulation.R
```diff
@@ -2,16 +2,65 @@
...
+  sim2 <- loadSimulation(simFile, loadFromCache = FALSE)
+  sim3 <- loadSimulation(simFile)
+
+  parameter1 <- getParameter(toPathString(c("Organism", "Liver", "Intracellular", "Volume")), sim1)
+  parameter2 <- getParameter(toPathString(c("Organism", "Liver", "Intracellular", "Volume")), sim2)
+  parameter3 <- getParameter(toPathString(c("Organism", "Liver", "Intracellular", "Volume")), sim3)
+
+  setParametersValues(parameters = parameter1, values = 1)
+
+  expect_false(isTRUE(identical(parameter1$value, parameter3$value)))
```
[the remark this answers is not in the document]
PavelBal: Here we expect the values to be no equal = changing the value of parameter1 does not change parameter3.

This is what I found on the internet how to test for *in*equality in testthat. Somehow there is no `expect_not_equal()`

PavelBal:
> @PavelBal Should we implement a clean method of the cache somewhere?

Yes, maybe a general cache-structure (based on environments) with methods like `register` or `get`.
Or what do you mean by "clean"?

Yuri05:
>We have to think about where it makes sense. I have one concern with the implementation/, Simulation are always cached I just realized. I am not sure if this is how it should be. There is no way to create a simulation that is not globally held somewhere. I think this is wrong. Let's not merge this for now until we discuss it with @Yuri05

I think the implementation is ok if we provide removal functions:
* Remove a single simulation from cache
* Clear cache

PavelBal:
```
- Remove a single simulation from cache
- Clear cache
```

What would be a scenario where we have to remove the simulation from cache?

msevestre:
I think this is a leaky abstraction if a user needs to delete a simulation from cache even if it was just used temporarily

Say you want to load sim1 to get the value of a compound lipo and applied it to sim2 with which you work. Because you are a good modeller, retrieval of parameter was made within a function. As such, sim1 was local to the function and the expectation would be that the sim1 would be garbage collected. Well, it won't!! 

Maybe loadSimulation needs another parameter. addToCache or something?

Yuri05:
>Maybe loadSimulation needs another parameter. addToCache or something?

Yes, i think we would need this

msevestre:
> What would be a scenario where we have to remove the simulation from cache?

See above. If we can by default, there needs to be a way to remove. I don't like this API but at least that way, we can clean up 

@PavelBal So let's add another method removeFromCache(sim) that will take the path associated to a simulation and remove the simulation. For that, you will need to add the `fileFullPath` or whatever we decided in #131 so that you can delete the entry if it exists.

Question: If the cached simulation at the given path is not the one passed as parameter, should the cached simulation be removed or not? This is a contrieve scenario but it could happen.

clearCache. Not sure of the use case.

PavelBal:
> Question: If the cached simulation at the given path is not the one passed as parameter, should the cached simulation be removed or not? This is a contrieve scenario but it could happen.

I think in this case it should **not** be removed from cache.

Yuri05:
>clearCache. Not sure of the use case.

Well, e.g. any scenario where something crashes and cache is not properly cleared at the end.

* E.g. running a qualification workflow in interactive session. Simulations are loaded multiple times there e.g. for getting different parameters (molweight etc.) and thus caching absolutely makes sense.
Once the workflow is finished, it probably should automatically remove all cached simulations at the end of the workflow (not via `Cache.Clear` but via removing _only those_ simulations which where _added by the qualification workflow_).
However if the workflow crashes and simulations are not properly removed from cache (which would be a bug in that case, but could happen) - user should clear the cache manually.

* If user loads and caches some simulations within a script and wants to remove them all at the end, `Cache.Clear` is easier to use than looping over all previously loaded simulations

And I guess something like `Cache.Clear` is not very hard to implement ;)

msevestre:
ok makes sense. 
@PavelBal All clear ?  
By default: loadFromCache is true. addToCache should be true I assume?

Also yes: If the ref are not the same, the sim should not be removed from cache

PavelBal:
Yes, will implement.

tests/testthat/test-load-simulation.R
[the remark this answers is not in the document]
PavelBal: Yes, fixed.

PavelBal:
Please review the new cache structure.

R/cache-structure.R
```diff
@@ -0,0 +1,78 @@
+#' @title Cache
+#' @docType class
+#' @description Structure for caching objects
+#'
+#' @field type Type of the object that are cached. All objects must be of the same type.
+#' @field keys Keys of the cached objects
+Cache <- R6::R6Class(
```
PavelBal: In theory we could cache everything using this class.

R/utilities-simulation.R
```diff
@@ -100,3 +145,46 @@ clearOutputs <- function(simulation) {
...
+#' @export
+#'
+#' @examples
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim1 <- loadSimulation(simPath)
+#' sim2 <- loadSimulation(simPath, loadFromCache = FALSE, addToCache = FALSE)
+#'
+#' removeSimulationFromCache(sim1) # returns TRUE
+#' removeSimulationFromCache(sim2) # returns FALSE
+removeSimulationFromCache <- function(simulation) {
```
PavelBal: This removes the given simulation from cache, if it has been cached before.

Issue #148, opened 2019-08-28
Do not use persistable in quantity. Instead use OutputSelection


msevestre:
At the moment, we are exposing the "persitable" property in a quantity to decide whether an entry should be added to the list of output or not. I think this is not the right abstraction anymore. Instead, we should use the same concept as PK-Sim and MoBi (e.g. OutputSettings) to decide which entity should be exported. This will also ensure that the building blocks are consistent when reimported in PKSim

SimulationSettings as the following properties
* Solver
* Output Schema
* Output Selections

Output selection is the one that we need for this entry. But we will need to expose solver and output schema anyways. OutputSelection is a list of QuantitySelection. A quantity selection can be created directly from a quantity. We will have to expose all the methods for those

@Yuri05 @PavelBal Thoughts?

PavelBal:
Sounds good to me. As long as we can select the outputs from R...

msevestre:
Yes. See description above. We'll need to expose method to add/ remove the QuantitySelection

PR #151, opened 2019-08-29
WIP #148 use output selection

msevestre:


msevestre:
@PavelBal @Yuri05 I don't think we actually need the distinction between a parameter, an observer, a MoleculeAmount etc.. everything is in fact a `quantity` in our system. Maybe that what we should use as abstraction

Because I am not sure, I have created a `getAllQuantitesMatching`. Probably the `getAllParametersMatching` might be obsolete. To discuss

msevestre:
Also this is a WIP. I just wanted to get some feedback from you guys asap

R/error-checks.R
```diff
@@ -29,7 +29,11 @@ isSameLength <- function(...) {
 }
 
 
-validateIsOfType <- function(object, type) {
+validateIsOfType <- function(object, type, nullAllowed = FALSE) {
```
msevestre: I think it should not necessarily crash if the object is null. (for example for optional parameters)

R/error-checks.R
```diff
@@ -52,3 +56,7 @@ validateIsSameLength <- function(...) {
 
   stop(messages$errorDifferentLength(arguments))
 }
+
+validateNotAllNull <- function(...){
```
msevestre: A todo here. I am not sure if we need this. I want to make sure that at least one of the arguments is not null

R/output-selections.R
```diff
@@ -0,0 +1,38 @@
+#' @title OutputSelections
```
msevestre: Same hiearchy as in Core. OutputSelection is a prop of SimulationSettings which is defined in a simulation

R/quantity-selection.R
```diff
@@ -0,0 +1,23 @@
+#' @title QuantitySelection
+#' @docType class
+#' @description  List of selected quantities selected as output for a given simulation
+#'
+QuantitySelection <- R6::R6Class(
```
msevestre: This is one quantity selected (path is the key)

R/quantity-type.R
```diff
@@ -0,0 +1,16 @@
+#' All possible quantity types as defined in OSPSuite
+#'
+#' @include enum.R
+#' @export
+QuantityType <- enum(list(
```
msevestre: Another enum to use.

R/utilities-simulation.R
```diff
@@ -43,3 +43,38 @@ runSimulation <- function(simulation) {
...
+#'
+#' @examples
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' outputs <- c("Organism|VenousBlood|Plasma|Caffeine", "Organism|ArterialBlood|**|Caffeine")
+#' addOutputs(paths=outputs, sim)
+#'
+#' @export
+addOutputs <- function(quantities = NULL, paths=NULL, simulation){
```
msevestre: This methods is the one that the user will be using.
I am not suire what's best from the API point of view. Having a way to add quantities instances OR paths (by having two parameters, with at least one required) or to have one parameters that can be of type string array (for path) or quantity array (for instances).

Once we have decided, I'll implement the tests

msevestre: Note: The advantaged of two parameters is that I can theoretically combined instances, and paths together. Will it be used? Probably not

R/utilities.R
```diff
@@ -0,0 +1,4 @@
+toObjectType <- function(netObject, class) {
+  sapply(netObject, function(x)
```
msevestre: Time to abstrat this method as we are copying it everywhere

R/error-checks.R
PavelBal: Can you give an example?

R/quantity-type.R
```diff
@@ -0,0 +1,16 @@
+#' All possible quantity types as defined in OSPSuite
```
PavelBal: Can we maybe name all enum files by the pattern "enum-xxx.R"?. Like `enum-quantity-type.R`.

R/utilities-simulation.R
```diff
@@ -43,3 +43,38 @@ runSimulation <- function(simulation) {
...
+
+  quantities <- uniqueEntities(c(quantities, getAllQuantitiesMatching(paths, simulation)), compareBy = "path")
+
+  outputSelections <- simulation$settings$outputSelections
+
+  for (quantity in quantities) {
+    outputSelections$addQuantity(quantity)
+  }
+
+  return(quantities)
```
PavelBal: I think this method should not return anything, or silently the simulation.

R/utilities-simulation.R
PavelBal: I would vote for path strings ontly... I think this is more convenient.

R/error-checks.R
msevestre: I have a method that takes two optional arguments, but at least one of them should not be NULL.
I will change the signature of my method anyways so this won't be required but it could be useful...maybe ?

R/quantity-type.R
msevestre: I don't think we should name file according to object type. Otherwise some other file should be class-parameter etc..
Why do you propose this change?

R/utilities-simulation.R
msevestre: why not also support instances? Let say you have retrieved a bunch of quantities using getAllParametersMatching (or getAllQuantityMatching). You have updated the start values etc.. 
Now you could just do `addOutput(quantities, sim)` and be done. No need to go back to strings. 

In this method, I had two optional inputs (quantities and paths). I am going to change the function to be quantitiesOrPaths as one parameter. That was however an example of a method that takes two optional paraemters. One of which cannot be null. Does it make sense?

R/utilities-simulation.R
msevestre: @PavelBal Why? If we do not return the selected quantities, there is no way for the user to actually verify what outputs were automatically added (when using wild cards selection for example)

If the actual instances are returned, you can actually print them. As a user of this method, you are free to forget about the returnes value obviously

msevestre: But I agree. The returned value should be silent!

R/error-checks.R
PavelBal: Yepp, could be useful.

PavelBal:
Looks good to me, but one test is failing.

msevestre:
ok going to fix the tests, push and merge. Thanks @PavelBal

Issue #37, opened 2019-06-07
Load population parameters from CSV


Yuri05:
Format:
\#comment line 1
...
\#comment line N
"IndividualId","Gender","RaceIndex","Population Name","Parameter path 1 [unit]"..."Parameter path N [unit]"

e.g.
![grafik](https://user-images.githubusercontent.com/25061876/59118425-6405c700-8950-11e9-9575-0de496c90184.png)

PR #157, opened 2019-09-10
WIP #37 load population parameters from CSV

msevestre:


msevestre:
@all: I am not sure exactly that I understand how a population will be used in R.
I see so far two use cases
* Directly loaded from PKSim using parameters such as Gender, Species, Ethnicity etc.. (Similar to what we have now in Matlab.). This option will only be available on Windows provided that PKSim ins installed (prio2 for now)
* Loaded from a CSV file. This is what this PR is implementing

What properties of a simulation do we need? Can I get some example of use cases

Off the top of my head I see the need for
* Get the number of individual in a population
* Get the list of all parmaeter paths that define variability in the population

### Maybe
* Remove one individual ? 
* Remove one parameter from being varied in the population (similar to removing a column in the csv)
* Adding a new parameter variability?
* Updating a parameter variability?

Please feedback

R/population.R
```diff
@@ -0,0 +1,25 @@
...
+#' \describe{
+#'   \count{containerType}{Type of container}
+#'   }
+#'
+Population <- R6::R6Class(
+  "Population",
+  inherit = DotNetWrapper,
+  active = list(
+    count = function(value) {
+      private$wrapReadOnlyProperties("Count", value)
```
msevestre: Exposing only the `count` property for now. Let's discuss what we need here

R/utilities-population.R
```diff
@@ -0,0 +1,8 @@
+#' @export
+loadPopulation <- function(fileName, simulation) {
+  validateIsOfType(simulation, "Simulation")
+  validateIsString(fileName)
+  populationImporter <- getNetTask("PopulationImporter")
+  population <- rClr::clrCall(populationImporter, "ImportPopulation", fileName, simulation$ref)
```
msevestre: This is in fact not really a population but a `bag` of parmaeter values by path. Because I do not think we will have a Population object in R, I think it makes sense to call this `bag` a population unless you see a problem

R/utilities-population.R
```diff
@@ -0,0 +1,8 @@
+#' @export
+loadPopulation <- function(fileName, simulation) {
```
msevestre: Note: Loading a population from file requires a simulation. Why? Because we need to be able to check that the file does not contain unknown parameters and also that we can remove the [unit] if part of the exported path, as it is now the default coming from PKSim. For that, we need a simulation to be able to retrieve valid parameter instances

tests/dev/script.R
```diff
@@ -12,23 +12,27 @@ addOutputs(parameter, simulation = sim)
 # parameter <- getParameter("Organism|Liver|Volume", sim)
 # print(parameter)
 
-results <- runSimulation(sim)
-exportResultsToCSV(results, sim, "C:/temp/export/results.csv")
 
-pkAnalyses <- calculatePKAnalyses(results, sim)
+population <- loadPopulation("C:/projects/OSPSuite-R/tests/data/pop_10.csv", sim)
```
msevestre: Example of population load. All the magic, parsing etc.. happens in Core

msevestre:
@StephanSchaller Maybe you could also explain how you would like, in an ideal world, to use a population in R

KatrinCoboeken:
So far it looks good to me.

Will it be possible to add parameters, which are part of the simulation but normally constant?
For example  add the dose as population parameter? We do this in the current implementation of the reporting engine if we have a dose wich is adjusted by body weight  range ( between 2 and 3 kg 1 mg, between 3 and 5 kg 1.2 ,....)

In the current application we add sometimes also parameters which are not part of the simulationfile, this parameters are then not used for the simulation but to generate plots like fraction unbound vs age, fraction unbound is than somthing calculated by the other parameters like the ontigeny factors. In princip same as BMI and BW.

R/utilities-population.R
PavelBal: I don't really like this. The "population" - which is just a set of parameters - is not really bound to any specific simulation. So in theory I could apply the same population to different simulation.

Also, at the moment, the simulation for which the population is loaded is not stored anywhere, right? So after loading the population for the simulation `sim1` I could apply it (though we still do not know how) to another simulation `sim2`?

> we can remove the [unit] if part of the exported path, as it is now the default coming from PKSim.

I assume that the values coming from PK-Sim are always in default units? In that case, we can just remove the unit part.

Also I do not see a problem if a population contains parameters that are not in the simulation.

msevestre: I agree this is not ideal.  However there is no way otherwise to check the validity of parameters. It becomes important because  your statement `In that case, we can just remove the unit part.` is not accurate

Unfortunately, some parameters have [UNIT] in their name!! (@Yuri05 why is this allowed again) so one cannot simply remove the [xx] part of the name. 

If I remove it all the time, some parameters won't be found. If I never remove it, no parameters will ever be found.

>Also I do not see a problem if a population contains parameters that are not in the simulation.

No that's ok. However a population that only has parameter that do not exist is useless. 

I am not sure what's the best way forward here. In PKSim we don't have the issue because s you alwyas have to select an individual to load a population from csv (which gives us the structure)

@Yuri05 Thoughts?

Yuri05: We should not remove any parts from parameter names. If parameter is called "X [unit]" for whatever reason - we have to live with this. In MoBi user can name parameters whatever he likes (unless we e.g. prohibit usage of [])

PavelBal: @Yuri05 But PK-Sim exports the population parameters with a `[unit]` suffix:

`"Organism|Weight [kg]"`

So in order to find this parameter in the simulation and change it's value the suffix must be removed...

PavelBal:
> Maybe you could also explain how you would like, in an ideal world, to use a population in R

Something like `runPopulationSimulation(Simulation, Population)`. The values of the `Population` are then applied to the model in `Simulation`.

R/utilities-population.R
Yuri05: Ah ok, I misunderstood what you meant.
Are **all** parameters exported like that? (e.g. how are dimensionless parameters exported: like `"P []"` or just `"P"`?)
* If yes: just remove **last occurance** of [...] in the parameter name
* If no:
  * first remove **last occurance** of [...] in the parameter name and check if parameter exists in simulation
  * if no: check if parameter with original name (nothing removed) exists

msevestre: The [] must be removed. However the only way to remove it  for sure is to have a simulation has a parameter
But having a simulation to create a population is weird.


>Are all parameters exported like that?

No dimensionsless parameter have no unit exporter

> first remove last occurance of [...] in the parameter name and check if parameter exists in simulation

That's the whole point. Here we need a simulation and it sucks from an API point of view

msevestre:
>Something like runPopulationSimulation(Simulation, Population). The values of the Population are then applied to the model in Simulation.

Yeah that's clear. I meant how is the OBJECT Population going to be used in R. What properties do we need on the object.

R/utilities-population.R
Yuri05: Well, I would _import_ a population without any change of parameter paths.
But **at the time point where a population is applied to a simulation**: I would check and try with and without units.

msevestre: Ok that was my next proposal. In that case, we'll have to split core and pksim implementation however because the behavior in PKSIm will be different.

PavelBal: > But at the time point where a population is applied to a simulation: I would check and try with and without units.

Actually I also wanted to propose this..

PavelBal:
>  What properties do we need on the object.

-  Nr. of individuals
- Maybe ranges of the most important properties (% female, weight range, age range), of possible
- Possibility to extract an individual (values set)
- Change values of parameters for individuals (possible?)
- Add parameters (path + value) per individual.

R/utilities-population.R
msevestre: Well we are all in agreement. I'll implement it that way

StephanSchaller:
What is the csv population file content, and how is it "loaded", i.e. where is it loaded to and how do you simiulate it?
E.g. currently we would loop over all individuals and set the parameters and simulate one by one, store the results in a R data frame and then plot.

Is the csv structured as it is now when we export a population from PK-Sim?

Ideally, we have the same functionality as in PK-Sim. E.g. add user-defined variability (for any parameter you want, e.g. "Dose" as Katrin is requesting).

msevestre:
@StephanSchaller The implementation will be completely different that what we have in the current implementation. You will not be in charge of looping, setting parameters etc.. .This is what the framework will do. 

>Is the csv structured as it is now when we export a population from PK-Sim?

yes. The CSV File is the file generated by pksim. 

As far as results is concerned, there is an ongoing discussion regarding the result structure #161

msevestre:
> Will it be possible to add parameters, which are part of the simulation but normally constant?
Yes. This will have to be added using a path and a value array (value for each individual)

>Possibility to extract an individual (values set)
Yep doable

>Change values of parameters for individuals (possible?)
Yep doable

>Add parameters (path + value) per individual.
Same as @KatrinCoboeken suggestion

R/error-checks.R
```diff
@@ -33,6 +33,14 @@ validateIsString <- function(object) {
   validateIsOfType(object, "character")
 }
 
+validateIsNumeric <- function(object) {
```
msevestre: @PavelBal I have added those guys because I am fed up of duplicating the same code all the time

R/population.R
```diff
@@ -0,0 +1,41 @@
...
+    }
+  ),
+  public = list(
+    has = function(parameterPath){
+      validateIsString(parameterPath)
+      rClr::clrCall(self$ref, "Has", parameterPath)
+    },
+    setValues = function(parameterPath, values){
```
msevestre: `setValues`  and `getValues` are methods that takes a path and either set or return the values

tests/testthat/test-population.R
```diff
@@ -0,0 +1,49 @@
...
+
+
+test_that("It can return the expected number of individual in the population", {
+  population <- loadPopulation(populationFileName)
+  expect_equal(population$count, 10)
+})
+
+test_that("It returns whether variability is defined for a parameter path", {
+  population <- loadPopulation(populationFileName)
+  expect_true(population$has("Organism|VenousBlood|Volume [l]"))
```
msevestre: I have updated the code so that it supports both query with and without units.
The reasoning behind that is that the path beign returned from a Quantity does not have the units. It would be very confusing for the user to ask `population$has("Organism|VenousBlood|Volume")` and get no as an answer.

tests/testthat/test-population.R
```diff
@@ -0,0 +1,49 @@
...
+test_that("It can add user defined variability using an existing parameter path with unit", {
+  population <- loadPopulation(populationFileName)
+  parameterPath <- "Organism|VenousBlood|Volume [l]"
+  expect_true(population$has(parameterPath))
+  population$setValues(parameterPath, values)
+  expect_true(population$has(parameterPath))
+  expect_identical(population$getValues(parameterPath), values)
+})
+
+test_that("It can add user defined variability using an existing parameter path without unit", {
```
msevestre: The logic is: If the parameter was added with unit (e.g. loaded from CSV for example) it can be retrieve without the unit
However if the parameter is added WITHOUT unit (for exampel to create user defined variability, one would use the parameter$path as a parameter), then you can only retrieve the parameter without unit (because anything else does not make sense)

I think it's fairly resilient that way and allows  all features requested to be done

msevestre:
ok merging this. We can update with new features as requested

PR #164, opened 2019-09-16
Fixes #35 load simulation results

### LINKED ISSUE #35: Load (pre-)calculated results (from CSV)
Additionally to the calculation of simulation results for a model it should be possible to load the results from a CSV file.

Data format:

IndividualId|Time [unit]|Output 1 [unit]|...|Output N  [unit]
-------------|-----------|-------------------|--|-----------------

e.g. 

IndividualId|Time [min]|Organism\|Bone\|Plasma\|C1\|Concentration [µmol/l]|...|Organism\|Brain\|Plasma\|C1\|Concentration [µmol/l]
-------------|-----------|-------------------|--|-----------------
0|0|0|...|0
0|15|0.01|...|0.21
...|...|...|...|...
1|0|0|...|0
1|15|0.33|...|0.001
...|...|...|...|...

If more than 1 individual is given: time raster must be the same for all individuals

msevestre:


R/utilities-results.R
```diff
@@ -0,0 +1,57 @@
...
+#' resultPath <- system.file("extdata", "res.csv", package = "ospsuite")
+#'
+#' # Load the simulation
+#' sim <- loadSimulation(simPath)
+#'
+#' # Run the simulation
+#' results <- importResultsFromCSV(sim, resultPath)
+#'
+#' @export
+importResultsFromCSV <- function(simulation, fileNames) {
```
msevestre: In order to import results, one needs to pass the simulation and one or mroe files to import (many files are potentially possible when simulating on cluster). 
This is exactly the same behavior as in PKSim (code wwas extracted and moved to core)

tests/testthat/test-utilities-results.R
```diff
@@ -0,0 +1,37 @@
...
+
+test_that("It throws an exception if the file imported are not valid results file", {
+  junkFile <- getTestDataFilePath("pop_10.csv")
+  expect_that(importResultsFromCSV(sim, junkFile), throws_error())
+})
+
+test_that("It throws an exception when importing a valid result file that does not match the simulation", {
+  otherSim <- loadTestSimulation("simple")
+  resFile <- getTestDataFilePath("res_10.csv")
+  expect_that(importResultsFromCSV(otherSim, resFile), throws_error())
```
msevestre: Obviously importing a valid file that does not match the simulation throws an exception. Tests are awesome!

R/utilities-pk-analyses.R
```diff
@@ -27,3 +27,20 @@ calculatePKAnalyses <- function(results, simulation) {
 toPKParameter <- function(netPKParameters) {
   toObjectType(netPKParameters, PKParameter)
 }
+
+
+#' Saves the pK-analyses  to csv file
+#'
+#' @param pkAnalyses pK-Analyses to exporte (typically calculated using \code{calculatePKAnalyses} or imported from file)
+#' @param simulation Instance of a simulation used to calculate the pK-Analyses
+#' @param filename Full path of where the pK-Analyses will be saved.
```
PavelBal: Please rename `filename` to `outputFilename` or something, so it is clear _what_ the file is for.

PavelBal: `outputCSVname`...

Is it really just the name, or the full path to the output file?

R/utilities-pk-analyses.R
```diff
@@ -27,3 +27,20 @@ calculatePKAnalyses <- function(results, simulation) {
...
+
+
+#' Saves the pK-analyses  to csv file
+#'
+#' @param pkAnalyses pK-Analyses to exporte (typically calculated using \code{calculatePKAnalyses} or imported from file)
+#' @param simulation Instance of a simulation used to calculate the pK-Analyses
+#' @param filename Full path of where the pK-Analyses will be saved.
+#'
+#' @export
+exportPKAnalysesToCSV <- function(pkAnalyses, simulation, filename) {
```
PavelBal: I think this is very confusing, passing the simulation all the time...

How about we store the reference to the simulation object in the `SimulationRersults` class, and a reference to the `SimulationRersults` object in the `SimulationPKAnalyses` class? This way the user could always have access to the specific results that were used for calculating the PK-analysis, and don't have to care about passing the simulation around?

This has not to be done in this PR, could open a separate issue.

R/utilities-results.R
```diff
@@ -0,0 +1,57 @@
...
+#' # Add some outputs to the simulation
+#' addOutputs("Organism|**|*", sim)
+#'
+#' # Run the simulation
+#' results <- runSimulation(sim)
+#'
+#' # Export the results to csv file
+#' exportResultsToCSV(results, sim, tempfile())
+#' @export
+exportResultsToCSV <- function(results, simulation, fileName) {
```
PavelBal: Same here - rename `fileName` to `outputCSVname` or even `outputCSVpath`?

R/utilities-results.R
PavelBal: `csvResultsPath`

R/utilities-pk-analyses.R
msevestre: Yes this is a great suggestion. It will make the API easier to use for the user
I'll do it in this PR

R/utilities-pk-analyses.R
PavelBal: I like when I know how to use the function without reading the comments :)

msevestre: Of course. but how do you know what parameters to call, the order etc.... At some point you need to read the docs or at least look at the intellisense in RStudio (which shows the description of parameters)

PavelBal: Hm, I'm not getting the description in the intellisense, only the names of the parameters... Have to check the settings.

msevestre: I think it only works when you import the packages (not in dev). 
At any rate, I'll rename. But we have to assume that people are going to read the F.. manual otherwise this will be a lost cause ;)

R/utilities-results.R
msevestre: I don't like having acrnonyn like CSV in the name of the parameters. What about `outputFilePath` ? That way we can be consistent and use this allover

msevestre: or we specify what the file is about (not the extension)
resultsFilePath, simulationFilePath, pkanalysesFilePath etc...

msevestre: or we follown ggplot. and call it filePath (does nto matther if it;s import export etc..)
the context is clear and there is doc
I'd vote for `filePath` to be hontest

PavelBal: Yeah OK, I can live with `filePath` :)

R/utilities-pk-analyses.R
msevestre: done. filePath

R/utilities-pk-analyses.R
msevestre: done. Sim is now a reference in SimPK and SimRes. That way we do not need to pass it as an extra parameter

R/api-config.R
```diff
@@ -6,10 +6,10 @@ ApiConfig <- R6::R6Class(
   inherit = DotNetWrapper,
   active = list(
     pkParametersFilePath = function(value) {
-      private$wrapProperties("PKParametersFilePath", value)
+      private$wrapProperty("PKParametersFilePath", value)
```
msevestre: @PavelBal I have also renamed this guys. They should be singluar not plural

R/dot-net-wrapper.R
```diff
@@ -42,6 +42,14 @@ DotNetWrapper <- R6::R6Class(
       }
     },
 
+    readOnlyProperty = function(propertyName, value, returnValue) {
```
msevestre: @PavelBal This is a new method to wrap an actual readonly prop wthout having to duplicate all the code everywhere. We can use it here and there to simplify some of the existing code. (haven't done it)

R/simulation-pk-analyses.R
```diff
@@ -20,5 +29,10 @@ SimulationPKAnalyses <- R6::R6Class("SimulationPKAnalyses",
       validateIsString(pkParameter)
       toPKParameter(rClr::clrCall(self$ref, "PKParameterFor", quantityPath, pkParameter))
     }
+  ),
+  active = list(
+    simulation = function(value) {
```
msevestre: Simulation is a readonly field.

R/simulation-pk-analyses.R
```diff
@@ -10,7 +11,15 @@
 #'
 SimulationPKAnalyses <- R6::R6Class("SimulationPKAnalyses",
   inherit = DotNetWrapper,
+  private = list(
+    .simulation = NULL
+  ),
   public = list(
+    initialize = function(ref, simulation) {
```
msevestre: simulation in constructor

PavelBal:
No further objections, can merge.

PR #167, opened 2019-09-17
22 change solver settings

msevestre:
@PavelBal I have also the changes in SimulationResults for you to return allQuantityPath


R/container.R
```diff
@@ -1,11 +1,8 @@
 #' @title Container
 #' @docType class
 #' @description  Contains other entities such as Parameter or containers
-#' @section Methods:
-#' \describe{
-#'   \item{containerType}{Type of container}
-#'   }
-#'
+#' @field containerType Type of container
```
msevestre: THis was wrong

R/simulation-results.R
```diff
@@ -24,6 +25,13 @@ SimulationResults <- R6::R6Class("SimulationResults",
     },
     simulation = function(value) {
       private$readOnlyProperty("simulation", value, private$.simulation)
+    },
+    allQuantityPaths = function(value){
```
msevestre: All Quantity Path. Going to add a test for this

msevestre:
@PavelBal I don't want to add a test for SimulationResults to avoid merge conflicts with what you are doing. I'll add the tests once your changes are integrated
My sugggestion would be for you to accept this PR. Then we merge and you can rebase of develop

PR #169, opened 2019-09-17
Get simulation result

PavelBal:
@msevestre WIP. See if the outputs are OK, I will write tests later.


R/simulation-results.R
```diff
@@ -16,6 +16,17 @@ SimulationResults <- R6::R6Class("SimulationResults",
       validateIsOfType(simulation, "Simulation")
       private$.simulation <- simulation
       super$initialize(ref)
+    },
+    hasResultsForIndividual = function(individualId) {
```
PavelBal: Some access to the field in .NET

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+#' simResults <- runSimulation(sim)
+#'
+#' paths <- c(
+#'   "Organism|PeripheralVenousBlood|Caffeine|Plasma (Peripheral Venous Blood)",
+#'   "Organism|ArterialBlood|Plasma|Caffeine|Concentration in container"
+#' )
+#'
+#' resultsValues <- getResultsValues(simResults, paths)
+#' # plot(resultsValues$`Organism|PeripheralVenousBlood|Caffeine|Plasma (Peripheral Venous Blood)`)
+getResultsValues <- function(simulationResults, quantitiesOrPaths) {
```
PavelBal: `getResultsValues` is a helper function that calls `getPopulationResultsValues`.

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+#' resultsValues <- getPopulationResultsValues(simResults, paths, individualIds = 0)
+#' # plot(resultsValues$`Organism|PeripheralVenousBlood|Caffeine|Plasma (Peripheral Venous Blood)`)
+getPopulationResultsValues <- function(simulationResults, quantitiesOrPaths, individualIds = NULL) {
+  validateIsOfType(simulationResults, "SimulationResults")
+  quantitiesOrPaths <- c(quantitiesOrPaths)
+  validateIsOfType(quantitiesOrPaths, c("Quantity", "character"))
+  validateIsNumeric(individualIds, nullAllowed = TRUE)
+
+  # If quantities are passed, get their paths.
+  paths <- quantitiesOrPaths
```
PavelBal: Can receive either quantities or paths.

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+  if (isOfType(quantitiesOrPaths, "Quantity")) {
+    paths <- lapply(quantitiesOrPaths, function(x) x$path)
+  }
+  paths <- unique(paths)
+
+  # If no specific individual ids are passed, iterate through all individuals
+  if (is.null(individualIds)) {
+    nrOfIndividuals <- simulationResults$count
+    # TODO: Can we assume that all individuals ids are ascending number from 0 to the number of individuals-1??
```
PavelBal: @msevestre I am really not sure if this would always work, but there was no way to get the IDs of all available individuals.

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+    nrOfIndividuals <- length(individualIds)
+  }
+
+  # Pre-alloate a list with the number of entries equal to the number of paths
+  output <- vector("list", length(paths))
+  names(output) <- paths
+  # All time values are equal
+  timeVals <- simulationResults$timeValues
+
+  for (path in paths) {
```
PavelBal: The output contains entries for all paths (or quantities) that were provided - even if they are not in the results! I thing this is fair since the user might try to access them, and it is better to return a `NULL` rather then just to completely remove the entry from the outputs...

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+    outputValues <- vector("numeric")
+
+    for (individualId in individualIds) {
+      if (simulationResults$hasResultsForIndividual(individualId)) {
+        individualResult <- simulationResults$resultsForIndividual(as.integer(individualId))
+        # Check if the output exist. If not, the entry in the list of outputValues will have the value NULL
+        if (rClr::clrCall(individualResult, "HasValuesFor", path)) {
+          quantityValues <- rClr::clrCall(individualResult, "ValuesFor", path)
+          vals <- rClr::clrGet(quantityValues, "Values")
+          outputValues <- cbind(outputValues, vals, deparse.level = 0)
```
PavelBal: Once we have a population simulation, we can actually test if it is a convenient output.

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+        }
+      }
+    }
+    # Only fill the entry if at least one individual has the output
+    if (!(length(outputValues) == 0)) {
+      output[[path]][["x"]] <- timeVals
```
PavelBal: Calling the entries x and y allows to simply use them in `plot(output$selectedPath)`.

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+
+  # Pre-alloate a list with the number of entries equal to the number of paths
+  output <- vector("list", length(paths))
+  names(output) <- paths
+  # All time values are equal
+  timeVals <- simulationResults$timeValues
+
+  for (path in paths) {
+    # Pre-alloate a list with the number of entries equal to the number of individuals
+    outputValues <- vector("numeric")
```
msevestre: No allocation here (contrary to comment directly above)

R/utilities-simulation-results.R
msevestre: This would not work if the simulation of a specific individual did not work. This needs to come from SimulationResults as well as  new property.

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+
+    for (individualId in individualIds) {
+      if (simulationResults$hasResultsForIndividual(individualId)) {
+        individualResult <- simulationResults$resultsForIndividual(as.integer(individualId))
```
msevestre: casting to integer to be done in `resultsForIndividual` method. This is purely a .net issue

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,132 @@
...
+        # Check if the output exist. If not, the entry in the list of outputValues will have the value NULL
+        if (rClr::clrCall(individualResult, "HasValuesFor", path)) {
```
msevestre: Since we do not have an `IndividualResult` object, this code should be moved into our `SimulationResults` . I don't think  we should have any `rClr` stuff leaking outside of our R6 class.

R/simulation-results.R
```diff
@@ -16,6 +16,17 @@ SimulationResults <- R6::R6Class("SimulationResults",
       validateIsOfType(simulation, "Simulation")
       private$.simulation <- simulation
       super$initialize(ref)
+    },
+    hasResultsForIndividual = function(individualId) {
+      rClr::clrCall(self$ref, "HasResultsFor", as.integer(individualId))
+    },
+    resultsForIndividual = function(individualId) {
+      rClr::clrCall(self$ref, "ResultsFor", individualId)
```
msevestre: use as.integer here so that caller does not have to worry about this

R/messages.R
```diff
@@ -8,7 +8,7 @@ messages <- list(
 
     message <- paste0(
       callingFunction, ": argument '", objectName,
-      "' is of type '", typeof(objectName), "', but expected '", expectedTypeMsg, "'!", optionalMessage
```
PavelBal: The type was always "character"..

R/simulation-results.R
```diff
@@ -3,19 +3,55 @@
...
+#' @field timeValues Vector of simulated time output values
+#' @field allQuantityPaths List of all paths for which results are defined.
+#' @field allIndividualId List of Ids of all individuals that have been simulated
 #' @format NULL
 SimulationResults <- R6::R6Class("SimulationResults",
   inherit = DotNetWrapper,
   private = list(
-    .simulation = NULL
+    .simulation = NULL,
+    getResultsForIndividual = function(individualId) {
```
PavelBal: I moved these to `private` as they return .NET objects and we don't want to expose them..

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,129 @@
...
+  quantitiesOrPaths <- c(quantitiesOrPaths)
+  validateIsOfType(quantitiesOrPaths, c("Quantity", "character"))
+  validateIsNumeric(individualIds, nullAllowed = TRUE)
+
+  # If quantities are passed, get their paths.
+  paths <- quantitiesOrPaths
+  if (isOfType(quantitiesOrPaths, "Quantity")) {
+    paths <- unlist(lapply(quantitiesOrPaths, function(x) x$path))
+    # Remove the name of the simulation
+    splitPaths <- strsplit(paths, "|", fixed = TRUE, perl = FALSE, useBytes = FALSE)
```
PavelBal: Don't really like this one, but I didn't come up with a better solution right now.

msevestre: If you go to entity class, we have
```
path = function(value) {
      private$wrapExtensionMethod("OSPSuite.Core.Domain.EntityExtensions", "EntityPath", "path", value)
    }
```

You can define another method using the "ConsolidatedPath" method
It removes the first entry effectively of the path for you

Something like
```
consolidatePath = function(value) {
      private$wrapExtensionMethod("OSPSuite.Core.Domain.EntityExtensions", "ConsolidatedPath", "consolidatePath", value)
    }
```

Don't forget to add the corresponding @field document and explain that this removes the simulation path. This should normally never be used outside of our helper methods

msevestre:
@PavelBal Going to update this branch with the latest OSPSuite.Core
One sec

PavelBal:
@msevestre wait a moment...

msevestre:
ok. But it should read `AllIndividualIds` not `AllIndividualId`.
Then I can update the binaries

PavelBal:
Waaaah can I just delete this branch and push a new one, based on develop? Git just completely destroyed my brain...

msevestre:
>Waaaah can I just delete this branch and push a new one, based on develop? Git just completely destroyed my brain...

Do you want me to check it out? As long as all your code is on this branch, I can update it

PavelBal:
I did a rebase based on develop and actually have a clean version locally, but when I try to push it to this branch, I got errors that my local branch is outdated...

msevestre:
ok `git push upstream get-simulation-result -f`

-f will force the update

PavelBal:
OK, at least it does not fail any more...

R/simulation-results.R
```diff
@@ -26,12 +61,26 @@ SimulationResults <- R6::R6Class("SimulationResults",
...
+      rClr::clrGet(quantityValuesTime, "Values")
+    },
+    allQuantityPaths = function(value) {
       if (missing(value)) {
         rClr::clrCall(self$ref, "AllQuantityPaths")
       } else {
         private$throwPropertyIsReadonly("allQuantityPaths")
       }
+    },
+    allIndividualIds = function(value) {
```
msevestre: @PavelBal I have renamed this guys with an `s`

tests/testthat/test-utilities-simulation-results.R
```diff
@@ -0,0 +1,58 @@
...
+context("getResultsValues")
+
+test_that("It can retrieve results by paths", {
+  results <- getResultsValues(simResults, resultsPaths)
+  expect_equal(length(results), length(resultsPaths))
+})
+
+
+test_that("It throws an error when the results contain more than one individual", {
+  # TODO
```
msevestre: @PavelBal There is still a TODO here

R/utilities-simulation-results.R
msevestre: I don't understand this methods. IT calls `getPopulationResultsValues` with exactly the same arguments. Why do we need this at all?

msevestre: aside from the check it is exactly the same method

PavelBal: Yes - it's just for convenience to differentiate between population and individual simulations.

tests/testthat/test-utilities-simulation-results.R
PavelBal: We don't have simulations with >1 individual yet, so cannot test..

R/utilities-simulation-results.R
msevestre: Still. Why do we ned to method. Same input, and same outputs. Aren't those methods equivalent?

Looking at the code, I would rename getPopulationResultsValues to getResultsValues and delete getResultsValues 

Or am I missing something ?

PavelBal: The only difference is that `getResultsValues` stops if trying to receive results with more than one individual - safety check that we are not using a population simulation where we expect a single individual...

But we can leave only one of the methods if there is no demand for this check.

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,126 @@
...
+
+  for (path in paths) {
+    # Pre-allocate the output matrix. The number of rows equals the number of simulated time points.
+    # The number of columns is the number of possible individuals.
+    outputValues <- matrix(nrow = length(timeVals), ncol = max(length(individualIds), simulationResults$count), dimnames = NULL)
+
+    for (individualIdx in seq_along(individualIds)) {
+      individualId <- individualIds[individualIdx]
+      vals <- simulationResults$getValuesForIndividual(resultPath = path, individualId = individualId)
+      if (is.null(vals)) {
```
msevestre: That would mean asking for invalid path or results. Should not we throw an exception here? This is clearly an input error no?

R/utilities-simulation-results.R
```diff
@@ -0,0 +1,126 @@
...
+        next
+      }
+      outputValues[, individualIdx] <- vals
+    }
+    # Remove all NAs, i.e. columns for individuals that are not present in the
+    # results. Solution from https://github.com/sfirke/janitor
+    mask_keep <- colSums(!is.na(outputValues)) > 0
```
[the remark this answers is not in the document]
PavelBal: I first allocate the matrix with the number of colums equal to the number of passed individual ids,  but limited by the total number of simulated individuals. So allocating with the most possible size. I did this because, if I understand it correctly, it prevents the matrix being copied every time it is changed in the loop, which would make the loop potentially very slow.

After filling the matrix, it is reduced to the dimensions equal to the number of individuals the results were found for. So yes, this will most probably happen only if the user passes wrong ids.

msevestre:
@PavelBal Output looks ok. However, we'll have to check with what will be required for TLF as we do not want to have to convert all the time to plot stuff. TLF will expect some data.frame as input. Maybe we'll have to make sure we create a data.frame instead of  a list.  To see in a few weeks

R/utilities-simulation-results.R
PavelBal: I am thinking of use cases where the user prepares a set of output paths and re-uses it between simulations/projects. Let's say my default output selection contain 20 output paths. Depending on the model I simulate (e.g. with or without the metabolite), the results might contain only a sub-set of this default output selection. It would be OK from my point of view to just return the results that actually are present and ignore the rest.

msevestre: Ok I am going to leave @KatrinCoboeken and @Yuri05 to decide on that. I think this could cause serious headache in the long run where the user won't understand what's going on because the wrong results are returned or missing.

PavelBal: How about a compromise in form of a `stopIfNotFound` argument, with default `TRUE`, as already discussed for getting parameters? In default state, the function would throw an error if an output or an individual id is not found. This behavior could be overridden by setting `stopIfNotFound = FALSE`, then it would behave like it does now.

R/utilities-simulation-results.R
msevestre: @PavelBal I would not know what would be the benefit for now. The beauty of having the same structure for single and multiple results (ind vs pop sim) is that you don't have to call different methods depending on the use case.

R/utilities-simulation-results.R
msevestre: sure. I am going to create another issue for that so that we can merge this guy

PR #170, opened 2019-09-17
168 change simulation intervals

msevestre:


msevestre:
@all: Please review the API and let me know if that works for you
It is possible to 
* Clear the intervals of an existing simulation (E.g. start from scratch)
* Add  a new interval
* Update an existing interval
* Add time points

R/output-schema.R
```diff
@@ -0,0 +1,75 @@
...
+      if (length(timePoints) > 1) {
+        rClr::clrCall(self$ref, "AddTimePoints", timePoints)
+      }
+      else {
+        rClr::clrCall(self$ref, "AddTimePoint", timePoints)
+      }
+      invisible(self)
+    },
+
+    print = function(...) {
```
msevestre: Example of output
```
OutputSchema: 
  Time points: 10, 20
Interval: 
  Name: Simulation interval high resolution
  Start time: 0.00 [min]
  End time: 120.00 [min]
  Resolution: 0.33333 [pts/min]
Interval: 
  Name: Simulation interval low resolution
  Start time: 120.00 [min]
  End time: 1440.00 [min]
  Resolution: 0.066667 [pts/min]
```

R/utilities-output-schema.R
```diff
@@ -0,0 +1,64 @@
...
+#' clearIntervals(sim)
+#'
+#' # Adds a new interval starting at 1h and ending at 10h with a resolution of 10 points per hour
+#' addInterval(sim, 1 * 60, 10 * 60, 1 / 6)
+#'
+#' # Adds another interval starting at 10h and ending at 17h with a
+#' # default resolution and a specified name
+#' addInterval(sim, 10 * 60, 17 * 60, intervalName = "Second Interval")
+#' @export
+addInterval <- function(simulation, startTime, endTime, resolution = 4 / 60, intervalName = NULL) {
```
msevestre: First API method that I believe will be quite useful. 
I do not know if we need to define a value for the resolution by default

R/interval.R
```diff
@@ -0,0 +1,46 @@
...
+#'
+#' @field startTime Start time of interval (instance of \code{Parameter})
+#' @field endTime End time of interval (instance of \code{Parameter})
+#' @field resolution Resolution of interval in pts/min (instance of \code{Parameter})
+#' @format NULL
+Interval <- R6::R6Class(
+  "Interval",
+  inherit = Container,
+  active = list(
+    startTime = function(value) {
```
msevestre: Start time, EndTime etc. .are actual instance of Parmaeter. One can do whatever we want we those parameters (update, value etc..)

R/utilities-output-schema.R
PavelBal: I would say the resolution must always be suppplied by the caller.

R/utilities-output-schema.R
```diff
@@ -0,0 +1,64 @@
...
+  validateIsOfType(simulation, "Simulation")
+  validateIsNumeric(c(startTime, endTime, resolution))
+  schema <- simulation$outputSchema
+  outputIntervalFactory <- getNetTask("OutputIntervalFactory")
+  interval <- toIntervals(rClr::clrCall(outputIntervalFactory, "CreateFor", schema$ref, startTime, endTime, resolution))
```
PavelBal: I thought "no clr stuff outside of R6" ;)

R/utilities-output-schema.R
```diff
@@ -0,0 +1,64 @@
...
+  if (!is.null(intervalName)) {
```
PavelBal: What happens if  the user adds two intervals with identical names?

R/utilities-output-schema.R
```diff
@@ -0,0 +1,64 @@
...
+#'
+#' clearIntervals(sim)
+#' @export
+clearIntervals <- function(simulation) {
+  validateIsOfType(simulation, "Simulation")
+  simulation$outputSchema$clear()
+  invisible(simulation)
+}
+
+
```
PavelBal: I would also like to have a method `setInterrvals`, which first calls `clearIntervals` and then `addInterval`.

R/utilities-output-schema.R
PavelBal: I also suggest calling the methods `addOutputInterval` etc. in line with MoBi and PK-Sim.

msevestre: OK. I don't like the internal unit of pts/min
This is just a unit added for consistency in our internal system. 
Even min and max in [min] is weird.

msevestre: ok

R/utilities-output-schema.R
msevestre: Fair enough. But that's a complete different use case here. 
Here we are resolving an instance of a Service "OutputIntervalFactory". We do this everywhere else to run a simulation, export stuff etc.. It's ok to call the method of the service using rClr because we do not WRAP the object with our own class.

In the case from yesterday, we have R6 classes and as such the behavior should be encapsulated. 
Does it make sense?

R/utilities-output-schema.R
msevestre: Exception is thrown. There is a test for that also

R/utilities-output-schema.R
msevestre: setIntervals with an S.
What would be the signature of this guy?

Maybe addOutputInterval should get an array of parameters instead of one

e.g. 
addOutputInterval (sim, c(min=1, max=2, res=3))

or 
addOutputInterval (sim,  min= c(1, 10, 20), max=c(2, 12, 22)....)

Thoughts?

Or is there a better construct in R

R/utilities-output-schema.R
PavelBal: Why is it weird? I think it strongly depends on the scenarios simulated. In typical PBPK hours maybe the more common unit, but in QSP minutes make more sense I would say.

R/utilities-output-schema.R
PavelBal: Yepp, just wanted to complain :P

R/utilities-output-schema.R
PavelBal: OK great

R/utilities-output-schema.R
PavelBal: No actually it was a typo - `setInterval` without s.
Most of the time I am working with only one interval, and, similar to what is done in the "old" R-toolbox, change the simulation time by setting the interval.

The way you implemented it now, the user has to remember to clear the outputs first and add new interval afterwards - two steps just to change the total simulation time.

So `setInterval` would just overwrite the output intervals by one single interval. If the user wants to have multiple intervals, he has to add the one by one - I wouldn't over-complicate the signature of the method to support multiple intervals in one tun..

R/utilities-output-schema.R
```diff
@@ -0,0 +1,82 @@
...
+#' clearOutputIntervals(sim)
+#'
+#' # Adds a new interval starting at 1h and ending at 10h with a resolution of 10 points per hour
+#' addOutputInterval(sim, 1 * 60, 10 * 60, 1 / 6)
+#'
+#' # Adds another interval starting at 10h and ending at 17h with a resolution of 4 points per hour
+#' # and a specified name
+#' addOutputInterval(sim, 10 * 60, 17 * 60, 4/60,  intervalName = "Second Interval")
+#' @export
+addOutputInterval <- function(simulation, startTime, endTime, resolution, intervalName = NULL) {
```
msevestre: AddOutputInterval

R/utilities-output-schema.R
```diff
@@ -0,0 +1,82 @@
...
+#' @examples
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#'
+#' # Load the simulation
+#' sim <- loadSimulation(simPath, addToCache = FALSE, loadFromCache = FALSE)
+#'
+#' # Adds a new interval starting at 1h and ending at 10h with a resolution of 10 points per hour
+#' setOutputInterval(sim, 1 * 60, 10 * 60, 1 / 6)
+#' @export
+setOutputInterval <-function(simulation, startTime, endTime, resolution, intervalName = NULL) {
```
msevestre: setOutputInterval

R/utilities-output-schema.R
```diff
@@ -0,0 +1,82 @@
...
+#' @examples
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#'
+#' #Make sure we create a new simulation so that we do not impact other examples
+#' sim <- loadSimulation(simPath, addToCache = FALSE, loadFromCache = FALSE)
+#'
+#' clearOutputIntervals(sim)
+#' @export
+clearOutputIntervals <- function(simulation) {
```
msevestre: clearOutputIntervals

PavelBal:
Example fails, otherwise good to go.

Issue #171, opened 2019-09-18
Allow user to set parameter value in a specific unit


msevestre:
Right now, the quantity class exposes the value {get;set} method wich allows to set or retrieve a value in base unit

However, I am wondering if we should not expose a way to set a value in the desire `displayUnit` as we do in PKSim and MoBi

For example. No one will want to set height in dm. But this is our internal unit
should we allow something like 

`quantity.value <- "175 cm"`

Or something similar?

We have the concept of `valudInDisplayUnit` also. Maybe too confusing for  a user 

Thoughts
@PavelBal @Yuri05 @KatrinCoboeken

PavelBal:
I would prefer having methods like

- `toUnit(quantity, value, targetUnit)` which returns  the `value ` which is given in default units of the `quantity` in the unit specified by `targetUnit`.
-  `toBaseUnit(quantity, value, originUnit)` returns the `value` given in `originUnit` in the base unit of the `quantity`.

Yuri05:
we could also provide the method `quantity.SetValue(value, unit)`.
This would convert `value` from `unit` to the internal unit (or crash if not convertable)

PR #175, opened 2019-09-19
WIP #171 set units

msevestre:


R/messages.R
```diff
@@ -6,51 +6,47 @@ messages <- list(
...
-    return(message)
   },
 
   errorGetEntityMultipleOutputs = function(path, container, optionalMessage = NULL) {
     # Name of the calling function
     callingFunctions <- sys.calls()
     callingFunction <- sys.call(-length(callingFunctions) + 1)[[1]]
 
-    message <- paste0(
+    paste0(
```
msevestre: No need to instantiate a local variable just to return it. In simple case, let's just use the standard R features here. There is no ambiguity

R/quantity.R
```diff
@@ -1,9 +1,14 @@
+withDimensionExtensions <- "OSPSuite.Core.Domain.WithDimensionExtensions"
```
msevestre: @PavelBal Is this ok to define a constant like that? Just for this file. I was fedup of repeating it

tests/testthat/test-utilities-simulation-results.R
```diff
@@ -1,58 +1,58 @@
...
+  expect_equal(length(results), length(resultsPaths))
+})
+
+test_that("It can retrieve results with provided individual id", {
+  results <- getPopulationResultsValues(simResults, resultsPaths, individualIds = c(0, 0, 1))
+  expect_equal(length(results), length(resultsPaths))
+  expect_false(is.null(results[[1]]))
+})
+
+test_that("It returns NULL for results if no individual id was simulated", {
```
msevestre: @PavelBal I ran the styler. no changes otherwise

R/utilities-units.R
```diff
@@ -0,0 +1,44 @@
...
+#'
+#' @examples
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' par <- getParameter(sim, "Organism|Liver|Volume")
+#'
+#' # Converts the value in base unit (1L) to ml => 1000
+#' valueInMl <- toUnit(par, 1, "ml")
+#' @export
+toUnit <- function(quantity, value, targetUnit) {
```
msevestre: toUnit as described

R/utilities-units.R
```diff
@@ -0,0 +1,44 @@
...
+#'
+#' @examples
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' par <- getParameter(sim, "Organism|Liver|Volume")
+#'
+#' # Converts the value in unit (1000 ml) to the base unit (l) => 1
+#' valueInBaseUnit <- toBaseUnit(par, 1000, "ml")
+#' @export
+toBaseUnit <- function(quantity, value, unit) {
```
msevestre: toBaseUnit as described

R/utilities-units.R
```diff
@@ -0,0 +1,44 @@
...
+  validateIsNumeric(value)
+  validateHasUnit(quantity, unit)
+
+  rClr::clrCallStatic(withDimensionExtensions, "ConvertToBaseUnit", quantity$ref, value, unit)
```
msevestre: the only change I could see is to return a vector if the method is called with an array of value instead of one value

R/quantity.R
PavelBal: Don't see a problem with it. But maybe put it in UPPER_CASE according to our coding standards @msevestre

PR #176, opened 2019-09-20
Fixes #156 use class name for validation

### LINKED ISSUE #156: Update validation to use class name instead of string

msevestre:


R/error-checks.R
```diff
@@ -1,33 +1,51 @@
...
+
+  stop(messages$errorWrongType(objectName, class(object)[1], objectTypes))
 }
 
+typeNamesFrom <- function(type){
+  if (is.character(type)) {
+    return(type)
+  }
+  type <- c(type)
+  sapply(type, function(t) t$classname)
```
msevestre: This is just an extension that allows to either pass the class name or the class itself
e,g, validateIfOfType(obj, "Simulation") or validateIsOfType(obj, Simulation)

R/quantity.R
```diff
@@ -42,7 +42,7 @@ Quantity <- R6::R6Class(
       private$printLine("Quantity Type", getEnumKey(QuantityType, self$quantityType))
     },
     printValue = function() {
-      private$printQuantityValue(self$name)
+      self$printQuantityValue(self$name)
```
msevestre: this was a bug

R/utilities-simulation.R
```diff
@@ -112,7 +112,7 @@ resetSimulationCache <- function() {
 #' removeSimulationFromCache(sim1) # returns TRUE
 #' removeSimulationFromCache(sim2) # returns FALSE
 removeSimulationFromCache <- function(simulation) {
-  validateIsOfType(simulation, "Simulation")
+  validateIsOfType(simulation, Simulation)
```
msevestre: example of use. We can use either way. Just a convenience to be able to not rely on strings

PavelBal: Good idea. Let's merge it.

PR #180, opened 2019-09-26
Fixes #38 calculate population

### LINKED ISSUE #38: Calculate population
Given a model file and a set of population parameters (e.g. saved as csv file described in #37): calculate population simulation results.

msevestre:


R/simulation-run-options.R
```diff
@@ -0,0 +1,27 @@
+#' @title SimulationRunOptions
+#' @docType class
+#' @description  Options to be passed to the simulation engine
+#' @field numberOfCoresToUse (Maximal) number of cores to be used (1 per default)
```
msevestre: Some kind of meta configuration of the simulation run. This will be updated with more parameters if needed

R/utilities-pk-analyses.R
```diff
@@ -12,15 +11,13 @@
 #'
 #' addOutputs("Organism|VenousBlood|*|Caffeine", sim)
 #' results <- runSimulation(sim)
-#' pkAnalyses <- calculatePKAnalyses(results, sim)
+#' pkAnalyses <- calculatePKAnalyses(results)
```
msevestre: WE also don't need the sim here as it is defined in the results

R/utilities-simulation.R
```diff
@@ -70,22 +70,38 @@ saveSimulation <- function(simulation, filePath) {
...
+runSimulation <- function(simulation, population = NULL, simulationRunOptions = NULL) {
   validateIsOfType(simulation, Simulation)
+  validateIsOfType(population, Population, nullAllowed = TRUE)
+  validateIsOfType(simulationRunOptions, SimulationRunOptions, nullAllowed = TRUE)
+  options <- simulationRunOptions %||% SimulationRunOptions$new()
   simulationRunner <- getNetTask("SimulationRunner")
-  results <- rClr::clrCall(simulationRunner, "RunSimulation", simulation$ref)
+
+  results <- ifNotNull(population,
+                       rClr::clrCall(simulationRunner, "RunSimulation", simulation$ref, population$ref, options$ref),
```
msevestre: I wonder if there is a way to be clever with the `...` here to avoid the duplication. But well it's concise and clear enough

R/utilities-simulation.R
```diff
@@ -70,22 +70,38 @@ saveSimulation <- function(simulation, filePath) {
...
 #' results <- runSimulation(sim)
+#'
+#' # Running a population simulation
+#' popPath <- system.file("extdata", "pop.csv", package = "ospsuite")
+#' population <- loadPopulation(csvPath)
+#' results <- runSimulation(sim, population)
+#'
 #' @export
-runSimulation <- function(simulation) {
+runSimulation <- function(simulation, population = NULL, simulationRunOptions = NULL) {
```
msevestre: One function to support all workflows for now. `runSimulation` runs a simulation. If a population is also passed as a parameter, it will be used for variations

R/utilities-simulation.R
```diff
@@ -70,22 +70,38 @@ saveSimulation <- function(simulation, filePath) {
...
   validateIsOfType(simulation, Simulation)
+  validateIsOfType(population, Population, nullAllowed = TRUE)
+  validateIsOfType(simulationRunOptions, SimulationRunOptions, nullAllowed = TRUE)
+  options <- simulationRunOptions %||% SimulationRunOptions$new()
```
msevestre: need utility methods to return a new object if the first object is null

tests/dev/script.R
```diff
@@ -11,27 +11,32 @@ addOutputs(parameter, simulation = sim)
...
 
-results <- runSimulation(sim)
-paths <- results$allQuantityPaths
+individualResults <- runSimulation(sim)
+paths <- individualResults$allQuantityPaths
+individualPkkAnalyses <- calculatePKAnalyses(individualResults)
 
-# exportResultsToCSV(results, sim, "C:/temp/export/results.csv")
+
+populationResults <- runSimulation(sim, population, simRunOptions)
```
msevestre: Example of call to runSImulation for a population

msevestre:
@PavelBal ok I need to merge that for now. Let's review when you have time

PR #181, opened 2019-09-26
Fixes #24 parameter info

### LINKED ISSUE #24: Basic UseCase: get properties of parameters/ouptuts
get following properties for Parameters/Outputs
- Unit
- isFormula (is this a formula or a static Parameter)
- Formula (string of formula)
- is table Parameter (is this a table Parameter?)
- full path

msevestre:


R/formula.R
```diff
@@ -0,0 +1,58 @@
...
+        private$throwPropertyIsReadonly("formulaString")
+      }
+    }
+  ),
+  public = list(
+    print = function(...) {
+      private$printClass()
+      self$printFormula()
+    },
+    printFormula = function() {
```
msevestre: @KatrinCoboeken @PavelBal 
Here I am only showing one formula state at time. In theory, we could show all of them
e.g.
isConstant: False
isFormula: True,
isTable: false tc..

not sure if this make sense or not

R/quantity.R
```diff
@@ -26,17 +28,40 @@ Quantity <- R6::R6Class(
...
+    formula = function(value) {
+      private$readOnlyProperty("formula", value, private$.formula)
+    },
+    isTable = function(value) {
+      private$readOnlyProperty("isTable", value, self$formula$isTable)
+    },
+    isConstant = function(value) {
+      private$readOnlyProperty("isConstant", value, self$formula$isConstant)
+    },
+    isFormula = function(value) {
```
msevestre: Is  a distributed parameter a formula parameter or a constant parameter... we need to discuss

msevestre: FYI: @Yuri05 
One could also say that at the Quantity Level, a constant is an entity with a formula (constant or distributed) and the a isFormula parmaeter is a quantity with a Table formula or Explicit formula

We can really be has specific or generic as required. I just not sure of the use cases

msevestre:
@KatrinCoboeken, @Yuri05 : No feedback. I am going to merge this. please create a NEW issue if changes are required

PR #183, opened 2019-09-27
178 rename

msevestre:


DESCRIPTION
```diff
@@ -62,7 +62,6 @@ Collate:
     'utilities-pk-analyses.R'
     'utilities-population.R'
     'utilities-quantity.R'
-    'utilities-results.R'
```
msevestre: There was a utilities-results but you had also created a utilties-simulation-results.
I moved everything into one

PR #189, opened 2019-10-04
Fixes #25 get set table parameters

### LINKED ISSUE #25: Basic UseCase: set and get Table Parameter
set and get table parameter

msevestre:


msevestre:
@PavelBal . Trivial changes. Merging .Please review at your earliest convenience

R/quantity.R
```diff
@@ -88,6 +106,9 @@ Quantity <- R6::R6Class(
     },
     allUnits = function() {
       rClr::clrCallStatic(WITH_DIMENSION_EXTENSION, "AllUnitNames", self$ref)
+    },
+    reset = function() {
+      self$isFixedValue <- FALSE
```
PavelBal: @msevestre Can you explain this? What exactly does it mean and when is it set to TRUE`?

msevestre: Setting it to FALSE means that we forget the value set by the user and we use the formula value instead.
Setting it to TRUE should not be done normally by the user. This flag should actually be a readonly flag in my opinion

That's why I have created a reset method instead to hide the implementation details

reset => isFixedValue = FALSE => we use the formula instead of the value that may have been set by the user 

Note. This flag has no meaning for a constant parameter

PR #191, opened 2019-10-04
Fixes #141 use enum

### LINKED ISSUE #141: Use enum for "compareBy"-values in "uniqueEntities()"

msevestre:


R/error-checks.R
```diff
@@ -39,6 +39,14 @@ validateIsOfType <- function(object, type, nullAllowed = FALSE) {
   stop(messages$errorWrongType(objectName, class(object)[1], objectTypes))
 }
 
+validateEnumValue <- function(enum, value) {
```
msevestre: Also added this method that validates the enum value.

R/utilities-entity.R
```diff
@@ -19,16 +30,14 @@
...
-uniqueEntities <- function(entities, compareBy = "id") {
+#' uniqueEntities(parameters, compareBy$id)
+uniqueEntities <- function(entities, compareBy = CompareBy$id) {
   if (is.null(entities)) {
     return(NULL)
   }
 
   validateIsOfType(entities, "Entity")
-  if (!compareBy %in% c("id", "name", "path")) {
-    stop(messages$errorUniqueEntitiesWrongCompareBy)
```
msevestre: no need for a specific message. We have now one way of validating enum

tests/testthat/test-utilities-entity.R
```diff
@@ -28,7 +28,7 @@ test_that("It can filter by name", {
     getParameter(toPathString(c("Organism", "Liver", "Pericentral", "Volume")), sim),
     getParameter(toPathString(c("Organism", "Liver", "Pericentral", "Weight (tissue)")), sim)
   )
-  expect_equal(length(uniqueEntities(parameters, compareBy = "name")), 2)
+  expect_equal(length(uniqueEntities(parameters, CompareBy$name)), 2)
```
msevestre: in fact using string would work too. ("e,g "name would also be valid)

Issue #193, opened 2019-10-07
Update output format to match TLF requirement


msevestre:
FYI: @PavelBal 
The values array is the concatenation of all the individual values!

I have 129 points. 10 indiviudal, my x array has 129 entries. My y array has 1290 entries!

```
outputValues <- getOutputValues(populationResults, populationResults$allQuantityPaths )
t <- outputValues[[1]]
 length(t$x)
[1] 129
> length(t$y)
[1] 1290
```

msevestre:
ok it's because y is a matrix of 129 rows and 10 columns.

msevestre:
@PavelBal Which makes me wonder if this structure is  the right default structure for a population?
 
@Yuri05 Based on the TLF Library, how should we organize our outputs

msevestre:
@abdullahhamadeh What should be according to you the best structure for the output so that we can consume the TLF function almost out of the box to graph standard plots such as mean and percentile of a population simulation?

abdullahhamadeh:
Typically (and perhaps most naturally) you might have this:
![image](https://user-images.githubusercontent.com/18409098/66339078-5d905780-e910-11e9-96b6-4136a49a3bfe.png)

But this isn't the best format for ggplot.  A better format is 

![image](https://user-images.githubusercontent.com/18409098/66339095-64b76580-e910-11e9-9120-dc8b3f663bca.png)

aka tidy format.  Of course instead of experiment number you might have IndividualID.

I believe the reshape2 toolbox can transform the first format to the second, so it's ok if we have the first format, but the user will have the extra step of needing to specify which columns belong to which experiment.

msevestre:
@abdullahhamadeh mean and std etc... will be calculated using some of the function you wrote
What comes out of the population engine is per simulated path, a matric of values (one for each indiviudal). Alternatively, it could be seen as one matrix for each individual( with one column per path)

What you are describing will definitely require to call one of the predefined aggregation function . But how should create the row output from the population engine? Happy to have  a chat

msevestre:
@Yuri05 Do you concur that the format that we return from the getOutputValue should be directy pluggable into the TLF lib?

Yuri05:
> Do you concur that the format that we return from the getOutputValue should be directly pluggable into the TLF lib?

Agree. It would be nice to have outputs directly matching the TLF inputs specifications

abdullahhamadeh:
[...]

PR #197, opened 2019-10-11
WIP #193 update output

msevestre:


R/utilities-simulation-results.R
```diff
@@ -83,6 +83,42 @@ getOutputValues <- function(simulationResults, quantitiesOrPaths, individualIds
   return(output)
 }
 
+#' @export
+getOutputValuesTLF <- function(simulationResults, quantitiesOrPaths, individualIds = NULL) {
```
msevestre: @All: This is just a first try to reorganize the way we export results for a population simulation

@abdullahhamadeh @pchelle 
Is this the type of format that would be expected in order to be able to  plug into the the TLF?

4 Columns: IndivdualId, Time, Path, value

image

R/utilities-simulation-results.R
```diff
@@ -83,6 +83,41 @@ getOutputValues <- function(simulationResults, quantitiesOrPaths, individualIds
...
+
+  # If no specific individual ids are passed, iterate through all individuals
+  individualIds <- ifNotNull(individualIds, unique(individualIds), simulationResults$allIndividualIds)
+
+  # All time values are equal
+  timeValues <- simulationResults$timeValues
+
+  output <- NULL;
+
+  for (individualId in individualIds) {
```
msevestre: This is quite simple here. Not sure if this is not good in terms of memory consumption. I am createing one data frame per {path, individualId" that I merge into the output one

R/utilities-simulation-results.R
msevestre: @Yuri05 FYI: what do you think?

msevestre:
@all: that would be interesting to see how it behaves witha population of 10000 individual and say 10 output calculated over a 2000 points per simulation range....

I'll try and let you know;-)

R/utilities-simulation-results.R
```diff
@@ -83,6 +83,75 @@ getOutputValues <- function(simulationResults, quantitiesOrPaths, individualIds
   return(output)
 }
 
+#' Export values that can be plugged directly into the TLF lib (TODO)
+#'
+#' @export
+getOutputValuesTLF <- function(simulationResults, population, quantitiesOrPaths = NULL, individualIds = NULL) {
```
msevestre: @PavelBal @pchelle @abdullahhamadeh 
This method works but is suboptimal
It takes about 15 sec to retrieve the value for 1000 individuals and one output! 
How can we profile this? Also are there any obvious place were we can optimize?

R/utilities-simulation-results.R
```diff
@@ -83,6 +83,75 @@ getOutputValues <- function(simulationResults, quantitiesOrPaths, individualIds
...
+    quantity <- getQuantity(path, simulationResults$simulation)
+    data <- NULL
+    metaData$Value <- list(unit = quantity$unit, dimension = quantity$dimension)
+    pathColumn <- rep(path, valueLength)
+
+    for (individualId in individualIds) {
+      individualProperties <- individualPropertiesCache$get(toString(individualId))
+      values <- simulationResults$getValuesForIndividual(path, individualId)
+      individualData <- data.frame(individualProperties, Path = pathColumn, Value = values)
+      data <- rbind(data, individualData)
```
msevestre: @PavelBal @pchelle @abdullahhamadeh Ok this is the line that brings the performance down. How to rewrite?

msevestre: ok clearly rbind in for loop is not good. Easy to rewrite though: )

msevestre: https://stackoverflow.com/questions/26843861/replace-rbind-in-for-loop-with-lapply-2nd-circle-of-hell

R/utilities-simulation-results.R
```diff
@@ -83,6 +83,77 @@ getOutputValues <- function(simulationResults, quantitiesOrPaths, individualIds
...
+    quantity <- getQuantity(path, simulationResults$simulation)
+    resultsForPath <- vector("list", length(individualIds))
+    metaData$Value <- list(unit = quantity$unit, dimension = quantity$dimension)
+    pathColumn <- rep(path, valueLength)
+
+    for (individualIndex in seq_along(individualIds)) {
+      individualId <- individualIds[individualIndex]
+      individualProperties <- individualPropertiesCache[[individualIndex]]
+      values <- simulationResults$getValuesForIndividual(path, individualId)
+      resultsForPath[[individualIndex]] <- data.frame(individualProperties, Path = pathColumn, Value = values)
```
msevestre: Removing the individualProperties from the data.frame removes about 1/3 of the time. 
Lol there is clearly a lot that can be optimized here.

msevestre:
@abdullahhamadeh @PavelBal @pchelle 
I'd like some feedback here please.
Do you know how we can optimize the code. I have done some optimization already and profiled the code. Most of the time is spent on two assignments:

Creating the dataframe per individual and path, and creating the combine data.frame per path. This is quite ridiculous to be honest to spend so much time just to create the output structure

image

abdullahhamadeh:
Would it make sense to put an entire individual object in the dataframe under a column called "Individuals" rather than have each individual property as a separate column?  Then if, eg, we need to group by gender we would need to look up the objects' $Gender field.  We'd need to modify our grouping algorithms for this.

abdullahhamadeh:
Ok seems like you cannot coerce an R6 object into a dataframe, but we can represent each individual with a list instead.

msevestre:
> Would it make sense to put an entire individual object in the dataframe under a column called "Individuals"

I don't know if this make sense. But it feels dirty to do that considering that the only reason why we have this structure is to follow the tidy way of organizing data. 

Now, not assigning the properties values to the data.frame drastically updates the time. Still not great, but at least manageable

see line 147
image


The data.frame only has two columns, path and values. At least factor  here

PavelBal:
@msevestre could you share the code and the pop file that results in those profiling results?

msevestre:
@PavelBal 
In order to profile, you need to install package profvis

the population is attached here as a zip file

The code is in scripts.R. Simply update the path accordingly
[Pop_5000.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/files/3757259/Pop_5000.zip)

msevestre:
the simulation used is `S1` under `OSPSuite-R/tests/data`

PavelBal:
You could use:

`data.frame(individualProperties, Path = pathColumn, Value = values, check.names = FALSE, stringsAsFactors = FALSE)` in here:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/fbd0316d239a8e7af765f365cfa84cc5ccde8eb2/R/utilities-simulation-results.R#L147

Speeds up the line from 3700 ms to 2400 ms for my case.

Also, why

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/fbd0316d239a8e7af765f365cfa84cc5ccde8eb2/R/utilities-simulation-results.R#L150

instead of simply 

`data <- rbind(resultsForPath)` ?

msevestre:
why not using factor would make a difference? Isn'it better from a memory point of view?
At any rate, the gain is quite small compare to the overall time. 

The code can probably be replaced with rbind. It was taken from a site trying to optimize data.frame

Why is it so slow to construct data.frame? It's not even a lot of data

msevestre:
Well it does not return the same reults it seems. Not sure what do.call is doing but there is some magic going on

msevestre:
ok do.call(rbind,  c(df1, df2, df3)) is equivalent to rbind(df1, df2, df3) whereas rbind(c(df2, df2, df3)) does nothing. So do.call is required in that case (unless there is a way to transform an list in arguments?

msevestre:
hum using `stringsAsFactors<-FALSE` everywhere brings the time to 15sec... any idea why this is so much better when FALSE?

msevestre:
read page 14 for why this is done like that
http://www.burns-stat.com/pages/Tutor/R_inferno.pdf

@pchelle Why do you think a loop is the problem?

R/cache.R
```diff
@@ -8,8 +8,9 @@ Cache <- R6::R6Class(
   "Cache",
   public = list(
     type = NULL,
-    initialize = function(type) {
+    initialize = function(type = NULL) {
       self$type <- type
+      private$cachedObjects <- new.env(hash = TRUE, parent = emptyenv())
```
msevestre: @PavelBal This was a massive bug here. The cache was only initialized when the class def was created (e.g. global) but not on every instance. Just spent two hours on that :) Glad I found it

PavelBal: Wow :scream:

pchelle:
I was reading this document about time optimization: https://csgillespie.github.io/efficientR/performance.html
Data.frame takes much longer to run since their column variables don't have necessarily the same properties. A first easy optimization which worked well on my tests was to create the data.frame after the IndividualID loop:

image

msevestre:
That's a great idea. Indeed, with all the changes I made, this was not
required anymore
It almost feel that I could return  ALL values for a given path as a giant
double array in fact and we would bypass that for completely.

Cool stuff.  Keep going :)

msevestre:
Ok this change is good enough . We are now in the realm of acceptable time.

msevestre:
@all: I have changed the structure to return a list with two outputs. data, and metaData. 
Let's see if this works of if we need to come back to one data frame per output. Perf are somewhat ok now

image

PR #199, opened 2019-10-14
Use Class instead of strings for isOfType

### LINKED ISSUE #198: Replace all occurences of validateIsOfType to use Class instead of String
validateIsOfType(results, "SimulationResults") -> validateIsOfType(results, SimulationResults)

PavelBal:
Fixes #198

### DIFF
--- a/R/error-checks.R
+++ b/R/error-checks.R
@@ -20,7 +20,6 @@ isOfType <- function(object, type) {
   object <- c(object)
   type <- typeNamesFrom(type)
 
-  paste(type)
   all(sapply(object, function(x) inherits(x, type)))
 }
 
@@ -48,11 +47,13 @@ validateEnumValue <- function(enum, value) {
 }
 
 typeNamesFrom <- function(type) {
-  if (is.character(type)) {
-    return(type)
-  }
   type <- c(type)
-  sapply(type, function(t) t$classname)
+  sapply(type, function(t){
+    if (is.character(t)){
+      return(t)
+    }
+    t$classname
+  })
 }
 
 validateIsString <- function(object, nullAllowed = FALSE) {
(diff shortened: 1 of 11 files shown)


R/error-checks.R
```diff
@@ -20,7 +20,6 @@ isOfType <- function(object, type) {
   object <- c(object)
   type <- typeNamesFrom(type)
 
-  paste(type)
```
PavelBal: I think this was a leftover...

R/error-checks.R
```diff
@@ -48,11 +47,13 @@ validateEnumValue <- function(enum, value) {
...
-  if (is.character(type)) {
-    return(type)
-  }
   type <- c(type)
-  sapply(type, function(t) t$classname)
+  sapply(type, function(t){
+    if (is.character(t)){
+      return(t)
+    }
+    t$classname
```
PavelBal: Otherwise it produces an error when mixing Class and strings (e.g. `validateIsOfType(Quantity, "character")`)

R/error-checks.R
msevestre: yep correct

msevestre:
I also realize that this method does not work with S3 class.
for example in the TLF lib, I copied paste the code. A ggplot plot is a S3 class I believe and it is seen internally as an array. As such the method tries to validate every entry of the class whch is wrong

msevestre:
I had to update the methode as follow
```
isOfType <- function(object, type) {
  if (is.null(object)) {
    return(FALSE)
  }

  type <- typeNamesFrom(type)

  inheritType <-  function(x) inherits(x, type);

  if(inheritType(object)){
    return(TRUE)
  }
  object <- c(object)

  all(sapply(object, inheritType))
}
```

msevestre:
I will create another commit in your branch and merge

PR #207, opened 2019-10-31
205 object not found

### LINKED ISSUE #206: masked object
if ospusite package is loaded after the tlf package was loaded, the following warning comes:

```
Attaching package: ‘ospsuite’
The following object is masked from ‘package:tlf’:
    enum

```

The same happens, if tlf is loaded after osp suite:
```
Attaching package: ‘tlf’
The following object is masked from ‘package:ospsuite’:
    enum

```

msevestre:
Also fixes #206


NAMESPACE
```diff
@@ -8,7 +8,6 @@ export(addOutputs)
 export(calculatePKAnalyses)
 export(clearOutputIntervals)
 export(clearOutputs)
-export(enum)
```
msevestre: @Yuri05 I am not exporting the enaum anymore. That means that I cannot run the tests . This fixes #206

PR #209, opened 2019-11-06
205 object not found

msevestre:
@Yuri05 @PavelBal Going to add the export so that we can try


msevestre:
@Yuri05 Can you try with the build currently created and see if that solves the issue?

Yuri05:
yep, works now!

Issue #205, opened 2019-10-31
'SimulationRunOptions' not found


Yuri05:
```
> library("ospsuite")
> t <- SimulationRunOptions$new()
Error: object 'SimulationRunOptions' not found

```

msevestre:
I can reproduce

Yuri05:
error still there

msevestre:
Make sure you uninstall the package completely and reinstall it.

Cannot reproduce on two installs. Also tests are running instantiating the
object

Yuri05:
>Make sure you uninstall the package completely and reinstall it.

Will try. However with the new version installed above the old: another error (masked enum) has gone, but this error didn't

msevestre:
This is so weird. What is different with this class name...I am not sure.
I might try to rename it to RunOption or something like that and see if if
makes a difference

@pavelbal. Any idea what is going on?

Yuri05:
ok, deinstalled the package and installed again. Still the same error.
I am testing under R **3.5.2** - maybe that's make the difference?

msevestre:
No that's what I have on my machine

Yuri05:
That's the package you are using, correct? https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r/builds/28528174/artifacts (version 1.0.57)

PavelBal:
Well we are not supposed to create objects outside of the package! `Population$new()` or `Simulation$new `also do not work.

Objects should be created from within `utilities-xxx`.

Yuri05:
@PavelBal Can you please adjust the test script https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/tests/dev/script.R accordingly?
Otherwise https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/ba8e002368efd0a1538e0118416fe526b6e8ec19/tests/dev/script.R#L65 crashes

msevestre:
>Well we are not supposed to create objects outside of the package

What? Of course we can create object outside of the package. Why wouldn't we be able to create object?

There is nothing to adjust in the script. Calling new should work. Maybe we need to export the class in that case but I don't understand why since we don't have to export other classes

msevestre:
Simulation$new() does not work because you need a ref to a .net object. The object in question behaves differently

PavelBal:
> What? Of course we can create object outside of the package. Why wouldn't we be able to create object?

[...]

Issue #200, opened 2019-10-14
Default value for number of cores


PavelBal:
We could also get the maximal available number by calling

`parallel:detectCores()`

`parallel` is included in the base R distribution. Then we could set the value to max-1, as it is by default e.g. in PK-Sim.

msevestre:
sure.
It is however `parallel::detectCores()` :)

Issue #179, opened 2019-09-26
Add stopIfNotFound to getResultValues


PavelBal:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/169#discussion_r325844778

msevestre:
this is done right? Please reopen if not

Issue #173, opened 2019-09-19
At the moment, getResults will silently ignore wrong inputs when retrieveing results


msevestre:
see  https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/169#discussion_r325844778 for  a discussion

@KatrinCoboeken @Yuri05 Please feedback

Yuri05:
I would throw an error (at least as per Default) in such a case

msevestre:
I think this is implemented as well

PR #210, opened 2019-11-06
Fixes #133 getXXX: "stopIfNotFound"

### LINKED ISSUE #133: getXXX: "stopIfNotFound"
I propose to add the option to the `getParameter/getContainer` etc. to stop if no entity can be found under the supplied path. 
The argument `stopIfNotFound` is set to `FALSE` by default, and the functions return `NULL` or an empty list. If `stopIfNotFound == TRUE`, the function stops with an error.

I think it is a useful feature as most of the time I definitely **want** to get that parameter/container, and if it is not found it means that the provided path is wrong.

PavelBal:
-implement stopIfNotFound (https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/173 and https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/133 https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/179)
-Change default number of cores to (max - 1) (https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/200)


NAMESPACE
```diff
@@ -3,6 +3,7 @@
 export(CompareBy)
 export(ContainerType)
 export(QuantityType)
+export(SimulationRunOptions)
```
PavelBal: This fixes https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/205

R/simulation-results.R
```diff
@@ -39,10 +39,15 @@ SimulationResults <- R6::R6Class("SimulationResults",
       rClr::clrCall(self$ref, "HasResultsFor", as.integer(individualId))
     },
 
-    getValuesByPath = function(path, individualIds) {
+    getValuesByPath = function(path, individualIds, stopIfNotFound = TRUE) {
       validateIsNumeric(individualIds)
       individualIds <- c(individualIds)
       values <- rClr::clrCall(self$ref, "AllValuesFor", path, as.integer(individualIds))
+
+      if (unique(is.nan(values)) && stopIfNotFound) {
```
PavelBal: Assuming that rClr call returns NaNs if the results is not present.

R/utilities-entity.R
```diff
@@ -24,13 +24,13 @@ CompareBy <- enum(c(
 #' sim <- loadSimulation(simPath)
 #'
 #' parameters <- c(
-#'   getParameter(toPathString(c("Organism", "Liver", "Pericentral", "Volume")), sim),
-#'   getParameter(toPathString(c("Organism", "Liver", "Pericentral", "Volume")), sim),
-#'   getParameter(toPathString(c("Organism", "Liver", "Pericentral", "Weight (tissue)")), sim)
+#'   getParameter(toPathString(c("Organism", "Liver", "Volume")), sim),
```
PavelBal: I had some pain here figuring out why the example would not run! The simple simulation is just too simple :)

R/simulation-results.R
msevestre: Nope. Check the tests. It returns NA

R/utilities-container.R
```diff
@@ -38,21 +38,28 @@ getAllContainersMatching <- function(paths, container) {
...
 #' @export
-getContainer <- function(path, container) {
+getContainer <- function(path, container, stopIfNotFound = TRUE) {
   containers <- getAllContainersMatching(path, container)
   if (length(containers) > 1) {
     stop(messages$errorGetEntityMultipleOutputs(path, container))
   }
 
   if (length(containers) == 0) {
+    if (stopIfNotFound) {
```
msevestre: We have the same exact lines 
```
if (length(containers) > 1) {
    stop(messages$errorGetEntityMultipleOutputs(path, container))
  }

  if (length(containers) == 0) {
    if (stopIfNotFound) {
      stop(messages$errorEntityNotFound(path, container))
    }
    return(NULL)
  }
``` 
3 times.

Can we refactor?

R/simulation-results.R
PavelBal: No, it returns NaN and is converted to NA in the next step. The check for `stopIfNotFound` occurs before that.

R/utilities-container.R
PavelBal: I have moved the `getXXX ` and `getAllXXXMatching` to utilities-entity.

R/simulation-results.R
msevestre: oh I did not see that this check was is IN the object itself. Is there a reason why you don't do the check in the calling method (e.g. the utilities?)

R/utilities-simulation-results.R
```diff
@@ -63,10 +66,7 @@ getOutputValues <- function(simulationResults, quantitiesOrPaths, individualIds
 
     for (individualIndex in seq_along(individualIds)) {
       individualId <- individualIds[individualIndex]
-      vals <- simulationResults$getValuesByPath(path = path, individualIds = individualId)
-      if (is.null(vals)) {
-        next
-      }
+      vals <- simulationResults$getValuesByPath(path = path, individualIds = individualId, stopIfNotFound)
```
msevestre: e.g. here you could do the check. AT that stage, you only have to deal with na no?

R/utilities-entity.R
```diff
@@ -70,3 +77,78 @@ unify <- function(groupEntitiesByPathFunc, paths) {
...
+#' # Returns all `Volume` parameters defined in `Organism` and all its subcontainers
+#' params <- getAllParametersMatching("Organism|**|Volume", sim)
+#' @export
+getAllEntitiesMatching <- function(paths, container, entityType) {
+  # Test for correct inputs
+  validateIsOfType(container, c(Simulation, Container))
+  validateIsString(paths)
+
+  className <- entityType$classname
+  if(length(which(names(ContainerTasks) == className)) == 0){
```
msevestre: Use `validateEnumValue` ?

R/simulation-results.R
PavelBal: Are we going to have one method for getting simulation results, or two - `getOutputValues` and `getOutputValuesTLF`? If it will be only one, I will put the check into one of them. Otherwise having the check here avoids duplication of code.

msevestre: Can't answer that yet. Ideally only one but  who knows

PR #211, opened 2019-11-06
78 update version appveyor

msevestre:


.gitignore
```diff
@@ -2,6 +2,10 @@
 .Rhistory
 .Rapp.history
 
+packages/
+
+inst/lib/
```
msevestre: @PavelBal This is new. Those dll's are not in github anymore. They need to be downloaded by script

README.md
```diff
@@ -9,6 +9,11 @@
 
 # Development tasks
 
+## update packages
+
+- Make sure you have ruby install and that it is available in your path
+- Run `rake postclean` or simply double click on `postclean.bat`. This will update all nuget packages and copy the dependencies in the package `inst/lib` folder.
```
msevestre: @PavelBal Make sure you have ruby install and in your path 
Then simply execute postclean.bat (with R Studio closed) to download all components

Note. you also need to do `git submodule update --init --recursive` do update the submodule

appveyor.yml
```diff
@@ -10,31 +10,38 @@ init:
...
+
+before_build:
+  - nuget sources add -name utility -source https://ci.appveyor.com/nuget/ospsuite-utility
+  - nuget sources add -name serializer -source https://ci.appveyor.com/nuget/ospsuite-serializer
+  - nuget sources add -name funcparser -source https://ci.appveyor.com/nuget/ospsuite-funcparser
+  - nuget sources add -name simmodel -source https://ci.appveyor.com/nuget/ospsuite-simmodel
+  - nuget sources add -name cvodes -source https://ci.appveyor.com/nuget/ospsuite-simmodel-solver-cvodes
+  - nuget sources add -name core -source https://ci.appveyor.com/nuget/ospsuite-core
+  - nuget install packages.config -OutputDirectory packages # cannot use automatic artifact publish because it's not recognized in a C++/CLI project
+  - rake "prepare_for_build[%APPVEYOR_BUILD_VERSION%]"
```
msevestre: This will copy all the dll and also update the package number dynamically

README.md
PavelBal: We will need a clear step-by-step instruction in the readme before the first release. Downloading and compiling the package should be easy.

msevestre: Why do you expect that people will want to compile the package? We will provide zip as tar package as all other R packages not on CRAN.

PR #216, opened 2019-11-08
Fixes #215 update to autofac

### LINKED ISSUE #215: Update to latest OSPSuite.Core and use Autofac instead of Castle

msevestre:


.gitignore
```diff
@@ -4,8 +4,6 @@
 
 packages/
 
-inst/lib/
```
msevestre: @PavelBal I am adding back those dlls. This will simplify getting started, at least on windows. No need to install anything. 
The build script will always update the dlls but this should be equivalent

packages.config
```diff
@@ -1,17 +1,17 @@
...
+  <package id="OSPSuite.Infrastructure.Autofac" version="9.0.32"/>
+  <package id="OSPSuite.R" version="9.0.32"/>
+  <package id="OSPSuite.FuncParser" version="4.0.0.42"/>
+  <package id="OSPSuite.SimModel" version="4.0.0.29"/>
+  <package id="OSPSuite.SimModelSolver_CVODES" version="4.0.0.5"/>
+  <package id="OSPSuite.Serializer" version="3.0.0.1"/>
+  <package id="OSPSuite.Utility" version="4.0.0.1"/>
+  <package id="LumenWorksCsvReader" version="4.0.0"/>
+  <package id="Microsoft.Extensions.Logging.Abstractions" version="2.2.0"/>
+  <package id="Autofac" version="4.9.4"/>
```
msevestre: I have changed a component that was not compatible with .net Standard

PR #221, opened 2019-11-20
159 utilities molecule

msevestre:


msevestre:
@PavelBal This is a WIP. I need to write more tests. But you can already review the code

R/molecule.R
```diff
@@ -0,0 +1,46 @@
...
+#' @docType class
+#' @name Molecule
+#'
+#' @keywords Molecule
+#' @format NULL
+Molecule <- R6::R6Class(
+  "Molecule",
+  inherit = Quantity,
+  active = list(
+    value = function(value) {
```
msevestre: @PavelBal @Yuri05 
Molecules are a different beast. Depending on the project type (amount or concentration) the value is either a constant or a formula But the real start values in for the second case saved in a parameter called Start value. I don't want the user to have to think about it so I basically save a reference to the paraemter start value if if exists and use this parameter to set and retrieve the value. If it does not exist, well it;s a Molecule in amount that can be used AS IS

R/quantity-type.R
```diff
@@ -35,9 +35,8 @@ getQuantityTypeAsString <- function(quantityInt){
   quantityTypes <- which(as.logical(quantityTypes))
   # Convert to integer to get values from the QuantityType-enum.
   # -2 because QuantityType starts at 2 and not 0
-  quantityTypes <- bitwShiftL(2, quantityTypes-2)
+  quantityTypes <- bitwShiftL(2, quantityTypes - 2)
 
   quantityTypes <- lapply(quantityTypes, getEnumKey, QuantityType)
-  #Collapse the separate entries to a string separated by a comma
```
msevestre: @PavelBal I remove this comment because it's obvious. Good job on describing the other lines however

R/utilities-entity.R
```diff
@@ -87,34 +89,19 @@ unify <- function(groupEntitiesByPathFunc, paths) {
 #'
 #' @return A list of entities matching the path criteria coerced to the \code{entityType}.
 #' The list is empty if no entities matching were found.
-#' @examples
```
msevestre: this was copied paste and was wrong

R/utilities-molecule.R
```diff
@@ -0,0 +1,82 @@
...
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' molecule <- getMolecule("Organism|Liver|Intracellular|Drug", sim)
+#' setMoleculeStartValues(molecule, 1)
+#' molecules <- getAllMoleculesMatching("Organism|Liver|Intracellular|Drug", sim)
+#' setMoleculeStartValues(molecules, c(2, 3))
+#' @export
+setMoleculeStartValues <- function(molecules, values) {
+  # Must turn the input into a list so we can iterate through even when only
```
msevestre: @PavelBal that looks a lot like setParameterValues. I am going to create a setQuantitiesVAlues and use it on both places

tests/testthat/test-quantity-type.R
```diff
@@ -1,21 +1,15 @@
...
-test_that("It returns the correct string names of the copmlex quantity types", {
-  #Drug, observer
-  expect_equal(getQuantityTypeAsString(QuantityType$Drug + QuantityType$Observer), "Drug, Observer")
-  #Undefined, Enzyme, Parameter
-  expect_equal(getQuantityTypeAsString(QuantityType$Undefined + QuantityType$Enzyme + QuantityType$Parameter), "Undefined, Enzyme, Parameter")
-})
+context("quantity-type")
+
+test_that("It returns the correct string names of the simple quantity types", {
+  expect_equal(getEnumKey(QuantityType,QuantityType$Drug), getQuantityTypeAsString(QuantityType$Drug))
```
msevestre: @PavelBal I have updated the code. We do not need to test every option exhaustively. We just need to make sure it behaves as expected for a few cases.

PR #223, opened 2019-11-21
Fixes #222 $print method of Quantity and QuantitySelection failing

### LINKED ISSUE #222: $print method of Quantity and QuantitySelection failing
`private$printLine(self$path, getQuantityTypesStringFromInt(self$quantityType))`

should read 

`private$printLine(self$path,getQuantityTypeAsString(self$quantityType))`

Also write tests covering the $print-methods of our classes.

PavelBal:


tests/testthat/test-container.R
```diff
@@ -45,3 +45,8 @@ test_that("It throws an error when trying to set the name of a container", {
   container <- getContainer(liverPath, sim)
   expect_that(container$name <- "name", throws_error())
 })
+
+test_that("It can print container", {
+  container <- getContainer(liverPath, sim)
+  expect_error(container$print(), NA)
```
msevestre: Expect error?? This reads weird. Can you explain?

PavelBal: `expect_error(someCall, NA)` is the way to test for no error being generated in testthat.

msevestre: Ok weird

PR #239, opened 2019-12-13
WIP 72 create individual

msevestre:
@PavelBal This is just the beginning. The code is working but documentation etc.. is required
I don't think we'll be able to unit test this unless we install pksim as part of the CI step, which is potentially an option


R/create-individual.R
```diff
@@ -0,0 +1,87 @@
...
+#'
+#' @export
+createIndividual <- function(originData, useDistribution = FALSE, moleculeOntogenies = NULL) {
+  validateIsOfType(originData, OriginData)
+  validateIsLogical(useDistribution)
+  validateIsOfType(moleculeOntogenies, MoleculeOntogeny, nullAllowed = TRUE)
+  individualFactory <- rClr::clrCallStatic("PKSim.R.Api", "GetIndividualFactory")
+  moleculeOntogenies <- c(moleculeOntogenies)
+
+  individualCharacteristics <-  rClr::clrNew("PKSim.R.Domain.IndividualCharacteristics")
```
msevestre: I have defined in PKSim a NEW object IndividualCharacteristics that is basically the argument used to the function.
Why? because I spent two hours trying to pass array of object from R to .NET and it did not work. So I am creating one object that has an array as member and all is good

R/create-individual.R
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  return(outputList)
+}
+
+getPropertyValues <- function(netObjects, propertyName) {
+  sapply(netObjects, function(x) rClr::clrGet(x, name = propertyName))
+}
+
+createOriginData <- function(
```
msevestre: A helper function that is going to be used to createOriginData 

Note that I am not validating that the species, population etc are in the predefined enum as it would prevent the user to work with user defined species in the PKSim DB. This could easily happen, especially during model validation

R/init-package.R
```diff
@@ -23,8 +22,27 @@ initPackage <- function() {
...
 
-  if (file.exists(libPath)) {
-    Sys.setenv(path = paste(libPath, Sys.getenv("path"), sep = ";"))
+#' Loads the PKSim.R that will enable create individual and create population workflows.
+#' @param pksimFolderPath Path where PK-Sim is installed. If this is not specified, path will be read from registry using the package version
+#'
+#' @note  This will only work on Windows machine and should not be called on any other OS
+#'
+#' @import rClr
+initPKSim <- function(pksimFolderPath = NULL) {
```
msevestre: Here we need to implement something to retrieve the path from regitry. This could be adapted from the code https://github.com/PavelBal/rDotNet_feasibility/blob/master/getPathToPKSimInstallDir.R

However initPKSIm cannot be called on load. This could be called in createIndividual for example (to see)

R/molecule-ontogeny.R
```diff
@@ -0,0 +1,33 @@
+#' @title MoleculeOntogeny
+#' @docType class
+#' @description  Use when retrieving indvidual values using the createIndividualAlgorithm. This class is a simple pair {MoleculeName, Ontogeny}
+#' allowing the user to retrieve potential ontogeny values.
+#'
+#' @field molecule Name of the molecule in the model
+#' @field ontogeny Name of the ontogeny to use for the molecule
+#' @format NULL
+#' @export
+MoleculeOntogeny<- R6::R6Class(
```
msevestre: Object that we have in PKSim. Just  a wrapper

R/origin-data.R
```diff
@@ -0,0 +1,105 @@
+#' @title OriginData
+#' @docType class
+#' @description  Simulation Interval (typically associated with an instace of \code{OutputSchema})
+#'
+#' @field startTime Start time of interval (instance of \code{Parameter})
+#' @field endTime End time of interval (instance of \code{Parameter})
+#' @field resolution Resolution of interval in pts/min (instance of \code{Parameter})
+#' @format NULL
+#' @export
+OriginData <- R6::R6Class(
```
msevestre: Same here. Wrapper around OriginData object

R/origin-data.R
```diff
@@ -0,0 +1,105 @@
...
+    }
+  ),
+  public = list(
+    initialize = function(ref = NULL, value = NULL, unit = NULL) {
+      validateIsNumeric(value, nullAllowed = TRUE)
+      validateIsString(unit, nullAllowed = TRUE)
+      ref <- ref %||% rClr::clrNew("PKSim.Core.Snapshots.Parameter")
+      super$initialize(ref)
+      # Because of weird issue with nullable value in rClr
+      if(!is.null(value)){
```
msevestre: This was also an interesting issue because of the value being a nullable type in .NET. I almost went crazy here

R/pk-sim.R
```diff
@@ -0,0 +1,39 @@
+
+#' Default species defined in PK-Sim
+#'
+#' @include enum.R
+#' @export
+Species <- enum(c(
```
msevestre: Some useful enums so that we have some intellisense

tests/dev/script-create-individual.R
```diff
@@ -0,0 +1,30 @@
...
+
+
+originData <- OriginData$new()
+originData$species <- Species$Human
+originData$population <- HumanPopulation$European_ICRP_2002
+originData$gender <- Gender$Female
+
+print(originData)
+
+originData <- createOriginData(
```
msevestre: Example of usage

R/pk-sim.R
```diff
@@ -0,0 +1,39 @@
...
+#' @export
+HumanPopulation <- enum(c(
+  "Asian_Tanaka_1996",
+  "BlackAmerican_NHANES_1997",
+  "European_ICRP_2002",
+  "Japanese_Population",
+  "MexicanAmericanWhite_NHANES_1997",
+  "Pregnant",
+  "WhiteAmerican_NHANES_1997"
+))
```
Yuri05: Missing **Preterm**

tests/dev/script-create-individual.R
msevestre: @Yuri05 @PavelBal The more I think about it, the more I think we should hide the origin data from the user API

We could expose the `IndividualCharacteristics` object that I had to create in PKSim thast has the originData and the MoleculeOntogeny. That way it can be reused etc.. 
Or we just have a big long list of parameters as we haev in Matlab. Not so nice but why not

R/pk-sim.R
msevestre: fixed

tests/dev/script-create-individual.R
PavelBal: Yes I think it is a good idea to expose `IndividualCharacteristics`.

I don't like the matlab-approach of having these lists of parameters, better to summarize them in an object.

R/init-package.R
PavelBal: Dominik implemented a more sofisticated solution here:
https://github.com/Open-Systems-Pharmacology/R-Toolbox/commit/3d00c2d4c1171aa69d224d2586d5cbce75aa716b

R/create-individual.R
```diff
@@ -0,0 +1,93 @@
...
+}
+
+.createSnapshotParameter <- function(value, unit) {
+  if (is.null(value)) {
+    return(NULL)
+  }
+  return(SnapshotParameter$new(value = value, unit = unit))
+}
+
+createIndividualCharacteristics <- function(
```
msevestre: Ok i have remvoed the concept of OriginData and I am now using the `IndividualCharacteristics` nomenclature

R/create-individual.R
```diff
@@ -0,0 +1,93 @@
...
+                             heightUnit = "cm",
+                             age = NULL,
+                             ageUnit = "year(s)",
+                             gestationalAge = 40,
+                             gestationalAgeUnit = "week(s)",
+                             moleculeOntogenies = NULL) {
+
+  # Assuming that if this function is called directly, PKSim was either initialized already
+  # or should be initialized automatically
+  initPKSim();
```
msevestre: If this was initialized by hand, (for example when using a portable instance) this line won't trigger anything
But if the user uses PKSim installed with the OSP, then this line will initialized automatically

R/ospsuite-env.R
```diff
@@ -36,3 +40,7 @@ ospsuiteEnv$sensitivityAnalysisConfig$variationRange <- 0.1
 # Represents the threshold of sensitivity that we want to achieve.
 # A value of 0.9 will select all parameters contributing to 90% total sensitivity
 ospsuiteEnv$sensitivityAnalysisConfig$totalSensitivityThreshold <- 0.9
+
+# Indicates whether PK-Sim was loaded already. This will prevent unrequired initialization of the pksim assemblies
+ospsuiteEnv$isPKSimLoaded <- FALSE
```
msevestre: I am saving in the environment whether pksim was loaded or not

R/ospsuite-env.R
```diff
@@ -5,6 +5,10 @@ ospsuiteEnv <- new.env(parent = emptyenv())
 # name of the package. This will be used to retrieve information on the package at run time
 ospsuiteEnv$packageName <- "ospsuite"
 
+ospsuiteEnv$suiteName <- "Open Systems Pharmacology"
+
+ospsuiteEnv$packageVersion <- "9.0"
```
msevestre: @PavelBal Any idea if I can read that dynamically from the package?

R/utilities-pksim.R
```diff
@@ -0,0 +1,162 @@
...
+#' Loads the PKSim.R that will enable create individual and create population workflows.
+#' @param pksimFolderPath Path where PK-Sim is installed. If this is not specified, path will be read from registry using the package version
+#'
+#' @note  This will only work on Windows machine and should not be called on any other OS.
+#' This function should also only be called explicitely when using a portable install of the package
+#'
+#' @import rClr
+#' @export
+initPKSim <- function(pksimFolderPath = NULL) {
+  if(ospsuiteEnv$isPKSimLoaded){
```
msevestre: if loaded, do nothing

R/utilities-pksim.R
```diff
@@ -0,0 +1,162 @@
...
+    return()
+  }
+
+  pksimFolderPath <- pksimFolderPath %||% .getPathToPKSimInstallDir()
```
msevestre: otherwise, either use path provided or try to retrieve the path from registry

R/utilities-pksim.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' @return
+#' The path to the PK-Sim installation for version pksim.version or NA if no path could be found.
+#' The path is separated with slashes (unix-style) and in compilance with base-R without a trailing slash.
+#' If more than one matching path is found a warning is produced.
+#'
+#' @examples
+#' path <- .getpathToPKSimInstallDirFromFileSystem("7.4")
+#' path2 <- .getpathToPKSimInstallDirFromFileSystem("7.5", "C:/MyOSPFolder/")
+.getpathToPKSimInstallDirFromFileSystem <- function(pksim.version,
```
msevestre: @PavelBal @Yuri05  Any idea why we need this? This is absolutely ridiculous to not use the registry. Or is this because of possible issue with acccess rights?
If pksim is installed properly, this value is in registry. Otherwise, there are serious problem with the install anyways

I have copied this code from the current R toolbox but I feel that this is not required

PR #250, opened 2020-01-13
248 split population

msevestre:


R/utilities-population.R
```diff
@@ -9,7 +9,30 @@
...
+#' @param outputFileName File names will be constructed using this parameter concatenated with the core index.
+#' @return A tring vector containing the full path of the population files created. Note that there might be less files than cores
+#'
+#' @examples
+#' csvPath <- system.file("extdata", "pop.csv", package = "ospsuite")
+#'
+#' # Split the population in up to 3 files, saved in the temp folder
+#' splitFiles <- splitPopulationFile(csvPath, 3, tempdir(), "PopFile")
+#' @export
+splitPopulationFile <- function(csvPopulationFile, numberOfCores, outputFolder, outputFileName) {
```
msevestre: @abdullahhamadeh This method is now available for you to use. It will split the csv files in `numberOfCores` subfiles (more precisely up to the number) and save the file in the output folder

for example

```
splitPopulationFile(csv, 5, "C:\temp", "popFile") 
```
will create 
C:\temp\popFile_1
C:\temp\popFile_2
etc..

The created files are returned from the function so you know where they are located and how many were created. It can be less than the number of cores. For example if you have 960 Individuals and 50 cores we will end up with 20 Individuals per cores and thus the last 2 cores will have nothing todo.

msevestre:
@PavelBal Please review if you can. I need to commit so that Abdullah can move forward

msevestre:
@abdullahhamadeh This should also fix the issue with wrong indexes. Please let me know if this is not the case

PR #259, opened 2020-01-21
249 unifiy outputs

msevestre:


msevestre:
@abdullahhamadeh @PavelBal @Yuri05 @pchelle 
Breaking changes
* getOutputValues is gone.
* getOutputValuesTLF renamed to getOutputValues 
* loadSimulation does not loadFrom cache by default.

This will be in the next package. Please update asap

PR #262, opened 2020-01-22
Fixes #253 create individual two lists

### LINKED ISSUE #253: Create Individual: Returns an object with two lists
DistributedParameters <- all parameters coming from the PKSim Database based on the individual characteristics
DerivedParameters <- formula parameters with BW, BMI, Q etc..

msevestre:


msevestre:
Merging, I will create the Linux package from this branch

PR #268, opened 2020-01-23
Fixes #267 remove rclr

### LINKED ISSUE #267: Remove rCLR from tools and install from github

msevestre:


msevestre:
nope does not work. It would compile rCLR form source and it asks for mono

PR #276, opened 2020-01-28
252 set scale factor

msevestre:
@PavelBal This is in fact called scale divisor


R/molecule.R
```diff
@@ -24,6 +24,11 @@ Molecule <- R6::R6Class(
           private$.startValue$value <- value
         }
       }
+    },
+    #' @field  scaleDivisor Scale divisor. Its purpose is to reduce numerical noise and to enhance computation performance.
+    #' see https://docs.open-systems-pharmacology.org/working-with-mobi/mobi-documentation/model-building-components#import-molecule-and-parameter-start-values-from-excel
+    scaleDivisor = function(value){
```
msevestre: @PavelBal Link does not work here. Do you know how to make it work (i tried with \link but this is the same

R/utilities-molecule.R
```diff
@@ -69,3 +69,33 @@ setMoleculeInitialValues <- function(molecules, values) {
...
+#' @examples
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' molecule <- getMolecule("Organism|Liver|A", sim)
+#' setMoleculeScaleDivisors(molecule, 0.001)
+#' molecules <- getAllMoleculesMatching("Organism|**|A", sim)
+#' setMoleculeScaleDivisors(molecules, c(0.002, 0.003))
+#' @export
+setMoleculeScaleDivisors <- function(molecules, values) {
```
msevestre: @PavelBal This is the method to use to set scale factors at once

msevestre:
@PavelBal Going to merge to continue. Please review

PR #277, opened 2020-01-28
Fixes #271 update table formula\

### LINKED ISSUE #271: Add getValueAt and addPoints to Table Formula
I would like to have methods to conveniently edit table formulas, with methods like:

-getAllPoints(TableFormula)

-getValueAt(xValue) - returns the value y at point x. If no exact point X exists, the value should be interpolated as it is done in MoBi

-addValues(xVals, yVals) - adds x-y pairs to existing values. If an x value already exists, it will be overwritten.

-setTableValues(xVals, yVals) - removes old values and adds new

msevestre:


R/api-config.R
```diff
@@ -3,6 +3,7 @@
 #' @description  Global configuration for the OSPSuite .NET API
 ApiConfig <- R6::R6Class(
   "ApiConfig",
+  cloneable = FALSE,
```
msevestre: @PavelBal Good to know here. By default it is true and it is possible to clone an object which would most likely be a complete mess with the .NET references. I have disabled it for all object inheriting from .NET wrapper or referencing any .NET object

PR #278, opened 2020-01-28
Fixes #233 issue with unicode

### LINKED ISSUE #233: Issue with HasUnit and BMI
Unit 'kg/m²' is not defined in dimension 'BMI' used by 'BMI'.
but the units available are "kg/m²" "kg/dm²"

msevestre:


R/utilities-units.R
```diff
@@ -48,9 +48,10 @@ toUnit <- function(quantity, values, targetUnit) {
   validateIsOfType(quantity, Quantity)
   validateIsNumeric(values)
   validateHasUnit(quantity, targetUnit)
+  targetUnitIndex <-  as.integer(quantity$unitIndexOf(targetUnit))
   values <- c(values)
   sapply(values, function(value) {
-    rClr::clrCallStatic(WITH_DIMENSION_EXTENSION, "ConvertToUnit", quantity$ref, value, targetUnit)
+    rClr::clrCallStatic(R_WITH_DIMENSION_EXTENSION, "ConvertToUnit", quantity$ref, value,targetUnitIndex)
```
msevestre: I am using the index here so that the string conversion does not happen in rClr but in .NET
Quite crazy but this changes is very much localized . We also always validate that the unit as string exists with `validateHasUnit(quantity, unit)` so we should be good to go

R/quantity.R
```diff
@@ -127,7 +136,14 @@ Quantity <- R6::R6Class(
     #' @param unit Unit to check
     hasUnit = function(unit) {
       validateIsString(unit)
-      rClr::clrCallStatic(WITH_DIMENSION_EXTENSION, "HasUnit", self$ref, unit)
+      unit %in% self$allUnits
```
msevestre: Here we do not use the .NET method. Instead we just check in the local array that is allocated only once

PR #281, opened 2020-01-29
279 update vignettes

msevestre:


msevestre:
Merging. Please update from develop

PR #291, opened 2020-02-03
Fixes #290 issues with normalizer

### LINKED ISSUE #290: Issue on Linux: Path normalizer does not work. Also env variable is different

msevestre:


R/utilities-file.R
```diff
@@ -21,7 +21,14 @@
...
 # Adds the paramter pathToAdd to the system path
 .addPathToSystemPath <- function(pathToAdd) {
-  if (file.exists(pathToAdd)) {
-    Sys.setenv(path = paste(pathToAdd, Sys.getenv("path"), sep = ";"))
+  if (!file.exists(pathToAdd)) {
+    return()
   }
+
+  # Adding path only supported on windows
+  if (.Platform$OS.type != "windows") {
```
msevestre: @Yuri05 Removing this call in Linux as this does not work

Also https://stackoverflow.com/questions/41494585/setting-ld-library-path-from-inside-r

Setting hte path under linux from R won't work
So the path to the lib must be in LD_LIBRARY_PATH or must be set before starting R

R/utilities-file.R
```diff
@@ -5,7 +5,7 @@
   }
 
   # get consistency (unix speparators) with R-base
-  result <- normalizePath(winPath, winslash = "/", mustWork = FALSE)
```
msevestre: This in fact does not work on linux.

msevestre:
@Yuri05 FYI

PR #299, opened 2020-02-05
Fixes #40 create population

### LINKED ISSUE #40: Create (PK-Sim) population
Inputs:
 * Species 
 * Population
 * Min Age (can be empty)
 * Max Age (can be empty)
 * Age Unit
 * Min GestationalAge (can be empty)
 * Max GestationalAge (can be empty)
 * Gestational Age Unit
 * Min Weight (can be empty)
 * Max Weight (can be empty)
 * Weight Unit
 * Min Height (can be empty)
 * Max Height (can be empty)
 * Height Unit
 * Min BMI (can be empty)
 * Max BMI (can be empty)
 * BMI Unit
 * Number of Individuals
 * Proportion of females (in %)
 * List of proteins (can be empty)

Outputs:
 \<Number of Individuals\> parameter sets, incl ontogeny information for the proteins of interest

msevestre:
Very similar to what we have for create individual. 

Flow:
Create PopulationCharacteristics
Create Population from characteristics


tests/dev/script-create-population.R
```diff
@@ -0,0 +1,36 @@
...
+  species = Species$Dog,
+  numberOfIndividuals = 50
+)
+
+print(dog)
+populationDog <- createPopulation(populationCharacteristics  = dog)
+
+print(populationDog)
+
+human <- createPopulationCharacteristics(
```
msevestre: Example of calls here

tests/dev/script-create-population.R
```diff
@@ -0,0 +1,36 @@
...
+  ageMin = 20,
+  ageMax = 30,
+  weightMin = 50,
+  moleculeOntogenies = moleculeOntogeny
+)
+
+print(human)
+
+populationHuman <- createPopulation(populationCharacteristics  = human)
+print(populationHuman)
```
msevestre: output would be
```
PopulationCharacteristics: 
  Species: Human
  Population: Asian_Tanaka_1996
  Number of individuals: 50
  Proportion of females: 100
  Age: [20.00 year(s)..30.00 year(s)]
  Gestational age: ]-Inf..+Inf[
  Weight: [50.00 kg..+Inf[
  Height: ]-Inf..+Inf[
  BMI: ]-Inf..+Inf[
  Molecule MyMolecule with ontogeny CYP3A4
```

R/create-population.R
```diff
@@ -0,0 +1,114 @@
...
+#'
+#' @return An instance of a population object
+#'
+#' @export
+createPopulation <- function(populationCharacteristics) {
+  validateIsOfType(populationCharacteristics, PopulationCharacteristics)
+
+  populationFactory <- rClr::clrCallStatic("PKSim.R.Api", "GetPopulationFactory")
+  population <- rClr::clrCall(populationFactory, "CreatePopulation", populationCharacteristics$ref)
+  Population$new(population)
```
msevestre: This returns the population object that we use everywhere => we can save to csv, etc.. and it will be loaded into pksim. Full circle

R/entity.R
```diff
@@ -40,6 +20,19 @@ Entity <- R6::R6Class(
     #' @field fullPath Same as \code{path}, but with the simulation name. (read-only)
     fullPath = function(value) {
       private$wrapExtensionMethod(EntityExtensions, "EntityPath", "fullPath", value)
+    },
+    #' @field parentContainer Returns a new wrapper instance to the .NET parent container. Multiple call to this method
+    #' will always return the same instance. However two children of the same parent will return two different instances of \code{Container}
+    #' pointing to the same .NET container
+    parentContainer = function(value) {
```
msevestre: Also parentContainer is now supported. Same .NET instance for all children. Different .net instance per child but always the same instance per child. So I think it's not bad

msevestre:
Merging. Please review and create a new issues if something needs to be changed

R/create-population.R
```diff
@@ -0,0 +1,114 @@
...
+  validateIsNumeric(ageMax, nullAllowed = TRUE)
+  validateIsString(ageUnit)
+
+  validateIsNumeric(BMIMin, nullAllowed = TRUE)
+  validateIsNumeric(BMIMax, nullAllowed = TRUE)
+  validateIsString(BMIUnit)
+
+  validateIsNumeric(gestationalAgeMin, nullAllowed = TRUE)
+  validateIsNumeric(gestationalAgeMax, nullAllowed = TRUE)
+  validateIsString(gestationalAgeUnit)
```
Yuri05: GAMin and GAMax must be an integer in [24..40] weeks. 
Also I would allow only "week (s)" as unit for GA.
Can we add this validations?

Yuri05: Actually don'T know if GAMin=null or GAMax=null will work. In PK-Sim we set it explicitly to [24..40] if we want a full range.
In the Matlab implementation of GA we ask OSPSuiteCore for the allowed [Min..Max] Range of GA and set it to this range per default.
https://github.com/Open-Systems-Pharmacology/Matlab-Toolbox-Core/blob/970fcd890cc563c0656d605b092aa3da682959a3/code/auxiliaries/DefaultPopulationSettings.m#L20
https://github.com/Open-Systems-Pharmacology/Matlab-Toolbox-Core/blob/970fcd890cc563c0656d605b092aa3da682959a3/code/auxiliaries/DefaultPopulationSettings.m#L29-L30

tests/dev/script-create-population.R
```diff
@@ -0,0 +1,36 @@
...
+  weightMin = 50,
+  moleculeOntogenies = moleculeOntogeny
+)
+
+print(human)
+
+populationHuman <- createPopulation(populationCharacteristics = human)
+print(populationHuman)
+
+populationHuman$getParameterValues("Organism|Age")
```
Yuri05: please also add a test for Preterm-Pop-Creation

R/create-population.R
msevestre: This uses the min max from the db if not specified. Nothing to change here in my opinion. 
We cannot check at the moment for integer values only

All units allowed for a given dimensions are allowed. So if there is something else than week(s) that's allowed. Why would we change this?

tests/dev/script-create-population.R
msevestre: These are no tests. Just local script for the dev to play with.

There are in fact no unit tests for create individual and population at all because it requires Pksim to be installed. Something we might have to do but it will require messing around with the build script

PR #300, opened 2020-02-06
Add some validations

msevestre:


R/create-population.R
```diff
@@ -88,8 +88,8 @@ createPopulationCharacteristics <- function(
   validateIsNumeric(BMIMax, nullAllowed = TRUE)
   validateIsString(BMIUnit)
 
-  validateIsNumeric(gestationalAgeMin, nullAllowed = TRUE)
-  validateIsNumeric(gestationalAgeMax, nullAllowed = TRUE)
+  validateIsInteger(gestationalAgeMin, nullAllowed = TRUE)
```
msevestre: I have added this new validation method that checks for integer.

R/error-checks.R
```diff
@@ -69,6 +69,22 @@ validateIsNumeric <- function(object, nullAllowed = FALSE) {
   validateIsOfType(object, c("numeric", "integer"), nullAllowed)
 }
 
+validateIsInteger <- function(object, nullAllowed = FALSE) {
+  if (nullAllowed && is.null(object)) {
+    return()
+  }
+
+  if (all(floor(object) == object, na.rm = TRUE)) {
```
msevestre: Very weird way to check for integer. Does not work with the integer keywords otherwise as expected

tests/dev/script-create-population.R
```diff
@@ -34,3 +34,23 @@ populationHuman <- createPopulation(populationCharacteristics = human)
...
 populationHuman$getParameterValues("Organism|Age")
+
+
+preterm <- createPopulationCharacteristics(
+  species = Species$Human,
+  population = HumanPopulation$Preterm,
+  ageMin = 1,
+  ageMax = 10,
+  ageUnit = "day(s)",
+  gestationalAgeMax = 32,
```
msevestre: When no min and max are used, values from the DB are retrieved.

PR #304, opened 2020-02-20
Fixes #244 PK Analysis as data frame

### LINKED ISSUE #244: Add option to return PK analysis as dataframe
Need to extract the PK analysis results from calculatePKAnalyses() as a dataframe.  This dataframe is to be combined with PK parameter results from user-defined functions.

msevestre:
FYI: @abdullahhamadeh


R/utilities-pk-analyses.R
```diff
@@ -23,15 +23,56 @@ calculatePKAnalyses <- function(results) {
...
+  SimulationPKAnalyses$new(pkAnalyses, simulation)
+}
+
+
+#' @title Convert the pk-Analysis to data frame
+#'
+#' @param pkAnalyses pK-Analyses to convert to data frame (typically calculated using \code{calculatePKAnalyses} or imported from file)
+#'
+#' @export
+pkAnalysesAsDataFrame <- function(pkAnalyses) {
```
msevestre: The methods is slightly different and requires the pkAnalysis object

msevestre: but under the hood, it does nothing more as exproted the pkanalysis to file and loading it using your code

R/utilities-pk-analyses.R
```diff
@@ -23,15 +23,56 @@ calculatePKAnalyses <- function(results) {
...
 }
+
+#' @title Loads the pK-analyses from csv file
+#'
+#' @param filePath Full path of the file containing the pK-Analyses to load.
+#' @param simulation Instance of the simulation for which the pk-Analyses were calculated. This is required to verify that the file
+#' matches the simulation
+#'
+#' @export
+importPKAnalysesFromCSV <- function(filePath, simulation) {
```
msevestre: new method that impors a pkanalysis as object from file

PR #311, opened 2020-02-26
273 user defined pk parameters

msevestre:


msevestre:
@abdullahhamadeh Methods added to add dynamic pk parameters (UserDefined PK Parmaeters). Maybe we should call it like that /cc @Yuri05 

@pchelle Utility to convert any numeric array from one dimension to another. Molweight is an optional parameter that can be passed along when converting from Mass to Amount or MassCocnentraiton to MolarConcentraiton

R/utilities-units.R
```diff
@@ -44,15 +45,21 @@ toBaseUnit <- function(quantity, values, unit) {
 #'
 #' valuesInMl <- toUnit(par, c(1, 5, 5), "ml")
 #' @export
-toUnit <- function(quantity, values, targetUnit) {
-  validateIsOfType(quantity, Quantity)
+toUnit <- function(quantityOrDimension, values, targetUnit, molWeight = NULL) {
```
msevestre: @pchelle This is the method to convert values in any type of unit

for example

`toUnit("Mass", c(1, 2, 3), "mol", 20)` will assume that you want to convert the array (1, 2,3) from the base unit of dimension Mass in mol by using the molweight value

We can discuss if this is the apprpriate interface once I see how you need to use it. Please feedback

tests/testthat/test-utilities-units.R
```diff
@@ -9,10 +9,14 @@ test_that("It can convert from a value in base unit to a target unit", {
   expect_equal(toUnit(par, 1, "l"), 1)
 })
 
-test_that("It can convert from a value in base unit to display unit", {
-  expect_equal(toDisplayUnit(par, 1), 1)
+test_that("It can convert from a value in base unit to a target unit using the dimension name", {
+  expect_equal(toUnit(par$dimension, 1, "ml"), 1000)
 })
 
+test_that("It can convert from a value in mass to a value in mol using a molweight parameter", {
```
msevestre: @pchelle Example
Note that MW internally is in kg/umol. You should never have to worry about it as long as you retrieve the value from the model. It wil lbe in the good unit

R/sensitivity-analysis-run-options.R
```diff
@@ -17,8 +17,13 @@ SensitivityAnalysisRunOptions <- R6::R6Class(
                               showProgress = ospsuiteEnv$showProgress) {
       ref <- rClr::clrNew("OSPSuite.R.Domain.SensitivityAnalysisRunOptions")
       super$initialize(ref)
-      self$numberOfCoresToUse <- numberOfCoresToUse
-      self$showProgress <- showProgress
+
+      if (!is.null(numberOfCoresToUse)) {
```
msevestre: @abdullahhamadeh This weird check in the code should not be required anymore. .NET does not like the NULL symbol it seems that we have to wrap that into a if condition

R/utilities-units.R
```diff
@@ -74,3 +81,15 @@ toDisplayUnit <- function(quantity, values) {
...
 }
+
+
+#' Returns the name of all available dimensions defined in the OSPSuite platform
+#'
+#' @examples
+#' dims <- allAvailableDimensions()
+#'
+#' @export
+allAvailableDimensions <- function(){
```
msevestre: @pchelle This also returns all dimensions available in the system. You need to use one of those names when you do conversion if the dimension name is not avialable from the quantity

tests/dev/script-sensitivity-analysis.R
```diff
@@ -5,18 +5,33 @@ sim <- loadSimulation("inst/extdata/simple.pkml", loadFromCache = FALSE)
 outputSelections <- sim$outputSelections
 params <- getAllParametersMatching("**", sim)
 print(params)
+
 sensitivity <- SensitivityAnalysis$new(sim)
-print(sensitivity)
-print(outputSelections)
 sensitivity$addParameterPaths("Organism|Liver|Volume")
+dynamicPkParameter = DynamicPKParameter$new(name = "Test", standardPKParameter =  StandardPKParameter$AucTend)
```
msevestre: @abdullahhamadeh This is how dynamic PK Parameters are added to a SA

R/sensitivity-analysis-run-options.R
abdullahhamadeh: OK I will remove the check in reporting engine, so now a NULL can be passed to SensitivityAnalysisRunOptions initializer for numberOfCoresToUse

PR #318, opened 2020-02-28
Fixes #317 add explorer function

### LINKED ISSUE #317: Add explorer function to traverse a simulation

msevestre:
@abdullahhamadeh This is almost copy paste from your code in Engine
I made  a few modifications:
1/ Removed the retrieveal of path as array from the simulation. This is now using a .net methods 
2/ You were transforming the originalPath to array and then back to string. I am just passing the string in the two inner methods


R/path-explorer.R
```diff
@@ -0,0 +1,73 @@
...
+#' @examples
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#'
+#' tree <- exploreSimulation(sim)
+#'
+#' liver_volume_path <- tree$Organism$Liver$Volume$path
+#'
+#' @export
+exploreSimulation <- function(simulationOrFilePath) {
```
msevestre: I did not like the getEnum name. I am not sure that `exploreSimulation` is the best name at the moment
Suggestion welcome!

PavelBal: `getSimulationTree`?

abdullahhamadeh: Either what @PavelBal suggested or `getSimulationPathsTree`

abdullahhamadeh:
Looks great! Glad it's so much more efficient now :)

msevestre:
Ok. I'll update the method name, and also update the setSimulationParameterValue name

PR #336, opened 2020-03-13
335 add cent os build

msevestre:


msevestre:
@Yuri05 I think it's working

image

PR #338, opened 2020-03-14
333 rename pk parameters

msevestre:


R/pk-parameter.R
```diff
@@ -1,29 +1,70 @@
+#' Standard PK-Parameters types defined in OSPSuite
+#' This is only used to defined how a user defined PK Parameter should be calculated
+#' @include enum.R
+#' @export
+StandardPKParameter <- enum(c(
```
msevestre: @abdullahhamadeh This enum should not be used as I previously stated. This is just used to know "How" a parameter should be calculated e.g. (cCmax, is max of conc, Auc is Aread under the curve etc..)

R/sensitivity-analysis.R
```diff
@@ -84,10 +74,6 @@ SensitivityAnalysis <- R6::R6Class(
       private$printLine("Variation range", self$variationRange)
       parameterLength <- length(private$.parameterPaths)
       private$printLine("Number of parameters to vary", if (parameterLength > 0) parameterLength else "Will be estimated at run time")
-      private$printLine("Dynamic PK Parameters:")
-      for (dynamicPkParameter in self$allDynamicPKParameters) {
```
msevestre: The concept of dynamic pk parameter is all gone now

R/user-defined-pk-parameter.R
```diff
@@ -83,26 +55,24 @@ DynamicPKParameter <- R6::R6Class("DynamicPKParameter",
...
+      self$name <- name
+      self$standardPKParameter <- standardPKParameter
     },
     #' @description
     #' Print the object to the console
     #' @param ... Rest arguments.
     print = function(...) {
       private$printClass()
       private$printLine("Name", self$name)
+      private$printLine("DisplayName", self$displayName)
```
msevestre: A PK Parameter (or user defined pk parameter which is a derived class) has a Name, DisplayName, Dimension and displayUnit
Values hould be exported in the table in the display unit

R/utilities-pk-parameter.R
```diff
@@ -0,0 +1,102 @@
...
+#' userDefinedPKParameter <- addUserDefinedPKParameter$new(name = "MyAUC", standardPKParameter = StandardPKParameter$AucTend)
+#' userDefinedPKParameter$startTime <- 50
+#' userDefinedPKParameter$endTime <- 80
+#'
+#' # Adds a user defined parameter named MyCMax that will calculat the value of Cmax between the 4th and 5th application
+#' userDefinedPKParameter <- addUserDefinedPKParameter$new(name = "MyCMax", standardPKParameter = StandardPKParameter$Cmax)
+#' userDefinedPKParameter$startApplicationIndex <- 4
+#' userDefinedPKParameter$endApplicationIndex <- 5
+#' @export
+addUserDefinedPKParameter <- function(name, standardPKParameter, displayName = NULL, displayUnit = NULL) {
```
msevestre: This is how we create a basic user defined pk parameter

R/utilities-pk-parameter.R
```diff
@@ -0,0 +1,102 @@
...
+#'
+#' @param name Name of PK-Parameter to update
+#' @param displayName Optional display name
+#' @param displayUnit Optional display unit. Note that the unit should be defined in unit of the dimension
+#'
+#' @examples
+#'
+#' updatePKParameter("t_max", "MyTmax", "min")
+#' @export
+updatePKParameter <- function(name, displayName = NULL, displayUnit = NULL) {
```
msevestre: This is how we update an existing pk parameter (either default or user defined)

R/utilities-pk-parameter.R
```diff
@@ -0,0 +1,102 @@
...
+#' @title Returns an instance of a PK-Parameter by name or NULL if the parameter by name is not found
+#'
+#' @param name Name of PK-Parameter to update
+#' @param stopIfNotFound Indicates whether an exception is thrown when the PK-Parameter was not found. Default is \code{TRUE}
+#'
+#' @examples
+#'
+#' pkParameter <- pkParameterByName(name = "t_max")
+#' @export
+pkParameterByName <- function(name, stopIfNotFound = TRUE) {
```
msevestre: retruens a pk parameter by name

R/utilities-pk-parameter.R
```diff
@@ -0,0 +1,113 @@
...
+}
+
+
+#' @title Returns the name of all pk parameters defined in the system
+#'
+#' @examples
+#'
+#' pkParameterNames <- allPKParameterNames()
+#' @export
+allPKParameterNames <- function(){
```
msevestre: Another usefull function @abdullahhamadeh : It returns all PK PArameter Names defined in the system. 
This is what I meant with maybe we do not need an enum . You could verify whatever inpuyt the uer is giving by checking if it is in this array or not

msevestre:
@abdullahhamadeh and @pchelle Going to merge this so that you can try the latest version
Please create issues for everything that is not ergonomic for you and I'll update

R/utilities-pk-parameter.R
abdullahhamadeh: Looks good 👍 
So the pk parameters and SA will be calculated for all elements returned by this?

msevestre: Correct. This is how it should work. You don't have to worry anymore about PKparameter type (user defined or not)

At least... that's the theory ;-)

abdullahhamadeh: I have this list
image

but the pk calculation was done for this subset of the above:
image

msevestre:
Well not all pk parameter apply
Single dosing vs multiple etc.... The list is the superset of all global parameters.

msevestre:
And normalized parameters are not available. It means that the dose per body weight could not be calculated. Is this a real PKSim simulation?

PR #348, opened 2020-03-24
346 better message if pksim not installed

msevestre:


R/utilities-pk-analysis.R
```diff
@@ -69,11 +69,10 @@ pkAnalysesAsDataFrame <- function(pkAnalyses) {
   dataFrame <- tryCatch({
     exportPKAnalysesToCSV(pkAnalyses, pkParameterResultsFilePath)
     pkResultsDataFrame <- read.csv(pkParameterResultsFilePath, encoding = "UTF-8", check.names = FALSE)
-    colnames(pkResultsDataFrame) <- c("IndividualId", "QuantityPath", "Parameter", "Value", "Unit", "Display")
```
msevestre: @pchelle This is now gone. Display name can be accessed directly by using the `pkParameterByName` method

PR #353, opened 2020-03-31
Fixes #352. Fixes #316

### LINKED ISSUE #316: createIndividual: return the unit of the parameter
Could we add an additional output list 'Unit' to each parameter path? Like

```
individual <- createIndividual(individualCharacteristics)
individual$distributedParameters$paths
individual$distributedParameters$values
individual$distributedParameters$units
```

I am creating individual parameter sets and want to save them in an excel file to be loaded in MoBi (this enables to have different individuals in MoBi). The Excel file loaded in MoBi must have the following columns:

| Container Path | Parameter Name | Value | Units |
| ------------- |:-------------:| -----:| ---|

To get the unit of a parameter, I have to pass a simulation - it would be much more convenient to return the units within the `createIndividual`-output.

### LINKED ISSUE #352: Discrepancy between PK-Sim and R in applying individual to simulation
I created an individual with createIndividual() and then applied each of individual$distributedParameters individually to my simulation. When I compare this modified R simulation to an "equivalent" PK-Sim simulation. There is a discrepancy in the simulation results.
image
I did a similar test with an IV formulation, but there didn't seem to be an issue there. So perhaps it has to do with the fact that I am using a table formulation?
image

I wrote a script to compare the parameters listed in the structure text file exported from PK-Sim and the parameters of a simulation. There are discrepancies between some of the parameters, but it's not clear to me whether the differing parameter values are the reason for the discrepancy.
[diff_values.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/files/4405611/diff_values.xlsx)

@msevestre I will send you the R scripts and PK-Sim file via email.

msevestre:


msevestre:
@PavelBal List of units is now returns (new PKSIm is required).  Can you please update the documentation?

PR #358, opened 2020-04-02
Fixes #357 discrepencies mobi

### LINKED ISSUE #357: Different behavior of MoBi and R when an observer evalutates to NaN with the initial conditions
I have an observer for "Gastric retention" of the drug, defined as `M / Total_oral_drug_mass`. As `Total_oral_drug_mass` is 0 at the beginning of the simulation, the value of the observer as NaN. After oral administration (at time = 0), the value of `Total_oral_drug_mass` becomes positive and the observer can be evaluated, which happens in MoBi.

However, in R the value of the observer stays `NaN`. Here is an example of running the simulation directrly exported from MoBi (where the observer works):

```
simulation <- loadSimulation(simPath)
addOutputs("Dapagliflozin|Total oral drug mass", simulation)
addOutputs("Organism|Lumen|Stomach|Dapagliflozin|Gastric retention", simulation)
simulationResults <- runSimulation(simulation = simulation)
getOutputValues(simulationResults,
                quantitiesOrPaths = "Organism|Lumen|Stomach|Dapagliflozin|Gastric retention"
)
Error in simulationResults$getValuesByPath(path, individualIds, stopIfNotFound) : 
  getOutputValues: No results exists for path 'Organism|Lumen|Stomach|Dapagliflozin|Gastric retention' for individual IDs '0'!
```

Setting the value of `Total_oral_drug_mass` to any positive number before running the simulation allows the evaluation of the observer:

```
setParameterValuesByPath( parameterPaths = "Dapagliflozin|Total oral drug mass", values = 1,simulation = simulation)
simulationResults <- runSimulation(simulation = simulation)
getOutputValues(simulationResults,
                quantitiesOrPaths = "Organism|Lumen|Stomach|Dapagliflozin|Gastric retention"
)
$data
    IndividualId Time Organism|Lumen|Stomach|Dapagliflozin|Gastric retention
1              0    0                                           0.0000000000
2              0    1                                           0.9355877042
3              0    2                                           0.8789033294
...
```

msevestre:


R/simulation-results.R
```diff
@@ -52,7 +52,7 @@ SimulationResults <- R6::R6Class(
       individualIds <- c(individualIds)
       values <- rClr::clrCall(self$ref, "AllValuesFor", path, as.integer(individualIds))
 
-      if (unique(is.nan(values)) && stopIfNotFound) {
```
msevestre, line 55: This would only evaluate the first entry in the array. In our case, it would return FALSE, TRUE therefor the error
In fact, we should use all here

msevestre: FYI @Yuri05

PR #360, opened 2020-04-06
Fixes #359 set parameters for one value

### LINKED ISSUE #359: Add a method to set a set of parameters to the same values
in fact, the functikon setParameterValues can do this if the value parameter is a scalar

msevestre:


R/utilities-quantity.R
```diff
@@ -71,17 +71,24 @@ getQuantity <- function(path, container, stopIfNotFound = TRUE) {
...
   # Test for correct inputs
   validateIsOfType(quantities, Quantity)
   validateIsNumeric(values)
-  validateIsSameLength(quantities, values)
+
+  if(length(values) > 1) {
+    validateIsSameLength(quantities, values)
+  }
+  else{
+    values <- rep(values, length(quantities))
```
msevestre, line 90: If we only have one value, set the same value in all quantities. This is useful to update for example all does of a multiple dosing scenario at once for dose escalation

PR #362, opened 2020-04-09
Vignettes update

PavelBal:
I have splitted the vignettes into separate files according to the topic.

-Added information about individual and populations
-Added information about various formula types.

### DIFF
--- a/vignettes/unit-conversion.Rmd
+++ b/vignettes/unit-conversion.Rmd
@@ -1,0 +1,73 @@
+---
+title: "Dimensions and Units"
+output: rmarkdown::html_vignette
+#output: pdf_document
+vignette: >
+  %\VignetteIndexEntry{Dimensions and Units}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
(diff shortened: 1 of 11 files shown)


PavelBal:
OK vignettes are failing because PK-Sim is not found... For "create individual" and "create population" cvignettes.

msevestre:
we'll have to try to install PK-Sim during the build process then

msevestre:
Working on this. I removed the section on initPKSim which was trying to initialize pksim with a path that is not valid

vignettes/unit-conversion.Rmd
```diff
@@ -0,0 +1,73 @@
...
+The values of a certain dimension can be presented in different units - for examlpe, *l* or *ml* for the dimension *Volume*, or *mol* and *µmol* for the dimension *Amount*. The list of all available units is a method of an entity:
+
+```{r $allUnits}
+# Dimension of the parameter
+livParam$dimension
+# Units of the parameter
+livParam$allUnits
+```
+
+Internally, `OSPS` works with the **base units**, and all the values that are shown or passed to functions are in base units by default.
```
msevestre, line 45: Please rename OSPS to OSPSuite

msevestre:
yeah works. That means that we can now add tests for PKSim as well :)

PR #367, opened 2020-04-17
Update OSPSuite-R to support NaN values

msevestre:
@pchelle Please start from the branch 366-study-support to ensure you have the right version of OSPSUite.Core


tests/dev/script-population.R
```diff
@@ -1,8 +1,22 @@
...
 
 df <- populationAsDataFrame(population = population)
+
+
+res <- runSimulation(simulation = sim, population = population)
+
+
+values[2] <- NA
+values[4] <- NaN
+population$setParameterValues("Organism|Lung|Volume", values)
```
msevestre, line 20: @pchelle @Yuri05 Here is a script showing NaN (and even NA) support. Internally, NA are converted into NaN.

tests/testthat/test-population.R
```diff
@@ -44,6 +44,26 @@ test_that("It can add user defined variability using an existing parameter", {
...
+})
+
+test_that("It can add user defined variability using values with NA", {
+  population <- loadPopulation(populationFileName)
+  parameterPath <- "Organism|MyParameterWithNA"
+  values_with_NA <- c(1:10) * 2.5
+  values_with_NA[2] <- NA
+  population$setParameterValues(parameterPath, values_with_NA)
+  expect_true(population$has(parameterPath))
+  expect_identical(population$getParameterValues(parameterPath), values_with_NA)
```
msevestre, line 64: @Yuri05 This is quite suprising to me. The getParameterValues returns NA even though the numbers are stored as NaN. Something happens in rCLR that I do not understand

Yuri05: could be a wrong conversion in rClr...

msevestre: how would it know that one was NA and the other one was NaN. I checked in .NET with console print and we only have NaN
At any rate it works so :)

PR #371, opened 2020-05-13
Fixes change signature of getDimensionForUnit

msevestre:


msevestre:
@Yuri05 @pchelle 
FYI: This methods now returns the name and not the object. This was a mistake

PR #375, opened 2020-05-18
Fixes #374 add enum for available ontogenies

### LINKED ISSUE #374: Add enum for available ontogeny

msevestre:


msevestre:
@PavelBal FYI: I have added an enum `StandardOntogeny` as well as some tests to describe the behavior
Also I have added some comments in the method

In short: by adding MoleculeOntogeny$new("MyMolecule", StandardOntogeny$CYP3A4) this is exactly equivalent as doing

image

I have also restricted the ontogeny name to be one of the predefined ontogeny

PR #377, opened 2020-05-18
Fixes #376 enhance toBaseUnit

### LINKED ISSUE #376: Enhance toBaseUnit to support dimensionName

msevestre:


R/utilities-units.R
```diff
@@ -16,15 +15,22 @@ WITH_DIMENSION_EXTENSION <- "OSPSuite.Core.Domain.WithDimensionExtensions"
 #'
 #' valuesInBaseUnit <- toBaseUnit(par, c(1000, 2000, 3000), "ml")
 #' @export
-toBaseUnit <- function(quantity, values, unit) {
-  validateIsOfType(quantity, Quantity)
+toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL) {
```
msevestre, line 18: @pchelle @Yuri05  the method `toBaseUnit` now supports also a dimension name
A first assumption would be that we can get the dimension name from the unit (with getDimensionForUnit). However this may result in very unexpected result

Example: The dose in saved in umol in core. However your table might have mg.
so a call to getDimensionForUnit(mg) would return the dimension MASS
A call to toBaseUnit(mass, 10, "mg") would return a value in KG not in umol

Therefor the only way to support real unit conversion in this case is to use the simulation (e..g having an instance of a parameter)

msevestre: @Yuri05 @PavelBal  Alternatively, one has to add the Dimension as an entry in the csv table. Then it would work. But this does not seem like an info that the user would really know

Yuri05: Actually the _Dose_ is saved in kg internally; the _Drugmass_ is saved in µmol.
So as long as the user varies the right parameter we should be fine here.
I would not add the dimension indeed.
grafik

It the moment I cannot think of any simulation parameter where we would not get the right dimension just from the unit.
It's different for the molecules start values, which are saved in µmol and user might want to enter them in mg or so. But molecules start values cannot be varied in RE at the moment anyway.

PR #383, opened 2020-05-19
380 wrong dimension use

msevestre:


R/error-checks.R
```diff
@@ -75,6 +75,11 @@ validateIsString <- function(object, nullAllowed = FALSE) {
 }
 
 validateIsNumeric <- function(object, nullAllowed = FALSE) {
+  # Only NA values. It is numeric
+  if(all(is.na(object))){
```
msevestre, line 79: @pchelle @PavelBal Interestingly, this was not working with NA
NA for some reason is seems as a boolean ??

R/utilities-pk-parameter.R
```diff
@@ -25,8 +25,11 @@ addUserDefinedPKParameter <- function(name, standardPKParameter, displayName = N
   validateIsString(displayName, nullAllowed = TRUE)
   validateIsString(displayUnit, nullAllowed = TRUE)
 
+  displayUnit <- enc2utf8(displayUnit %||% "")
+  displayName <- enc2utf8(displayName %||% "")
```
msevestre, line 29: we should always encore string when passing them to R,  especially user inputs

R/utilities-units.R
```diff
@@ -21,11 +21,16 @@ toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL) {
...
   dimension <- quantityOrDimension
+  dimensionTask <- getNetTask("DimensionTask")
+
   if (isOfType(quantityOrDimension, Quantity)) {
     dimension <- quantityOrDimension$dimension
   }
-  values <- c(values)
-  dimensionTask <- getNetTask("DimensionTask")
+
+  if(all(is.na(molWeight))){
```
msevestre, line 30: is.na(NULL) returns an empty array and if(is.na(NULL)) crashes
all(is.na(NULL) returns TRUE, so does all(is.na(NA)) 
We need to assign null after that and we are good to go

Issue #386, opened 2020-05-20
Documentation of "SensitivityAnalysis" mentiones "ospsuiteEnv", which is not exported.


PavelBal:
The documentation of the constructor of `SensitivityAnalysis` mentions `ospsuiteEnv`:

```
...
    #' @param numberOfSteps Number of steps used for the variation of each parameter (optional, default specified in \code{ospsuiteEnv$sensitivityAnalysisConfig})
    #' @param variationRange Variation applied to the parameter (optional, default specified in \code{ospsuiteEnv$sensitivityAnalysisConfig})
...
```

I don't think it is a good idea, as the documentation is visible to the user, but `ospsuiteEnv` is not exported.

msevestre:
sure. Can you change it as part of your documentation update? I think we are using ospsuiteEnv in the doc a few times.

msevestre:
also in
SimulationRunOptions
SensitivityAnalysisRunOptions
SensitivityAnalysis



or should we expose this env? if yes how do we do this?

PavelBal:
Looking at the contents of the env I would not expose the whole env. Actually, Only parameters of the senstivity analysis may be of interest to expose. Maybe write a method that would return the value of a property of the env? Could be useful also e.g. to see which `packageVersion` is defined, but the user will not be able to change anything. E.g.

```
getOSPSuiteSetting <- function(settingName){
return (ospsuiteEnv[[settingName]]
}
```

(well and add a check for key not in the env...)

msevestre:
> Actually, Only parameters of the senstivity analysis may be of interest to expose.

I somewhat disagree with this. These entries are all user-defined parameters

# Default values for the formatNumerics() helper function
ospsuiteEnv$formatNumericsDigits <- 5
ospsuiteEnv$formatNumericsSmall <- 2

# Number of cores to use for simualtions and sensitivity. Default to number of cores on the machine - 1
ospsuiteEnv$numberOfCores <- parallel::detectCores() - 1

# Specificies the default behavior fo progress visualization. By default FALSE
ospsuiteEnv$showProgress <- FALSE

# Default number of steps a new SA instance
ospsuiteEnv$sensitivityAnalysisConfig$numberOfSteps <- 2

# Default variation range for a new SA instance
ospsuiteEnv$sensitivityAnalysisConfig$variationRange <- 0.1

# Represents the threshold of sensitivity that we want to achieve.
[...]

PR #390, opened 2020-05-22
Vignette sensitivity analysis

PavelBal:


R/sensitivity-analysis-results.R
```diff
@@ -57,6 +57,8 @@ SensitivityAnalysisResults <- R6::R6Class("SensitivityAnalysisResults",
     #' @param ... Rest arguments.
     print = function(...) {
       private$printClass()
```
PavelBal, line 59: I think it is a useful addition to the print method.

R/sensitivity-analysis.R
```diff
@@ -26,7 +26,7 @@ SensitivityAnalysis <- R6::R6Class(
     #' @description
     #' Initialize a new instance of the class
     #' @param simulation Simulation for which a sensitivity analysis should be performed
-    #' @param parameterPaths List of parameters to use for sensitivity calculation (optional).If undefined, the sensitivity will be performed automatically
+    #' @param parameterPaths Vector of parameter paths to use for sensitivity calculation (optional).If undefined, the sensitivity will be performed automatically
```
PavelBal, line 29: Fixed the docu regarding list vs vector.

vignettes/create-run-population.Rmd
```diff
@@ -29,7 +29,7 @@ myPopulation <- loadPopulation(csvPopulationFile = popFilePath)
 print(myPopulation)
 ```
 
-## Creating populations {#SecretionCreatePopulation}
```
PavelBal, line 32: Was a typo

vignettes/introduction-ospsuite.Rmd
```diff
@@ -89,3 +90,9 @@ The most important classes are:
     : PK analyses for a simulations result.
   : `QuantityPKParameter`
     : A certain PK-parameter for an output quantity. Has the fields `$name` which is the name of the pk-parameter (e.g. "C_max"), `$quantityPath` the path of the output the parameter has been calculated for, and `$values` the value (or list of values for a population simulation).
+  : `SensitivityAnalysis`
```
PavelBal, line 93: Added class description to intro

vignettes/sensitivity-analysis.Rmd
```diff
@@ -0,0 +1,118 @@
+---
```
PavelBal, line 1: Cover the basics of SA.

R/sensitivity-analysis-results.R
msevestre: ok

vignettes/create-run-population.Rmd
msevestre: lol

PavelBal: Too much modeling... :D

vignettes/sensitivity-analysis.Rmd
```diff
@@ -0,0 +1,118 @@
...
+)
+```
+
+The models built with OSPS depend on a lot of input parameters which are based on literature values, measurements, databases, assumptions. For a given set of input parameters a number of output curves is computed in a simulation. To assess which input parameters have most impact on the output curves, a sensitivity analysis of the simulaiton can be performed. For more information about theoretical background, see [OSPS documentation](https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/sensitivity-analysis).
+
+In brief, the values of the chosen parameters are changed by a certain percentage and the impact of these changes on PK parameters of model outputs is assessed.
+
+## Performing a sensitivity analysis
+The first step of performing a sensitivity analysis is creating an object of the class `SensitivityAnalysis`. At this step, the user can define which parameters should be considered for the sensitivity analysis, in which range the values are variated, and how may steps are performed in one direction (plus and minus). If no parameters are specified, all constant parameters of the simulation will be variated.
+
```
Yuri05, line 24: If no parameters are specified, all constant **and suitable for sensitivity calculations** parameters of the simulation will be variated.

"suitable for sensitivity calculations" means: 
- is not zero
- is **effectively** used in simulation (btw: is this parameter property accessible from R and if yes: is it also explained somewhere?)
- is non-categorical 
- is not marked with "CanBeVaried=false" in the simulation
- is not marked as read-only
- is not a table
@msevestre Anything else?
The list of all such parameters can be get from the simulation by `potentialVariableParameterPathsFor(simulation)`.

vignettes/sensitivity-analysis.Rmd
```diff
@@ -0,0 +1,118 @@
...
+                                                                      "Aciclovir|Fraction unbound (plasma)",
+                                                                      "Organism|Age"))
+print(sa)
+
+# Show which parameters will be variated
+sa$parameterPaths
+```
+
+New parameters can be added to an existing `SensitivityAnalysis` by calling the method `addParameterPaths`. WARNING: If no parameters were specified during creating of a `SensitivityAnalysis` (all constant parameters are considered), calling `addParameterPaths` will make only the manually added parameters being variated.
+
```
Yuri05, line 47: same: all constant and suitable for sensi calculation

vignettes/sensitivity-analysis.Rmd
```diff
@@ -0,0 +1,118 @@
...
+# Create a `SensitivityAnalysis` with all constant parameters
+sa <- SensitivityAnalysis$new(simulation = sim)
+
+# Run the sensitivity analysis
+saResult <- runSensitivityAnalysis(sa)
+print(saResult)
+```
+The method `allPKParameterSensitivitiesFor` returns a list of `PKParameterSensitivity` objets. `PKParameterSensitivity` describes the sensitivity (field `$value`) of a PK-Parameter (`$pkParameterName`) for the output `$outputPath` calculated for the variated parameter `$parameterPath`. The argument `totalSensitivityThreshold` of the method `allPKParameterSensitivitiesFor` is used to filter out the most impactful parameters. A threshold of 0.9 means that only parameters participating to a total of 90 percent of the sensitivity would be returned. A value of 1 would return the sensitivity for all parameters.
+
```
Yuri05, line 81: Is the threshold optional or mandatory in R? If optional: you could mention that 0.9 is also the **default** value

PavelBal: The threshold is optional. I would prefer implementing a method to read the variables of the ospsuiteEnv, where such defaults are defined, as proposed here https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/386

Then we could put a referenrece to it in the documentation.

msevestre: I agree I think we need this

PR #396, opened 2020-05-28
395 expose sa parameter value

msevestre:


R/pk-parameter-sensitivity.R
```diff
@@ -7,8 +7,8 @@ PKParameterSensitivity <- R6::R6Class("PKParameterSensitivity",
   inherit = DotNetWrapper,
   cloneable = FALSE,
   active = list(
-    #' @field parameterPath  Unique name of parameter in sensitivity analysis
-    parameterPath = function(value) {
+    #' @field parameterName  Unique name of parameter in sensitivity analysis
+    parameterName = function(value) {
```
msevestre, line 11: @pchelle @abdullahhamadeh I just realized that what was called parameterPath was actually not a path, but the name as defined in the SA. I have rename this function to be consistent with the rest of the suite. So if you were using explicitly this property, please rename

ALso note that it will already be simplified (normally)
e.g. organism|liver|volume => liver-volume

R/sensitivity-analysis-results.R
```diff
@@ -53,6 +57,16 @@ SensitivityAnalysisResults <- R6::R6Class("SensitivityAnalysisResults",
       toObjectType(pkParameterSentitivities, PKParameterSensitivity)
     },
     #' @description
+    #' Returns the sensisivity value for a given pkParameter, output and parameter. If the sensitivity result does not exist, returns `NaN`
+    #' @param pkParameterName Name of pkParameter for which sensitivity should be retrieved
+    #' @param outputPath Path of the output for which the sensitivity should be retrieved
+    #' @param parameterName Name of the sensitivity parameter for which the sensitivity should be retrieved
+    pkParameterSensitivityValueFor = function(pkParameterName, outputPath, parameterName) {
```
msevestre, line 64: @abdullahhamadeh This is the function that returns the value for the combo {PK, path, parameterName }

R/sensitivity-analysis-results.R
```diff
@@ -53,6 +57,16 @@ SensitivityAnalysisResults <- R6::R6Class("SensitivityAnalysisResults",
       toObjectType(pkParameterSentitivities, PKParameterSensitivity)
     },
     #' @description
+    #' Returns the sensisivity value for a given pkParameter, output and parameter. If the sensitivity result does not exist, returns `NaN`
+    #' @param pkParameterName Name of pkParameter for which sensitivity should be retrieved
+    #' @param outputPath Path of the output for which the sensitivity should be retrieved
+    #' @param parameterName Name of the sensitivity parameter for which the sensitivity should be retrieved
+    pkParameterSensitivityValueFor = function(pkParameterName, outputPath, parameterName) {
+      value <- rClr::clrCall(self$ref, "PKParameterSensitivityValueFor", pkParameterName, outputPath, parameterName)
+      value[is.nan(value)] <- NA
```
msevestre, line 66: returns NA if not found

R/utilities-parameter.R
```diff
@@ -64,6 +64,21 @@ getParameter <- function(path, container, stopIfNotFound = TRUE) {
...
+#' @param simulation A simulation used to find the entities
+#'
+#' @return a display path for each parameter in paths
+#' @examples
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' displayPath <- getParameterDisplayPaths("Organism|Liver|Volume", sim)
+#' @export
+getParameterDisplayPaths <- function(paths, simulation) {
```
msevestre, line 79: @abdullahhamadeh  @pchelle Also this function just in case that returns display path for a list of paths

PR #400, opened 2020-05-29
Fixes #399 export individual simulations

### LINKED ISSUE #399: exportIndividualSimulations creates files without extension
I would expect that `exportIndividualSimulations `creates files with the extension pkml

msevestre:


msevestre:
@Yuri05 FYI

PR #406, opened 2020-06-09
Fixes #405 solver stuff

### LINKED ISSUE #405: Error changing solver properties
When I try to set e.g. `simulation$solver$absTol `or `sim$settings$solver$absTol` the error 
`Property '$solver' is readonly` is shown.
However it seems that the property is set correctly nevertheless (when exporting modified simulation to pkml, AbsTol has the new value.
```
> sim = loadSimulation("C:/MyModel.pkml")
> sim$solver$absTol
[1] 1e-09
> sim$settings$solver$absTol
[1] 1e-09
> sim$solver$absTol=1e-11
Error: Property '$solver' is readonly
> sim$solver$absTol
[1] 1e-11
> sim$settings$solver$absTol
[1] 1e-11
> saveSimulation(sim, "C:/savedModel.pkml")
```       
And: why do we have solver properties **twice** in the simulation?
```
> class(sim$solver)
[1] "SolverSettings" "DotNetWrapper"  "Printable"      "R6"            
> class(sim$settings$solver)
[1] "SolverSettings" "DotNetWrapper"  "Printable"      "R6"     
```

msevestre:


R/simulation.R
```diff
@@ -21,17 +21,11 @@ Simulation <- R6::R6Class(
     path = function(value) {
       private$readOnlyProperty("path", value, self$root$path)
     },
-    #' @field settings SimulationSettings object for the simulation (read-only)
-    settings = function(value) {
```
msevestre, line 25: @Yuri05 I have removed this completely since it seems to cause more trouble. It's for internal use only

R/simulation.R
```diff
@@ -21,17 +21,11 @@ Simulation <- R6::R6Class(
...
-      if (missing(value)) {
-        private$.settings
-      } else {
-        private$throwPropertyIsReadonly("settings")
-      }
-    },
     #' @field solver SimulationSolver object for the simulation (read-only)
     solver = function(value) {
-      private$readOnlyProperty("solver", value, private$.settings$solver)
```
msevestre, line 34: This was the issue. Not sure why but eh... it seems that doing a$b$c is trying to reassign b . weird

Yuri05: R...

PR #409, opened 2020-06-12
Fixes #407 expose application parameters

### LINKED ISSUE #407: Expose application parameters in the simulation
related to https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/223

msevestre:


R/application.R
```diff
@@ -0,0 +1,49 @@
+#' @title Application
+#' @docType class
+#' @description  Application representing one instance of an application in a Simulation
+#'
+#' @format NULL
+Application <- R6::R6Class(
```
msevestre, line 6: @pchelle @abdullahhamadeh @Yuri05 
This is one instance of an application in the model
Parameters are
startTime, infusion, drugmass

starttime and drugmass should always be present. Infusion time maybe

@pchelle in your case you only need start time

This is a parameter object so to get the value you have to do app$starTime$value

R/simulation.R
```diff
@@ -82,7 +84,14 @@ Simulation <- R6::R6Class(
       mw <- rClr::clrCall(self$ref, "MolWeightFor", quantityPath)
       mw %||% NA
     },
-
+    #' @description
+    #' Returns the applications ordered by start time associated to the quantity with path `quantityPath` or an emtpty list if not found
+    #' @param quantityPath Path of quantity used to retrieve the applications (e.g. applications resulting in this quantity being applied)
+    allApplicationsFor = function(quantityPath) {
```
msevestre, line 90: This is how you will get all the application for a given quantity

I do not know if using `quantityPath` makes sense here of if it should be `moleculeName`

This is for the moment using the same approach as for `MoleculeWeight` .Let me know what makes more sense

PR #412, opened 2020-06-16
Fix for detecting cores

PavelBal:


R/ospsuite-env.R
```diff
@@ -20,7 +20,7 @@ ospsuiteEnv$formatNumericsDigits <- 5
 ospsuiteEnv$formatNumericsSmall <- 2
 
 # Number of cores to use for simualtions and sensitivity. Default to number of cores on the machine - 1
-ospsuiteEnv$numberOfCores <- parallel::detectCores() - 1
+ospsuiteEnv$numberOfCores <- function(){parallel::detectCores() - 1}
```
PavelBal, line 23: I think the value was assigned upon building the package. It has to be a function that is called every time the value is asked for.

R/ospsuite-env.R
```diff
@@ -60,5 +60,11 @@ getOSPSuiteSetting <- function(settingName){
     stop(messages$errorOSPSuiteSettingNotFound(settingName))
   }
 
-  return(ospsuiteEnv[[settingName]])
+  obj <- ospsuiteEnv[[settingName]]
```
PavelBal, line 63: A workaround to always return a value, and not a function.

msevestre: ah ah this is kind of badass cool

msevestre:
Fix is missing for SensitivityAnalysisRunOptions. I'll update the PR

R/sensitivity-analysis-run-options.R
```diff
@@ -13,17 +13,13 @@ SensitivityAnalysisRunOptions <- R6::R6Class(
     #' @param numberOfCores Number of cores to use for the simulation. Default value is \code{getOSPSuiteSetting("numberOfCores")}
     #' @param showProgress Should a progress information be displayed. Default value is \code{getOSPSuiteSetting("showProgress")}
     #' @return A new `SensitivityAnalysisRunOptions` object.
-    initialize = function(numberOfCores = ospsuiteEnv$numberOfCores,
-                              showProgress = ospsuiteEnv$showProgress) {
+    initialize = function(numberOfCores = NULL,
```
msevestre, line 16: This was missing here. Also I am differing initialization of number of cores  etc... to the function.
We had some issues in tlf or ospsuite::engine  and this also fixed some issues

PR #417, opened 2020-06-29
Fixes #416 create population should not return derived parameters

### LINKED ISSUE #416: CreatePopulation should not return derived parameters
@Yuri05 The implementation of `createPopulation` in R `https://docs.open-systems-pharmacology.org/working-with-r/r-introduction#creating-populations` follows exactly the one that was previously available in Matlab.

I think this is a mistake. We are returning distributed parameters but also derived parameters such as blood flow etc. If one updates the values for blood flow spec, the blood flows are therefore not updated and one gets wrong results.

I think this should be amended. It should be done in PKSim but this would require a new version. I'd suggest to patch ospsuite-r instead to remove all derived parameters

msevestre:
The population returned by createPopulation also contains formula parameters which is not the expected behavior
This workaround fixes the issue by returning a list with two objects
1/a population (what was returned before)
2/a list of path values to see the value of all derived parameters at TIME OF CREATION of the population

Updating the population object will obviously not update the derived parameters but at least we have a good separation of concern

### DIFF
--- a/R/create-individual.R
+++ b/R/create-individual.R
@@ -73,7 +73,7 @@ createIndividualCharacteristics <- function(
                                             species,
                                             population = NULL,
                                             gender = NULL,
-                                            weight,
+                                            weight = NULL,
                                             weightUnit = "kg",
                                             height = NULL,
                                             heightUnit = "cm",
@@ -90,7 +90,7 @@ createIndividualCharacteristics <- function(
   validateIsString(species)
   validateIsString(population, nullAllowed = TRUE)
   validateIsString(gender, nullAllowed = TRUE)
-  validateIsNumeric(weight)
+  validateIsNumeric(weight, nullAllowed = TRUE)
   validateIsString(weightUnit)
   validateIsNumeric(height, nullAllowed = TRUE)
   validateIsString(heightUnit)
--- a/R/create-population.R
+++ b/R/create-population.R
@@ -3,15 +3,50 @@
 #' @param populationCharacteristics Characteristics of the population to create as an instance of \code{OriginData}
 #' that are actually distributed parameters
 #'
-#' @return An instance of a population object
+#' @return An list with two entries:
+#' The \code{population} An instance of a population object.
+#' The \code{derivedParameters} containing the parameter values modified indirectly by the algorithm. Those parameters are typically formula parameters.
 #'
 #' @export
 createPopulation <- function(populationCharacteristics) {
   validateIsOfType(populationCharacteristics, PopulationCharacteristics)
 
   populationFactory <- rClr::clrCallStatic("PKSim.R.Api", "GetPopulationFactory")
-  population <- rClr::clrCall(populationFactory, "CreatePopulation", populationCharacteristics$ref)
-  Population$new(population)
+  netPopulation <- rClr::clrCall(populationFactory, "CreatePopulation", populationCharacteristics$ref)
+  population <- Population$new(netPopulation)
+
+  individualCharacteristics <- NULL
+
+  #NOTE THIS IS A WORKAROUND UNTIL THE CODE IN PKSIM IS UPDATED
+  if (populationCharacteristics$species == Species$Human) {
+    # create an individual with similar properites Species and population. WEIGHT AND AGE DO NOT MATTER as long as we can create an invidiual
+    individualCharacteristics <- ospsuite::createIndividualCharacteristics(
+      species = populationCharacteristics$species,
+      population = populationCharacteristics$population,
+      weight = 70,
+      age = 30
+    )
+  }
+  else {
+    # create an individual with similar properites Species and population. WEIGHT AND AGE DO NOT MATTER as long as we can create an invidiual
+    individualCharacteristics <- ospsuite::createIndividualCharacteristics(
+      species = populationCharacteristics$species,
+      population = populationCharacteristics$population,
+      weight = populationCharacteristics$weightMin
+    )
+  }
+
+  individual <- createIndividual(individualCharacteristics = individualCharacteristics)
+
+  derivedParameters <- list()
+  for (derivedParameterPath in individual$derivedParameters$paths) {
+    if (population$has(derivedParameterPath)) {
+      derivedParameters[[derivedParameterPath]] <- population$getParameterValues(derivedParameterPath)
+      rClr::clrCall(population$ref, "Remove", derivedParameterPath)
+    }
+  }
+
+  return(list(population = population, derivedParameters = derivedParameters))
 }
 
 #' Creates the population characteristics used to create a population
--- a/R/create-population.R
+++ b/R/create-population.R
@@ -42,7 +42,7 @@ createPopulation <- function(populationCharacteristics) {
   for (derivedParameterPath in individual$derivedParameters$paths) {
     if (population$has(derivedParameterPath)) {
       derivedParameters[[derivedParameterPath]] <- population$getParameterValues(derivedParameterPath)
-      rClr::clrCall(population$ref, "Remove", derivedParameterPath)
+      population$remove(derivedParameterPath)
     }
   }
 
--- a/R/create-population.R
+++ b/R/create-population.R
@@ -17,13 +17,12 @@ createPopulation <- function(populationCharacteristics) {
 
   individualCharacteristics <- NULL
 
-  #NOTE THIS IS A WORKAROUND UNTIL THE CODE IN PKSIM IS UPDATED
+  # NOTE THIS IS A WORKAROUND UNTIL THE CODE IN PKSIM IS UPDATED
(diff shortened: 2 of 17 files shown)


R/create-population.R
```diff
@@ -3,15 +3,50 @@
...
+  else {
+    # create an individual with similar properites Species and population. WEIGHT AND AGE DO NOT MATTER as long as we can create an invidiual
+    individualCharacteristics <- ospsuite::createIndividualCharacteristics(
+      species = populationCharacteristics$species,
+      population = populationCharacteristics$population,
+      weight = populationCharacteristics$weightMin
+    )
+  }
+
+  individual <- createIndividual(individualCharacteristics = individualCharacteristics)
```
msevestre, line 39: This code is required here to retrieve the list of derived parameters from the individual. We therefor make  sure that everything is consistent

vignettes/create-run-population.Rmd
```diff
@@ -52,7 +52,8 @@ populationCharacteristics <- createPopulationCharacteristics(
 print(populationCharacteristics)
 
 # Create population from population characteristics
-myPopulation <- createPopulation(populationCharacteristics = populationCharacteristics)
+result <- createPopulation(populationCharacteristics = populationCharacteristics)
+myPopulation <- result$population
```
msevestre, line 56: All: This is the change in API. Instead of getting a population object directly, one needs to get the $population object from the result.

msevestre:
and + Coverage    80.23%   87.34%   +7.11%      :)

R/create-population.R
```diff
@@ -3,15 +3,50 @@
...
+    )
+  }
+  else {
+    # create an individual with similar properites Species and population. WEIGHT AND AGE DO NOT MATTER as long as we can create an invidiual
+    individualCharacteristics <- ospsuite::createIndividualCharacteristics(
+      species = populationCharacteristics$species,
+      population = populationCharacteristics$population,
+      weight = populationCharacteristics$weightMin
+    )
+  }
```
Yuri05, line 37: why do we need this differentiation between human and not-human here?
`individualCharacteristics `defined as in lines 24-27 would work also for animals, or not?

vignettes/create-run-population.Rmd
Yuri05: Why not just extend the population R class with the `DerivedParametersPaths `and `DerivedParametersValues `instead?

msevestre: because this would mean changing everything in core. AT the moment, a population is what is being varied. The concept of derived etc.. does not exist. It is also possible to override a formula with some values so this would create a serious headache. I don't think this is something we would want to do

R/create-population.R
msevestre: maybe but I do not want ot create a 70kg mouse. I'd rather use the default from the DB.

vignettes/create-run-population.Rmd
Yuri05: I disagree.
You could just add `derivedParameters` property to the **population R6-class only** (the property would be **read/write for now**. As soon as the core is adjusted, it can be made read-only). 
So instead of `return(list(population = population, derivedParameters = derivedParameters))` I would just do
```
population.derivedParameters = derivedParameters
return population
```
Then no API change now (and no API change in the future when the core is adjusted).

msevestre: We are going to have to agree to disagree. While this would work, for the population object, this is also the object that is being passed to runSimulation. 
Or do you suggest to create a separate list in the population object?

msevestre: >population.derivedParameters = derivedParameters

I see. The problem with this is that we have to add much more code. For example, we need to verify that a parameter is not in the two lists, etc. 

For simple consistency, because createIndividual returns a list with two items, I think createPopulation should do the same

msevestre: Let's discuss it tomorrow. I don't undersnad what readonly means for example. one should be allowed to overwrite this stuff.
So let's think about what we would like to have from an API point of view and we can update

R/create-individual.R
```diff
@@ -90,7 +90,7 @@ createIndividualCharacteristics <- function(
   validateIsString(species)
   validateIsString(population, nullAllowed = TRUE)
   validateIsString(gender, nullAllowed = TRUE)
-  validateIsNumeric(weight)
+  validateIsNumeric(weight, nullAllowed = TRUE)
```
msevestre, line 93: @Yuri05 In fact it was not required indeed. MeanBW is used in this case

R/create-population.R
msevestre: age is required for human however.

R/create-population.R
```diff
@@ -3,15 +3,64 @@
...
+
+  individual <- createIndividual(individualCharacteristics = individualCharacteristics)
+
+  derivedParameters <- list()
+
+  # Even though those parameters are derived parmaeters, we keep them in the population for consistency purpose with the PKSim export.
+  standardDerivedParametersToKeep <- c(StandardPath$Weight, StandardPath$BMI, StandardPath$BSA)
+
+  for (derivedParameterPath in individual$derivedParameters$paths) {
+    if (derivedParameterPath %in% c(StandardPath$Weight, StandardPath$BMI, StandardPath$BSA)) {
```
msevestre, line 45: Keeping those guys

R/create-population.R
```diff
@@ -3,15 +3,64 @@
...
+    }
+
+    if (population$has(derivedParameterPath)) {
+      derivedParameters[[derivedParameterPath]] <- population$getParameterValues(derivedParameterPath)
+      population$remove(derivedParameterPath)
+    }
+  }
+
+  # other parameters to remvove that should not have been exported in the first place
+  standardDistributedParametersToRemove <- c(toPathString(StandardContainer$Organism, "MeanBW"), toPathString(StandardContainer$Organism, "MeanHeight"))
```
msevestre, line 56: Also removing this. 
Man the ONE CODE THAT I DID NOT CHECK from the old matlab toolbox is full of crap. Crazy

R/create-population.R
PavelBal: The requirement for age has been added at some point - it was possible to create an individual without specifying the age. Why is it mandatory now?

msevestre: It's been like that in PKSim.Matlab for a while.  What is the use case of creating an individual without knowing the age?

R/create-population.R
PavelBal, line 45: Why not `%in% standardDerivedParametersToKeep`? You are repeating the code. Also, where is `standardDerivedParametersToKeep` actually used?

vignettes/create-run-population.Rmd
PavelBal: > > population.derivedParameters = derivedParameters
> 
> I see. The problem with this is that we have to add much more code. For example, we need to verify that a parameter is not in the two lists, etc.
> 
> For simple consistency, because createIndividual returns a list with two items, I think createPopulation should do the same

But `createIndividual` returns two lists of parameters. `createPopulation` would return a population-object and a list of parameters - rather inconvenient for me.

R/create-population.R
PavelBal: Creating a "standard" individual :) It did definitely work without specifying the age, I had it in one of our scripts and had to update it yesterday as I used it for the first time with the latest release of osps-R.

vignettes/create-run-population.Rmd
msevestre: Why is it iconvenenient? Those derived parmaeters are for "info" only and are stale as soon as you work with the population parameters (in a way that would change those derived parameters). I think that this is the best way to not have inconsistencies

R/dot-net-wrapper.R
```diff
@@ -21,6 +21,10 @@ DotNetWrapper <- R6::R6Class(
     #' @return A new `DotNetWrapper` object.
     initialize = function(ref) {
       self$ref <- ref
+    },
+    finalize = function() {
+      # maybe dispose should be called to if available.
+      self$ref <- NULL
```
msevestre, line 27: @Yuri05 As discussed :)

appveyor.yml
```diff
@@ -12,7 +12,7 @@ install:
   - git submodule update --init --recursive
 
 environment:
-  app_version: "9.0"
+  app_version: "9.1"
```
msevestre, line 15: Also bump version to 9.1 so that we can load PKSim correctly

msevestre:
@all. I have updated the version to 9.1 to be compatible with the latest suite coming out soon. That way, we are consistent if people stick with 9.0

R/create-population.R
PavelBal: @msevestre I think this has not been resolved...

PR #449, opened 2020-11-30
446 ensure that object ids are reset

msevestre:


R/utilities-simulation.R
```diff
@@ -35,7 +38,7 @@
 #' setParameterValues(parameters = parameter3, values = 1)
 #' parameter2$value == parameter3$value # FALSE#'
 #' @export
-loadSimulation <- function(filePath, loadFromCache = FALSE, addToCache = TRUE) {
+loadSimulation <- function(filePath, loadFromCache = FALSE, addToCache = TRUE, resetIds = TRUE) {
```
msevestre, line 41: I have added this parmaeter here. Default is true so by default, ids should be resetted. I do now know of a use case where one would absolutely want to keep the ids but I am keeping it just to be compatible

rakefile.rb
```diff
@@ -16,7 +16,7 @@
 
   update_package_version(product_version)
 
-  install_pksim()
+  install_pksim('feature/relative_expression_redesign')
```
msevestre, line 19: This will have to be changed to develop once pksim is merged into develop

msevestre (commented):
@PavelBal here you have it
v10 however.

PavelBal (approved):
Looks good.

R/output-schema.R
```diff
@@ -14,7 +14,7 @@ OutputSchema <- R6::R6Class(
         intervals <- rClr::clrGet(self$ref, "IntervalsAsArray")
         toObjectType(intervals, Interval)
       } else {
-        private$throwPropertyIsReadonly("solver")
+        private$throwPropertyIsReadonly("intervals")
```
msevestre, line 17: @PavelBal fixed

R/output-schema.R
```diff
@@ -51,8 +51,9 @@ OutputSchema <- R6::R6Class(
     },
 
     #' @description
-    #' Adds the time points to the schema
-    #' @param timePoints Time points to add to the interval
+    #' Adds the time points to the schema. Note that time points and intervals exists concurrently.
```
msevestre, line 54: @PavelBal Fixed

msevestre:
@PavelBal @Yuri05  Well this proves to be much mroe challenging than expected. 
There is a conflict with some dll versions. We are using an higher version of some logging components that was updated by Abdel and Georgios a while back. This seems to create some conflicts at the moment. I am not quite sure what to do. I don't even think that downgrading is an option.

PR #454, opened 2020-12-03
447 remove organism

msevestre:


msevestre:
@Yuri05 This just updates CORE and adds a test. Going to merge this

msevestre:
actually the build is failing because we have global parameters for reactions
For instance R1-k1 is now becoming k1 (we are using the display name as key for the sensitive analysis items)
so I think we should only remove if the entry is oranism in fact for now

Issue #462, opened 2020-12-22
Performance: avoid creation of new Quantity-objects in getOutputValues


PavelBal:
In one particular scenario I have identified a bottleneck in the performance of retrieving the output values of simulated results.

I am simulating a steady-state of the model. For this, I first retrieve all quantities that are described by an ODE - first all molecules, and then all parameters and sorting the parameters that are state variable:

```
  quantities <- getAllMoleculesMatching("Organism|**", container = simulation)
  # Get all state variable parameters
  allParams <- getAllParametersMatching("**", container = simulation)
... extract param$isStateVariable ...
```

This already takes almost 4 minutes, and the majority of time is spent on creating the quantities, in particular in creating the quantities' formula objects:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/44d8471562ff53db5fd7134f26f9d3237766a36b/R/quantity.R#L93-L97

Once identified all quantities (lets call them `allQuantities`), these are used as outputs in the simulation.  Simulating and getting the results consumes 53 seconds, of which only 20 seconds are for the simulation itself - the other 30 seconds are spent on getting the results. Once again, much time is spent on creating `Quantity`-objects, including here:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/44d8471562ff53db5fd7134f26f9d3237766a36b/R/utilities-simulation-results.R#L81-L85

However, creating Quantity-objects is not necessary if a list of quantities is provided instead of paths. By re-writing the method I could improve the time from 53 to 40 seconds - which is not negligible if the scenario is performed multiple times (e.g. in a parameter identification).

I will submit a PR.

msevestre:
>This already takes almost 4 minutes

That I find very surprising. How large is your model?
Also `getAllParametersMatching(**)` looks like you want all parameters. Don't we have a more optimized function for this?

PavelBal:
> That I find very surprising. How large is your model?

Yes this is crazy. Well, the model is quite large - the exported pkml is 50mb...

`getAllMoleculesMatching("Organism|**", container = simulation)` takes 20 seconds, with quite some time spent on creating/accessing the formulas (returns 3800 elements):

![image](https://user-images.githubusercontent.com/25903040/102897026-a70e8900-4467-11eb-84c4-ca2d42025028.png)


And  `getAllParametersMatching(**)` is ridiculously slow (returns 37500 elements) with 3.6 minutes!. 

[...]

PR #463, opened 2020-12-22
Do not create new quantity objects if utilities-simulation-results if…

PavelBal:
… quantites are passed.

See https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/462

### DIFF
--- a/R/dot-net-wrapper.R
+++ b/R/dot-net-wrapper.R
@@ -59,6 +59,17 @@ DotNetWrapper <- R6::R6Class(
       }
     },
 
+    wrapExtensionMethodCached = function(typename, methodName, propertyName, cachedValue, value) {
+      if (missing(value)) {
+        if (is.null(cachedValue)) {
+          return(rClr::clrCallStatic(typename, methodName, self$ref))
+        }
+        return(cachedValue)
+      } else {
+        private$throwPropertyIsReadonly(propertyName)
+      }
+    },
+
     readOnlyProperty = function(propertyName, value, returnValue) {
       if (missing(value)) {
         returnValue
--- a/R/entity.R
+++ b/R/entity.R
@@ -14,11 +14,9 @@ Entity <- R6::R6Class(
     .path = NULL
   ),
   active = list(
-    #' @field path The path of the entity in the container hiearchy without the simulation name. (read-only)
+    #' @field path The path of the entity in the container hierarchy without the simulation name. (read-only)
     path = function(value) {
-      if (is.null(private$.path)) {
-        private$.path <- private$wrapExtensionMethod(EntityExtensions, "ConsolidatedPath", "path", value)
-      }
+      private$.path <- private$wrapExtensionMethodCached(EntityExtensions, "ConsolidatedPath", "path", private$.path, value)
       return(private$.path)
     },
     #' @field fullPath Same as \code{path}, but with the simulation name. (read-only)
--- a/R/entity.R
+++ b/R/entity.R
@@ -10,12 +10,16 @@ Entity <- R6::R6Class(
   cloneable = FALSE,
   inherit = ObjectBase,
   private = list(
-    .parentContainer = NULL
+    .parentContainer = NULL,
+    .path = NULL
   ),
   active = list(
     #' @field path The path of the entity in the container hiearchy without the simulation name. (read-only)
     path = function(value) {
-      private$wrapExtensionMethod(EntityExtensions, "ConsolidatedPath", "path", value)
+      if (is.null(private$.path)) {
+        private$.path <- private$wrapExtensionMethod(EntityExtensions, "ConsolidatedPath", "path", value)
+      }
+      return(private$.path)
     },
     #' @field fullPath Same as \code{path}, but with the simulation name. (read-only)
     fullPath = function(value) {
--- a/R/quantity.R
+++ b/R/quantity.R
@@ -17,7 +17,10 @@ Quantity <- R6::R6Class(
     },
     #' @field unit The base unit in which the quantity value is defined (Read-Only)
     unit = function(value) {
-      private$wrapExtensionMethod(WITH_DIMENSION_EXTENSION, "BaseUnitName", "unit", value)
+      if (is.null(private$.unit)) {
+        private$.unit <- private$wrapExtensionMethod(WITH_DIMENSION_EXTENSION, "BaseUnitName", "unit", value)
+      }
+      return(private$.unit)
     },
     #' @field displayUnit The unit in which the quantity value is usually displayed (Read-Only)
     displayUnit = function(value) {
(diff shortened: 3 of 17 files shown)


R/utilities-simulation-results.R
```diff
@@ -34,15 +34,23 @@ getOutputValues <- function(simulationResults,
   quantitiesOrPaths <- quantitiesOrPaths %||% simulationResults$allQuantityPaths
   quantitiesOrPaths <- c(quantitiesOrPaths)
 
-  # If quantities are passed, get their paths.
-  paths <- quantitiesOrPaths
-  if (isOfType(paths, Quantity)) {
-    paths <- unlist(lapply(paths, function(x) x$path))
+  if (length(quantitiesOrPaths) == 0) {
```
PavelBal, line 37: Better check as soon as possible.

R/utilities-simulation-results.R
```diff
@@ -34,15 +34,23 @@ getOutputValues <- function(simulationResults,
...
-  if (isOfType(paths, Quantity)) {
-    paths <- unlist(lapply(paths, function(x) x$path))
+  if (length(quantitiesOrPaths) == 0) {
+    return(list(data = NULL, metaData = NULL))
   }
-  paths <- unique(paths)
 
-  if (length(paths) == 0) {
-    return(list(data = NULL, metaData = NULL))
+  # If quantities are passed, get their paths.
```
PavelBal, line 41: Ensure that we have a list of paths and a list of corresponding quantities. If quantities are provided, get their paths, and vice versa.

R/utilities-simulation-results.R
```diff
@@ -77,12 +80,17 @@ getOutputValues <- function(simulationResults,
   # Cache of all individual properties over all individual that will be duplicated in all resulting data.frame
   allIndividualProperties <- do.call(rbind.data.frame, c(individualPropertiesCache, stringsAsFactors = FALSE))
 
-
-  for (path in paths) {
-    quantity <- getQuantity(path, simulationResults$simulation, stopIfNotFound = stopIfNotFound)
-    metaData[[path]] <- list(unit = quantity$unit, dimension = quantity$dimension)
-    values[[path]] <- simulationResults$getValuesByPath(path, individualIds, stopIfNotFound)
-  }
+  values <- lapply(paths, function(path){
```
PavelBal, line 83: Better to use lapply instead of iterating.

msevestre: it it? you are lapply twice now instead of one loop.

R/utilities-simulation-results.R
```diff
@@ -34,15 +34,23 @@ getOutputValues <- function(simulationResults,
...
 
-  if (length(paths) == 0) {
-    return(list(data = NULL, metaData = NULL))
+  # If quantities are passed, get their paths.
+  if (isOfType(quantitiesOrPaths, Quantity)) {
+    quantities <- uniqueEntities(quantitiesOrPaths)
+    paths <- unlist(
+      lapply(quantities, function(x) x$path)
+      )
+    names(quantities) <- paths
```
msevestre, line 47: That can be moved outside of if then since you are doing it everywhere

R/utilities-simulation-results.R
```diff
@@ -34,15 +34,23 @@ getOutputValues <- function(simulationResults,
...
+    return(list(data = NULL, metaData = NULL))
   }
-  paths <- unique(paths)
 
-  if (length(paths) == 0) {
-    return(list(data = NULL, metaData = NULL))
+  # If quantities are passed, get their paths.
+  if (isOfType(quantitiesOrPaths, Quantity)) {
+    quantities <- uniqueEntities(quantitiesOrPaths)
+    paths <- unlist(
```
msevestre, line 44: unless the formatter does it autimatically (!?!) put this in one line. (unlist(lapply())

msevestre:
A few minor comments. Looks good.

R/entity.R
```diff
@@ -10,12 +10,16 @@ Entity <- R6::R6Class(
...
   private = list(
-    .parentContainer = NULL
+    .parentContainer = NULL,
+    .path = NULL
   ),
   active = list(
     #' @field path The path of the entity in the container hiearchy without the simulation name. (read-only)
     path = function(value) {
-      private$wrapExtensionMethod(EntityExtensions, "ConsolidatedPath", "path", value)
+      if (is.null(private$.path)) {
```
PavelBal, line 19: Cache the path as it never changes.

R/quantity.R
```diff
@@ -17,21 +17,27 @@ Quantity <- R6::R6Class(
     },
     #' @field unit The base unit in which the quantity value is defined (Read-Only)
     unit = function(value) {
-      private$wrapExtensionMethod(WITH_DIMENSION_EXTENSION, "BaseUnitName", "unit", value)
+      if (is.null(private$.unit)) {
```
PavelBal, line 20: Cache unit and dimension to avoid rClr-overhead.

R/utilities-pk-analysis.R
```diff
@@ -66,18 +66,19 @@ importPKAnalysesFromCSV <- function(filePath, simulation) {
...
-    colnames(pkResultsDataFrame) <- c("IndividualId", "QuantityPath", "Parameter", "Value", "Unit")
-    pkResultsDataFrame$QuantityPath <- as.factor(pkResultsDataFrame$QuantityPath)
-    pkResultsDataFrame$Parameter <- as.factor(pkResultsDataFrame$Parameter)
-    pkResultsDataFrame$Unit <- as.factor(pkResultsDataFrame$Unit)
-    return(pkResultsDataFrame)
-  },
-  finally = {
-    file.remove(pkParameterResultsFilePath)
-  }
+  dataFrame <- tryCatch(
```
PavelBal, line 69: Only styler changes.

R/utilities-simulation-results.R
PavelBal: OK changed to iteration.

R/utilities-simulation-results.R
PavelBal: Done.

R/utilities-simulation-results.R
PavelBal: Re-run styler on the package.

PavelBal:
@msevestre I updated according to your comments.

Use-case scenario: simulate a steady-state of a model when all quantities are provided. Version 9.2 needs 55 seconds, of which 17 are simulation time. With this PR, first run takes 33 seconds, the consequent runs only 20. The fist run takes longer because of caching of the unit, dimension, and path of each quantity.

msevestre:
@PavelBal This is kind of crazy that it takes so much just to read from the .NET layer. I think this is a very good investigation brother and we should be careful in the future to cache such read only info as early as possible

R/utilities-simulation-results.R
msevestre: @PavelBal I like the usage of lapply better (easier to read) But I am not sure that this it is all around better because of double iteration. 
Would you mind putting it back in?

msevestre (commented):
Looks very good. I think this is more an overhead or using extension methods than pure .NET properties. But good go know
Good for me. I'd love to see the lapply back because it's nicer on the eye but up to you

msevestre:
@PavelBal I'll merge this for now and we can revisit

msevestre:
@PavelBal However the tests are failing. We lose the "Error when setting values" etc.. You need to throw an exception when value is provided

R/dot-net-wrapper.R
```diff
@@ -59,6 +59,17 @@ DotNetWrapper <- R6::R6Class(
       }
     },
 
+    wrapExtensionMethodCached = function(typename, methodName, propertyName, cachedValue, value) {
```
PavelBal, line 62: @msevestre look at this, I added a wrapper for cached fields.

PavelBal:
@msevestre Please check the method for cached fields, otherwise we are good to go I think.

msevestre (commented):
Looks good to me

msevestre:
@PavelBal I am wondering if we should nto cash the isTable stuff in the formula as well. This was also taking a lot of time and is till using extension methods

PavelBal:
I wasn't sure whether these fields are always constant or can change their values. E.g., if changing the value of a parameter to a constant... If you can make a list of those fields that are definitely immutable, caching them would be a good thing.

msevestre:
There are using extension methods anyways which is readonly. So caching it would not change the behavior. Just a bit of overhead

PR #464, opened 2020-12-23
Fixes #444 batch simulation

### LINKED ISSUE #444: Enable Batch Simulation Run
FYI @Yuri05 @PavelBal @StephanSchaller @abdelr

This is so that one can support batch mode in R as well

Something like this could be defined

```R

# simulation: A simulation loaded from pkml
#parameters: Array of parameter by path or parameter instanced that will be varied

simulationBatch<- SimulationBatch$new(simulation, parameters)

# THEN

simulationBatch.setParameterValues("Organism|P1", 5)

# or
simulationBatch.setParameterValues(c("Organism|P1", "Organism|P2"), c(5, 10))

# and 

res <- simulationBatch.Run()

```

How does that sound?

msevestre:


msevestre:
Hi everyone (specifically @PavelBal  and @StephanSchaller )
I have a first version of the SimulationBatch concept running where parameters and initial values can be varied in a simulation
Be mindful that one run will only be on one thread so it's up to you to create multiple instances of the Simulationbatch  to achieve parallelism

R/simulation-batch-run-values.R
```diff
@@ -0,0 +1,46 @@
+#' @title SimulationBatchRunValues
+#' @docType class
+#' @description  Options to be passed to the SimulationBatch run
+#' @export
+#' @format NULL
+SimulationBatchRunValues <- R6::R6Class(
```
msevestre, line 6: Internal class. Not sure if we should export it

R/simulation-batch-options.R
```diff
@@ -0,0 +1,46 @@
+#' @title SimulationBatchOptions
+#' @docType class
+#' @description  Options to be passed to the SimulationBatch
+#' @export
+#' @format NULL
+SimulationBatchOptions <- R6::R6Class(
```
msevestre, line 6: Internal class. Not sure if we should export it

R/utilities-simulation.R
```diff
@@ -122,6 +122,59 @@ runSimulation <- function(simulation, population = NULL, simulationRunOptions =
...
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#'
+#' # Create a simulation batch that will allow batch run for one parameter value
+#' simulationBatch <- createSimulationBatch(sim, "Organism|Liver|Volume")
+#'
+#' # Create a simulation batch that will allow batch run for multiple parameter values and initial values
+#' simulationBatch <- createSimulationBatch(sim, c("Organism|Liver|Volume", "R1-k1"), c("Organism|Liver|A"))
+#' @export
+createSimulationBatch <- function(simulation, parametersOrPaths = NULL, moleculesOrPaths = NULL) {
```
msevestre, line 150: @PavelBal @StephanSchaller @Yuri05 This is the entry point
One creates a simulation batch object passing a bunch of parameters (one or more as path or instances) or molecules
Then the batch will be the one object tot use to run simulations

tests/dev/script-simulation-batch.R
```diff
@@ -0,0 +1,15 @@
+library(ospsuite)
```
msevestre, line 1: @PavelBal @StephanSchaller @Yuri05 
Here is a simple example of how to use the simulation batch

msevestre (commented):
I think this is ready to use

PavelBal:
Looks great! I might have to skip the holidays, this is too exciting :D

Will try to review next week.

StephanSchaller:
> so it's up to you to create multiple instances of the Simulationbatch to achieve parallelism

We are working on it ;-)

StephanSchaller:
@abdelr please also have a look ;-)

msevestre:
Here is a package that can be downloaded
https://ci.appveyor.com/api/buildjobs/vcxxn0r52851ss6f/artifacts/ospsuite_10.0.0.zip

Please test it out and let me know. This is using the same infrastructure used by the ParameterIdentificaiton. I merely exposed the functions and extended the behavior to also support initial values for molecules

Yuri05 (approved):

PR #469, opened 2021-01-14
Performance improvement

PavelBal:


msevestre (commented):
@pchelle  @abdullahhamadeh  Check it out for small quick win apparently for performances. 
(especially %%  vs any) . Hard to quantify but easy to implement

msevestre:
@PavelBal Thanks

PR #484, opened 2021-03-08
Fixes #295 population calculation aging

### LINKED ISSUE #295: Population calculations and aging
Following the discussion in https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1406: should aging data be supported by the R-Toolbox?

At the moment population simulation without aging can be executed:
runSimulation(sim, population)
Should it be possible to incorporate aging data, e.g.
runSimulation(sim, population, agingData)?

Aging data would contain {Time, Value}-(table)-profiles of some simulation parameters.

msevestre:


R/utilities-population.R
```diff
@@ -90,3 +90,23 @@ exportPopulationToCSV <- function(population, filePath) {
...
+#' Loads aging data (typically generated from PK-Sim) i
+#'
+#' @param filePath Full path containing an aging data table.
+#'
+#' @examples
+#' csvPath <- system.file("extdata", "aging_data.csv", package = "ospsuite")
+#'
+#' agingData <- loadAgingDataFromCSV(tempfile())
+#' @export
+loadAgingDataFromCSV <- function(filePath){
```
msevestre, line 102: @pchelle @abdullahhamadeh @PavelBal 
I am losing my mind here. 
I have 4 columns in the table. IndividualId is an integer, ParameterPath is string, Time and Value are doubles
Can I make sure R does not try to be too clever and does not convert my string into factor AND also interprets my number as numbers? Using stringAsFactor = FALSE converst all number as string.

R/utilities-population.R
```diff
@@ -90,3 +90,23 @@ exportPopulationToCSV <- function(population, filePath) {
...
+  validateIsString(filePath)
+  df <- read.csv(filePath, encoding = "UTF-8", check.names = FALSE)
+  colnames(df) <- c("IndividualId", "ParameterPath", "Time", "Value")
+  agingData <- AgingData$new();
+  agingData$individualIds <- as.integer(df$IndividualId)
+  agingData$parameterPaths <- as.character(df$ParameterPath)
+  agingData$times <- as.numeric(df$Time)
```
msevestre, line 109: This does not work but I do not know how to get it right

tests/data/baby_aging.csv
```diff
@@ -0,0 +1,2510 @@
+﻿#Project: asd,,,,
```
msevestre, line 1: @pchelle @Yuri05 @PavelBal @abdullahhamadeh 
ok I am havign issues reading this file bacaue it looks corrupted! WTF
That being said, how do you deal with # in csv .using the `comment.char = '#'` does nto seem to do the trick

R/utilities-pk-analysis.R
```diff
@@ -68,7 +67,7 @@ pkAnalysesAsDataFrame <- function(pkAnalyses) {
   dataFrame <- tryCatch(
     {
       exportPKAnalysesToCSV(pkAnalyses, pkParameterResultsFilePath)
-      pkResultsDataFrame <- read.csv(pkParameterResultsFilePath, encoding = "UTF-8", check.names = FALSE)
+      pkResultsDataFrame <- readr::read_csv(pkParameterResultsFilePath, locale = readr::locale(encoding = "UTF-8"), comment = "#")
```
msevestre, line 70: @PavelBal @Yuri05 Moved to read_csv here.
readr is now a required package

PR #485, opened 2021-03-09
471 document update

msevestre:


R/ospsuite-env.R
```diff
@@ -1,3 +1,11 @@
+
+.getSuiteVersion <- function() {
+  version <- getNamespaceVersion("ospsuite")
```
msevestre, line 3: @Yuri05 @PavelBal This is in fact the better way apparently to load the version. Otheriwse, if you have two version installed, the proposed packageVersion reads the first one all the time. This reads the one loaded

R/ospsuite-env.R
```diff
@@ -7,7 +15,7 @@ ospsuiteEnv$packageName <- "ospsuite"
 
 ospsuiteEnv$suiteName <- "Open Systems Pharmacology"
 
-ospsuiteEnv$packageVersion <- "10.0"
+ospsuiteEnv$suiteVersion <- .getSuiteVersion();
```
msevestre, line 18: I have renamed this to be more idiomnatic

msevestre:
ALso I ran the styler on the package as some file were out  of style :)

R/ospsuite-env.R
PavelBal: Looks good!

Issue #486, opened 2021-03-17
Expose DataRepository (observedData) to R


msevestre:
It should be possible to load a dataRepository from pkml (as exported by mobi or pksim) 
* Time
* Values
* Error 
* ErrorType (arithmetic vs geemoetric)
* TimeDimension
* ValuesDimension

We should aim at having a similar interface as our TableFormula for consistency sake

Also MetaData should be made available as a list of string, values

@PavelBal @Yuri05 @abdelr @StephanSchaller 
Anything else?

PavelBal:
What about units? Or are the values always in default units?

Yuri05:
>What about units? Or are the values always in default units?

The values are stored always in the internal units, but maybe it makes sense to expose the saved UI units?

>TimeDimension

Isn't the TimeDimension always "Time"? Or do you have "general" XY-data repositories in mind?

> Also MetaData should be made available as a list of string, values

Should we distinguish between **string** and **numeric** values in the metadata? Especially when exposing LLOQ/MW/... numerci would make more sense.

msevestre:
>Isn't the TimeDimension always "Time"? Or do you have "general" XY-data repositories in mind?

In general yes. But since DataRepository does support sthg else, why not allow it to be different. Similar to a TableFormula with Dimension and YDimension, we can have something similar. 

Or we restrict it to observed Data with TIME only as XDimension. 

>Should we distinguish between string and numeric values in the metadata? Especially when exposing LLOQ/MW/... numerci would make more sense.

Yes that's why I wrote list of {string, values}. string would be the key, value would be whatever is saved in the system. As far as I know, values can;t be of different type in a list right?

msevestre:
but what should be the interface of the object
```R
obs$value(s) => numeric array
obs$time(s)  => numeric array
obs$error(s)  => numeric array
obs$valueAt(time) => numeric
obs$metaData$studyId => string/numeric
obs$YDisplayUnit ?
obs$XDisplayUnit ?

```

PavelBal:
```
obs$YDisplayUnit  -> string
obs$XDisplayUnit  -> string
```

Yuri05:
>But since DataRepository does support sthg else, why not allow it to be different. Similar to a TableFormula with Dimension and YDimension, we can have something similar.

Agree. Let's then just turn the "TimeDimension" into "XDimension" or "ArgumentDimension".

PR #487, opened 2021-03-18
WIP #486 expose data repository

msevestre:
This is just a WIP. I would like to understand what we need exactly
So far

```R
library(ospsuite)

obsData <- loadObservedData("tests/data/obs_data.pkml")

> obsData$baseGrid
  Time: [min]

> obsData$columns
[[1]]
  Time: [min]

[[2]]
  Measurement: [kg/l]

[[3]]
  Error: [kg/l]

> obsData$columns[[2]]$values
[1] 1.01e-09 8.30e-10 7.30e-10 2.80e-10 1.20e-10 3.05e-11 5.93e-12

> col<-obsData$columns[[2]]
> col$dimension
[1] "Concentration (mass)"
> col$displayUnit
[1] "mg/l"
> col$unit
[1] "kg/l"
> col$name
[1] "Measurement"
> col$values
[1] 1.01e-09 8.30e-10 7.30e-10 2.80e-10 1.20e-10 3.05e-11 5.93e-12
```


R/data-repository.R
```diff
@@ -0,0 +1,28 @@
...
+  "DataRepository",
+  inherit = DotNetWrapper,
+  cloneable = FALSE,
+  active = list(
+    #' @field baseGrid Returns the base column for the population (typically time column)
+    baseGrid = function(value) {
+      DataColumn$new(private$wrapReadOnlyProperty("BaseGrid", value))
+    },
+    #' @field columns Returns all columns (including baseGrid defined in the data)
+    columns = function(value){
```
msevestre, line 15: @PavelBal @Yuri05 I will cache this so that we always return the same instances of DataColumn wrapper

R/data-repository.R
```diff
@@ -0,0 +1,28 @@
...
+#' @docType class
+#' @description  An object typically holding observed data
+#' @format NULL
+DataRepository <- R6::R6Class(
+  "DataRepository",
+  inherit = DotNetWrapper,
+  cloneable = FALSE,
+  active = list(
+    #' @field baseGrid Returns the base column for the population (typically time column)
+    baseGrid = function(value) {
```
msevestre, line 11: in core, we have basegrid, columns, and allButBaseGrid.
I will implement the same methods (baseGrid is typically time, columns are all columns in the dataRepo, allButBaseGrids are all except basegrid)

Should we have a property called Time that returned baseGrid? or property called Measurement that returns col2 and a property called error that returns col3? not sure

msevestre (commented):
WIP

R/data-repository.R
Yuri05: >Should we have a property called Time that returned baseGrid? or property called Measurement that returns col2 and a property called error that returns col3?

I would say no and no. Basegrid might be different from "Time", we could have several measurements and several errors, so I wouldn't provide shortcuts for the specific use case "_ObservedDataRepo = {Time, Measurement, Error}_"

R/data-column.R
```diff
@@ -0,0 +1,52 @@
...
+    }
+  ),
+  public = list(
+    #' @description
+    #' Print the object to the console
+    #' @param ... Rest arguments.
+    print = function(...) {
+      if (self$unit == "") {
+        private$printLine(self$name)
+      } else {
```
PavelBal, line 42: @msevestre 
I would not make this differentiation. `[]` is assumed to be the unit of `Dimensionless` and is used like this e.g. in excel files when reading the data. From my perspective, it would be ok to print something like

> Data  name, []

R/data-repository.R
```diff
@@ -0,0 +1,56 @@
+#' @title DataRepository
+#' @docType class
+#' @description  An object typically holding observed data
```
PavelBal, line 3: Description is identical with that of `DataColumn`- please change it to something more distinguishable.

R/data-repository.R
```diff
@@ -0,0 +1,56 @@
+#' @title DataRepository
+#' @docType class
+#' @description  An object typically holding observed data
+#' @format NULL
+DataRepository <- R6::R6Class(
+  "DataRepository",
+  inherit = DotNetWrapper,
+  cloneable = FALSE,
+  active = list(
+    #' @field baseGrid Returns the base column for the population (typically time column)
```
PavelBal, line 10: Why `population`?

R/utilities-observed-data.R
```diff
@@ -0,0 +1,16 @@
+#' Loads a population from a csv file and returns the population.
```
PavelBal, line 1: Not a `population`.

PavelBal (changes requested):
@msevestre I propose some changes to the naming.

R/data-repository.R
msevestre: Yes sorry. All descriptions were copied paste so that I could get going. Still WIP

R/data-repository.R
msevestre: In theory results from a simulation etc.. are DataRepositories
So yes we can have one time and multiple measurements. Each measurements can in theory have multiple errors. I think we are using this for the PI results. 
Practically probably not

R/data-column.R
msevestre: I think we are removing it when printing a quantity value etc so I just wanted to ber consistent

msevestre: see 
```
 #' @description
    #' Print the the value and unit of the quantity
    #' @param  caption Text to prepend to the value
    printQuantityValue = function(caption) {
      if (self$unit == "") {
        private$printLine(caption, formatNumerics(self$value))
      } else {
        private$printLine(caption, paste0(formatNumerics(self$value), " [", self$unit, "]"))
      }
    },
```

R/utilities-observed-data.R
```diff
@@ -0,0 +1,16 @@
+#' Loads a population from a csv file and returns the population.
+#'
+#' @param filePath Full path of pkml file containing the observed data to load
+#'
+#' @examples
+#' filePath <- system.file("extdata", "obsData.pkml", package = "ospsuite")
+#'
+#' obsData <- loadObservedData(filePath)
+#' @export
+loadObservedData <- function(filePath) {
```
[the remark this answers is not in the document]
msevestre: ok. I did not want to use the term data-repository because it's a bit abstract but why not

R/init-package.R
```diff
@@ -18,6 +18,4 @@ initPackage <- function() {
   apiConfig$pkParametersFilePath <- filePathFor("OSPSuite.PKParameters.xml")
 
   rClr::clrCallStatic("OSPSuite.R.Api", "InitializeOnce", apiConfig$ref)
-  dimensionTask <- getNetTask("DimensionTask")
```
msevestre, line 21: @Yuri05 THis was creating some issues. And then I wondered. Do we actually need to load if from .NET 
isn't it enough to have it a constant as it is now?

R/utilities.R
```diff
@@ -48,6 +48,10 @@ ifNotNull <- function(condition, outputIfNotNull, outputIfNull = NULL) {
 #' See https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/476 for details
 #' @import stringr
 encodeUnit <- function(unit) {
+  
+  #TODO maybe we don't even need to load from .NET
+  # dimensionTask <- getNetTask("DimensionTask")
```
msevestre, line 53: @Yuri05 If we want to load from .NET we would need to do it here the first time is is being used. But why do we need this in the first place since ` ospsuiteEnv$muSymbol` is already set 
see ospsuite-env.R

# Specifies the symbol used for µ. This will be set by the .NET layer
ospsuiteEnv$muSymbol <- "µ"

R/data-column.R
PavelBal: All right then.

R/data-repository.R
```diff
@@ -0,0 +1,73 @@
...
+      private$throwPropertyIsReadonly("columns")
+    },
+    #' @field metaData Returns a named list of meta data defined for the data repository.
+    #' where the name is the name of the metaData and the value is the meta data value.
+    metaData = function(value){
+      if (missing(value)) {
+        if (is.null(private$.metaData)) {
+          netExtendedProperties <- private$wrapReadOnlyProperty("ExtendedProperties", value)
+          netMetaData <- rClr::clrGet(netExtendedProperties, "All")
+          names <- unlist(lapply(netMetaData,  function(data) rClr::clrGet(data, "Name")))
```
PavelBal, line 48: > unlist(x, use.names = FALSE) is much faster than unlist(x).

No idea what the actual benefit would be here, but why not add it..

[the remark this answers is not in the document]
PavelBal: Can't think of anything.

R/init-package.R
PavelBal, line 21: Why do you remove "DimensionTask"? I was counting on it for the `utilities-dimensions`. 

Btw. I only was creating it the first time it was requested:

In ospsuite-env (esqlabsEnv in my code):

```
# NetTask "DimensionTask" cached for performance benefits. Created the first time it is requested.
esqlabsEnv$DimensionTask <- NULL

```

Then a function somewhere:

```
#' Return an instance of the .NET Task "DimensionTask".
#'
#' @return An instance of the Task
getDimensionTask <- function() {
  dimTask <- esqlabsEnv$DimensionTask
  if (is.null(dimTask)) {
    dimTask <- getNetTask("DimensionTask")
    esqlabsEnv$DimensionTask <- dimTask
  }
  return(dimTask)
}
```

R/utilities.R
```diff
@@ -48,6 +48,10 @@ ifNotNull <- function(condition, outputIfNotNull, outputIfNull = NULL) {
 #' See https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/476 for details
 #' @import stringr
 encodeUnit <- function(unit) {
+  
+  #TODO maybe we don't even need to load from .NET
```
PavelBal, line 52: Oh you have it here... See my suggestion above.

PavelBal (changes requested):

R/init-package.R
msevestre: I don't remove it. Instatntiating object in the init creates weird error messages during build.

Can you quantify ` cached for performance benefits.`.

R/utilities.R
msevestre: I don't understand what you mean

R/data-repository.R
msevestre: updated

R/init-package.R
msevestre: I think you are looking at sthg else. :)

msevestre: but I am going to add it to core because it's probably a good optimization since this is used quite often

PR #491, opened 2021-03-24
Added methods for working with enums.

PavelBal:


R/enum.R
```diff
@@ -25,6 +25,108 @@ enum <- function(enumValues) {
...
+#' it will be overwritten with the new value. Otherwise, an error is thrown. Default is FALSE.
+#'
+#' @return Enum with added key-value pair.
+#' @export
+#'
+#' @examples
+#' myEnum <- enum(c(a = "b"))
+#' myEnum <- enumPut("c", "d", myEnum)
+#' myEnum <- enumPut(c("c", "d", "g"), c(12, 2, "a"), myEnum, overwrite = TRUE)
+enumPut <- function(keys, values, enum, overwrite = FALSE) {
```
PavelBal, line 95: Allows to extend existing enums

msevestre:
@PavelBal CAn you add a test file for this? The behavior is starting to justify it :)

R/enum.R
```diff
@@ -25,6 +25,108 @@ enum <- function(enumValues) {
...
+#' WARNING: the original object is not modified!
+#'
+#' @return Enum without the removed entries
+#' @export
+enumRemove <- function(keys, enum) {
+  for (key in keys) {
+    enum[[key]] <- NULL
+  }
+
+  return(enum)
```
msevestre, line 121: SAy enum is 
myEnum <- enum(c(k1 = "a", k2="b"))

and I say 
new <- enumRemove(c(k1), myEnum)
Isn't new with an entry for k1 = NULL?

R/enum.R
```diff
@@ -25,6 +25,108 @@ enum <- function(enumValues) {
...
+
+#' Remove an entry from the enum.
+#'
+#' @param keys Key(s) of entries to be removed from the enum
+#' @param enum Enum from which the entries to be removed
+#' WARNING: the original object is not modified!
+#'
+#' @return Enum without the removed entries
+#' @export
+enumRemove <- function(keys, enum) {
```
msevestre, line 116: woudl it work with keys being ONE value and not an array?

msevestre (changes requested):

R/enum.R
PavelBal: > With lists, you can use x[[i]] <- NULL to remove a component. To add a literal NULL, use x[i] <- list(NULL):

https://adv-r.hadley.nz/subsetting.html#subassignment

R/enum.R
PavelBal: Yepp.

PavelBal:
@msevestre I  added some tests.

msevestre:
@PavelBal Build is failing

msevestre:
oh yeah it's failing because I need to merge the other branch first. then if you can merge/rebase develop into your branch, we should be good to go

R/enum.R
msevestre: cool. Thanks :)

R/enum.R
msevestre: So why do we always do sthg like this in ospsuite-r

quantities<- c(quantities) to ensure that we have a list

PavelBal: I think... Because a quantity (or any R6-object) by themselves are kind of a list, and trying to iterate over an R6-object produces some mess... Having an object as enum-key would not work.

PavelBal:
@msevestre rebased

R/enum.R
msevestre: oh yeah that's right. Thanks brother

msevestre:
looks good. I'll merge when the build is finished

msevestre:
@PavelBal You need to create thew new documentation file (CTRL+SHIFT+D )to update the documentation. Otheriwse it won't be vibisle

PavelBal:
Oh, I thought it's re-created on build... Added!

msevestre:
still one example that is being run...pain in the butt. 
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r/builds/38385449

You probably need to regenerate the documentation so that it is being updated.

PavelBal:
@msevestre Actually, I just realized... I expose all those methods, but we do not expose `enum` itself (also this is why examples are failing). I think we should either export everything or nothing. I would prefer exporting `enum` so it can be used by user. What do you think?

msevestre:
This was an issue because enum is used in TLF and other Lib and we were getting conflicts. That why we didn't export. Yes you can still use it from the outside?

PavelBal:
>Yes you can still use it from the outside?

You cannot create a new one:

```
> library(ospsuite)
> myEnum <- enum(c("a", "b"))
Error in enum(c("a", "b")) : could not find function "enum"
```

msevestre:
What we can try to do is
expor the one in ospsuite.core
Do not create one at all in ReportingEngine and use this one
Rename the one in tlf. 
For some reason, I think this was not working

Anyway, in the meantime, can you put the dontrunn flag or sthg like this in the example so that the build passes?

PR #494, opened 2021-03-25
Fixes #493 units for dimensions

### LINKED ISSUE #493: Expose units for dimension
Ensure that it is possible to retrieve all available units for a given dimension

msevestre:


R/utilities-units.R
```diff
@@ -121,7 +121,34 @@ allAvailableDimensions <- function() {
...
+  dimensionTask <- getDimensionTask()
+  rClr::clrCall(dimensionTask, "AllAvailableUnitNamesForDimension", dimension)
+}
+
+
+#' Return an instance of the .NET Task `DimensionTask`
+#' This is purely for optimization purposes
+#'
+#' @return An instance of the Task
+getDimensionTask <- function() {
```
msevestre, line 147: @PavelBal I added your method here. I'd love to know how much of a difference such as small optimization is making

NAMESPACE
```diff
@@ -65,6 +65,7 @@ export(getParameterDisplayPaths)
 export(getQuantity)
 export(getSimulationTree)
 export(getStandardMoleculeParameters)
+export(getUnitsForDimension)
```
msevestre, line 68: @abdullahhamadeh This will be soon available in ospsuite-r Returns all units for a given dimension name

R/utilities-units.R
PavelBal: To be honest I did not investigate it. Might think of a use case and profile it.

PavelBal (approved):
Looks good.

R/utilities-units.R
msevestre: Yeah. I am pretty sure this is completely insignificant to be honest. But well why not

Issue #478, opened 2021-03-04
Enums to look up units and dimensions


abdullahhamadeh:
It would be useful at the user level to have enums to easily input units and dimensions when needed, and also for these enums to be loaded automatically with the package.

Examples:

To correctly spell dimensions with worrying about, eg, capitalization:
![image](https://user-images.githubusercontent.com/18409098/109980829-52408780-7cce-11eb-82d4-20138e10665d.png)

To avoid typing superscripts and to know what units are available for any given dimension:
![image](https://user-images.githubusercontent.com/18409098/109980987-7ac88180-7cce-11eb-9f34-512d0a493dd3.png)

To know what base units are used may also be useful:
![image](https://user-images.githubusercontent.com/18409098/109981435-e4e12680-7cce-11eb-8956-724eb72b6026.png)

msevestre:
@abdullahhamadeh `getUnitsForDimension` is in ospsuite-r

msevestre:
@abdullahhamadeh Can you give me an ETA for this?

abdullahhamadeh:
PR #499

PR #495, opened 2021-03-25
Utilities dimensions

PavelBal:
@msevestre No tests so far.


R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+#' @param mw Double value of the molecular weight of the molecule. Used
+#' for conversion between molar and mass concentrations or amounts. Can be \code{NULL}
+#' for other dimensions (default)
+#'
+#' @return A numerical factor the values of the first dimensions must be multiplied by to
+#' be converted to second dimension. The conversion assumes values being in base units
+#' of both dimensions.
+#' If the values cannot be transformed between the dimensions, an error is thrown.
+#' @export
+dimensionsConversionFactor <- function(dimension1, dimension2, mw = NULL) {
```
PavelBal, line 76: This is very useful if you want e.g. to plot your data in molar or mass concentrations. Is there a better way to check if the dimensions can be converted than to explicitly consider every possible use case?

PavelBal:
Maybe some things can be done better (using some .NET methods)?

PavelBal:
Something is broken here..

msevestre:
@PavelBal I have merge develop into the code here. Still it won't work but we have a clean slate

DESCRIPTION
```diff
@@ -89,6 +89,7 @@ Collate:
     'user-defined-pk-parameter.R'
     'utilities-container.R'
     'utilities-data-repository.R'
+    'utilities-dimensions.R'
```
msevestre, line 92: we have a utility-units. Let's move the code here instead.

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+  convFac <- convFac * rClr::clrCall(dimension, "BaseUnitValueToUnitValue", toUnitObj, 1)
+  return(convFac)
+}
+
+#' #' Supported dimensions.
+#' #'
+#' #' @include enum.R
+#' #'
+#' #' @export
+#' Dimensions <- enum(allAvailableDimensions())
```
msevestre, line 162: This is a problem here. It is being loaded directly before initialization. This works when you load ospsuite.core first but does not work in ospsuite.core

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' @return Numerical value. A value in \code{fromUnit} multiplied  by the factor
+#' corresponds to the value in \code{toUnit}
+#' @export
+getUnitConversionFactor <- function(fromUnit, toUnit, dimension) {
+  validateDimension(dimension)
+  validateUnit(fromUnit, dimension)
+  validateUnit(toUnit, dimension)
+
+  dimensionTask <- getDimensionTask()
```
msevestre, line 148: I am not sure why you need sthg like this ? toUnit converts already everything from you. .The goal is to abstract exactly all of this away

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+#' @param mw Double value of the molecular weight of the molecule. Used
+#' for conversion between molar and mass concentrations or amounts. Can be \code{NULL}
+#' for other dimensions (default)
+#'
+#' @return A numerical factor the values of the first dimensions must be multiplied by to
+#' be converted to second dimension. The conversion assumes values being in base units
+#' of both dimensions.
+#' If the values cannot be transformed between the dimensions, an error is thrown.
+#' @export
+dimensionsConversionFactor <- function(dimension1, dimension2, mw = NULL) {
```
msevestre, line 76: oh my. what is this monster?

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+
+  # Concentration (molar) <-> Concentration (mass)
+  if (dimension1 == Dimensions$`Concentration (molar)`) {
+    if (dimension2 == Dimensions$`Concentration (mass)`) {
+      if (is.null(mw)) {
+        stop(messages$errorCannotConvertDimensionsNoMW(dimension1, dimension2))
+      }
+      return(1 * 1e-6 * mw * 1e-3)
```
msevestre, line 101: This is not something I want to see here. I need to understand the usage of such concept. We have already everything encapsulated in Core

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+  }
+}
+
+#' Get base unit of a dimension
+#'
+#' @param dimension Dimension for which the base unit is returned.
+#'
+#' @return String name of the base unit.
+#' @export
+getBaseUnit <- function(dimension) {
```
msevestre, line 54: This should move to core

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+  validateDimension(dimension)
+  dimensionTask <- getDimensionTask()
+  # Get the dimension object
+  dimension <- rClr::clrCall(dimensionTask, "DimensionByName", enc2utf8(dimension))
+  unit <- rClr::clrCall(dimension, "get_BaseUnit")
```
msevestre, line 59: there is clrGet...

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+existsDimension <- function(dimension) {
+  enumHasKey(dimension, enum = Dimensions)
+}
+
+#' Validate dimension
+#'
+#' @param dimension String name of the dimension.
+#' @details Check if the provided dimension is supported. If not, throw an error
+#' @export
+validateDimension <- function(dimension) {
```
msevestre, line 15: should not we have one method (validateDimension or existDimension with a flag instead of two?) we have the same concept already somewhere else

R/utilities-dimensions.R
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#' Unit existence
+#'
+#' @param unit String name of the unit
+#' @param dimension String name of the dimension.
+#' @details Check if the unit is valid for the dimension.
+#' @export
+hasUnit <- function(unit, dimension) {
+  validateIsString(c(unit, dimension))
```
msevestre, line 28: This should move ot core too

msevestre (changes requested):

R/utilities-dimensions.R
PavelBal: The idea is to use the conversion capabilities of OSPS without having the simulation objects (quantities). E.g. read the data from excel and plot it in another unit. `toUnit` requires a Quantity.

R/utilities-dimensions.R
PavelBal: Would be happy to see another way. E.g. - I read data in mg/dl and want to convert it to mol/l. How? Without having a Quantity object.

R/utilities-dimensions.R
```diff
@@ -0,0 +1,160 @@
+#' Dimension existence
+#'
+#' @param dimension String name of the dimension.
+#' @details Returns \code{TRUE} if the provided dimension is supported otherwise \code{FALSE}
+#' @export
+hasDimension <- function(dimension) {
```
msevestre, line 6: @PavelBal I have renamed this to hasDimension and it is using the core method

R/utilities-dimensions.R
```diff
@@ -0,0 +1,160 @@
...
+#'
+#' @param unit String name of the unit
+#' @param dimension String name of the dimension.
+#' @details Check if the unit is valid for the dimension.
+#' @export
+hasUnit <- function(unit, dimension) {
+  validateIsString(unit)
+  validateDimension(dimension)
+  dimensionTask <- getDimensionTask()
+  rClr::clrCall(dimensionTask, "HasUnit", enc2utf8(dimension),  enc2utf8(unit))
```
msevestre, line 34: usign hasUnit from core

R/utilities-dimensions.R
```diff
@@ -0,0 +1,160 @@
...
+#' Get base unit of a dimension
+#'
+#' @param dimension Dimension for which the base unit is returned.
+#'
+#' @return String name of the base unit.
+#' @export
+getBaseUnit <- function(dimension) {
+  validateDimension(dimension)
+  dimensionTask <- getDimensionTask()
+  rClr::clrCall(dimensionTask, "BaseUnitFor", enc2utf8(dimension))
```
msevestre, line 58: using baseUnitFor from core

R/utilities-dimensions.R
msevestre: There are plenty of method exposted directly in core doing this. Creating a conversation factor from one dimension to another dimension does not make sense BTW. It needs to be from one unit to another unit

In core see

```
  double[] ConvertToUnit(string dimensionName, string targetUnit, double[] valuesInBaseUnit, double molWeight);
      double[] ConvertToUnit(string dimensionName, string targetUnit, double[] valuesInBaseUnit);

      double[] ConvertToUnit(string dimensionName, string targetUnit, double valueInBaseUnit, double molWeight);
      double[] ConvertToUnit(string dimensionName, string targetUnit, double valueInBaseUnit);

```

and 
```
 double[] ConvertToBaseUnit(string dimensionName, string displayUnit, double[] valuesInDisplayUnit, double molWeight);
      double[] ConvertToBaseUnit(string dimensionName, string displayUnit, double[] valuesInDisplayUnit);

      double[] ConvertToBaseUnit(string dimensionName, string displayUnit, double valueInDisplayUnit, double molWeight);
      double[] ConvertToBaseUnit(string dimensionName, string displayUnit, double valueInDisplayUnit);

```

All exposed via the DImensionTag. IT deals EXPLICITELY with those use cases

R/utilities-dimensions.R
msevestre: nope

```
toUnit <- function(quantityOrDimension, values, targetUnit, molWeight = NULL) {
```

Specifically the parameter is called quantityOrDimension.

None of the proposed changes regarding reimplementing conversion by hand belong in Core. Helper function to ensure that dimension and unit etc.. exists, yes.  The rest probably not :)

msevestre (commented):
This sitll needs to be moved to utilities-units

R/utilities-dimensions.R
PavelBal: Why do you prefer the core method? We still have the enum Dimensions (or do you think we do not need it? I would disagree) which is created by calling `allAvailableDimensions`.

PavelBal: Why is it called `HasDimension` btw? What does or does not have the dimension? :D @msevestre

R/init-package.R
```diff
@@ -18,4 +18,6 @@ initPackage <- function() {
   apiConfig$pkParametersFilePath <- filePathFor("OSPSuite.PKParameters.xml")
 
   rClr::clrCallStatic("OSPSuite.R.Api", "InitializeOnce", apiConfig$ref)
+
```
PavelBal, line 21: This allows to populate the enum after loading the package..

R/utilities-units.R
```diff
@@ -136,9 +196,47 @@ getDimensionForUnit <- function(unit) {
...
+  rClr::clrCall(dimensionTask, "AllAvailableUnitNamesFor", dimension)
+}
+
+#' Get conversion factor from one unit to another
+#'
+#' @param fromUnit String name of the unit to convert from
+#' @param toUnit String name of the unit to convert to
+#' @param dimension Dimension of the units (conversion between mass and molar possible when
+#' \code{MW} is provided)
+#' @param MW Optional molecule weight in g/mol to use when converting, for example, from molar to mass amount or concentration
```
PavelBal, line 208: I think MW should be passed as g/mol, this is the only unit a user expects.
@msevestre

R/utilities-units.R
```diff
@@ -136,9 +196,47 @@ getDimensionForUnit <- function(unit) {
...
+  )
+  return(convFac)
 }
 
+#' Supported dimensions.
+#'
+#' @include enum.R
+#'
+#' @export
+Dimensions <- list()
```
PavelBal, line 238: Initialize empty and then populate after package is loaded.

R/utilities-dimensions.R
msevestre: You are looking at this from the system point of view " Does OSPSUITE HAVE The dimension xxx"
so  `if(ospsuite::hasDimension("Mass"))` reads better than `if(ospsuite::existsDImension("Mass"))` in my opinion

But this is also because we had hasUnit. and not existsUnit. Naming is hard and consistency is even harder

R/utilities-units.R
msevestre: >MW Optional molecule weight in g/mol

yes?

R/utilities-units.R
msevestre: I would like not to rely on this too much for now because @abdullahhamadeh is going to implement a parsign method see #478

R/utilities-units.R
PavelBal: Yes? :D

msevestre: Ok I see your point. The problem is that we always pass parameters in core unit
height should be dm etc.. 
Introducing a change like this would not be logical, specifically because the weight generally comes from simulation$molWeightFor which returns in core unit.
I am still not 100% sure I understand why we need such a method. If this is just for the MW, then we could add another flag to toUnit (mw.in.base.unit = true) by default or sthg like this

PavelBal: No its not for the MW, its for conversion of values that are not in base units - and most probably these would be the values that do not come from a simulation (that can be converted with toUnit). It is for data that are defined in other units - loaded from excel, generated by hand, or whatsoever.. And in this case MW most definitely will be in g/mol (even in the data exporter we always assume MW is in g/mol).

I see your point that this will be inconsistent with the rest of methods. But I also do not see the use case for this function with MW being in the base unit.

msevestre: LEt's discuss WEdnesday. I do not understand why a user would not convert an array using toUnit instead of dealing with some weird factors

R/utilities-units.R
```diff
@@ -15,7 +76,7 @@
 #'
 #' valuesInBaseUnit <- toBaseUnit(par, c(1000, 2000, 3000), "ml")
 #' @export
-toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL) {
+toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL, molWeightUnit = NULL) {
```
PavelBal, line 79: To be consistent, both "toBaseUnit" and "toUnit" allow now molWeight in custom unit.

R/utilities-units.R
```diff
@@ -55,12 +121,19 @@ toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL) {
 #' valueInMl <- toUnit(par, 1, "ml")
 #'
 #' valuesInMl <- toUnit(par, c(1, 5, 5), "ml")
+#'
+#' # Converts a numerical value in from mmol/l to mg/dl
+#' valuesInMgDl <- toUnit(Dimensions$`Concentration (molar)`, 5, targetUnit = "mmol/l",
+#' sourceUnit = "mg/dl", molWeight = 180, molWeightUnit = "g/mol")
 #' @export
-toUnit <- function(quantityOrDimension, values, targetUnit, molWeight = NULL) {
+toUnit <- function(quantityOrDimension, values, targetUnit, sourceUnit = NULL, molWeight = NULL, molWeightUnit = NULL) {
```
PavelBal, line 129: This would be the new signature that allows to convert from non-base units.

R/utilities-units.R
```diff
@@ -74,8 +147,20 @@ toUnit <- function(quantityOrDimension, values, targetUnit, molWeight = NULL) {
   }
 
   if (is.null(molWeight)) {
+    #Convert values to base unit first if the source unit is provided
+    if (!is.null(sourceUnit)){
```
PavelBal, line 151: This is repeated in both if-cases, but I cannot think of a better solution now.

R/utilities-units.R
```diff
@@ -139,6 +223,16 @@ getUnitsForDimension <- function(dimension) {
...
+#' Supported dimensions.
+#'
+#' @export
+Dimensions <- list()
+
+#' Supported units of dimensions
+#' enum of all units for each dimension
+#'
+#' @export
+DimensionsUnits <- list()
```
PavelBal, line 235: How about this naming for the enum with the units?

PavelBal:
@msevestre I don't get it, why is the build failing? Obviously it somehow uses  the old dlls, but I cannot figure out why. It works on my local machine.

msevestre:
@PavelBal Ok I am going to check it out

R/utilities-units.R
```diff
@@ -1,9 +1,70 @@
...
+#'
+#' @param unit String name of the unit
+#' @param dimension String name of the dimension.
+#' @details Check if the unit is valid for the dimension.
+#' @export
+hasUnit <- function(unit, dimension) {
+  validateIsString(unit)
+  validateDimension(dimension)
+  dimensionTask <- getDimensionTask()
+  rClr::clrCall(dimensionTask, "HasUnit", enc2utf8(dimension), encodeUnit(unit))
```
msevestre, line 34: this needs to be encodeUnit for unuit

R/utilities-units.R
```diff
@@ -154,21 +239,40 @@ getDimensionTask <- function() {
...
+
+#' Supported units defined as a named list of lists
+#'
+#' ospUnits$Mass$kg => "kg"
+#' @export
+ospUnits <- list()
+
+initializeDimensionAndUnitLists <- function() {
+  #This initializes the two lists in the parent environment which is the package environments
+  ospDimensions <<- getDimensionsEnum()
```
msevestre, line 276: @PavelBal  Here is how we do not polute the global namespace (I think...)

PR #499, opened 2021-03-29
create getUnitsEnum and getDimensionsEnum functions #478

abdullahhamadeh:
Two functions to load the dimensions and units into enums

The function `getUnitsForDimension` throws an error when the input is in `c("CV mmHg*s²/ml","Time²","Flow²")`, so these dimensions are not included.  Encoding issue?


msevestre:
What's the error?

abdullahhamadeh:
image

PavelBal:
@abdullahhamadeh 

Here:
change
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/4c3654b623091f2060026e7010b6df3f449cdedf/R/utilities-units.R#L139

to  
`rClr::clrCall(dimensionTask, "AllAvailableUnitNamesForDimension", enc2utf8(dimension))`

then you do not have to exclude any dimensions.

abdullahhamadeh:
Great - thanks @PavelBal

R/utilities-units.R
```diff
@@ -136,7 +136,7 @@ getDimensionForUnit <- function(unit) {
 getUnitsForDimension <- function(dimension) {
   validateIsString(dimension)
   dimensionTask <- getDimensionTask()
-  rClr::clrCall(dimensionTask, "AllAvailableUnitNamesForDimension", dimension)
+  rClr::clrCall(dimensionTask, "AllAvailableUnitNamesForDimension", enc2utf8(dimension))
 }
```
msevestre, line 140: Yep. Good stuff

R/utilities-units.R
```diff
@@ -152,3 +152,23 @@ getDimensionTask <- function() {
...
+  units <- lapply(dimensions,function(dimension){enum(getUnitsForDimension(dimension = dimension))})
+  names(units) <- dimensions
+  return(units)
+}
+
+#' #'Function to return an enum of all available dimensions
+#' @return enum of all dimensions
+#' @export
+getDimensionsEnum <- function(){
+   dimensions <- enum(allAvailableDimensions())
```
msevestre, line 172: Can you add a test that loads these to ensure that it works as expected?
Also why do we need the getEnumDimensiob?

abdullahhamadeh: > Also why do we need the getEnumDimensiob?

Only because it complements getUnitsEnum and does a similar job, but we could stick with allAvailableDimensions().  But then the user would need to create the enum.  Prefer to remove it?

msevestre: oh no I see... now. You cannot use the other one directly
Sthg like x<-getUnitEnum
x$Mass

that would not work right?

I feel that we are going to load those two enums on load (see @PavelBal other PR). So I am going to merge this as soon as we have tests for this and they we can consolidate with his work

abdullahhamadeh: Yes x$Mass would give the units, not "Mass"

PR #511, opened 2021-04-27
Parallel run

PavelBal:
Implementation of running simulations concurrently.

Concurrent batch simulations will follow in a separate PR.

FYI @abdelr


PavelBal:
Tests require the latest Core.

R/utilities-simulation.R
```diff
@@ -140,6 +140,67 @@ runSimulation <- function(simulation, population = NULL, agingData = NULL, simul
...
+  for (i in seq_along(results)) {
+    resultObject <- results[[i]]
+    id <- rClr::clrGet(resultObject, "Id")
+    succeeded <- rClr::clrGet(resultObject, "Succeeded")
+    if (succeeded) {
+      # Get the correct simulation and create a SimulationResults object
+      simulationResults[[id]] <- SimulationResults$new(ref = rClr::clrGet(resultObject, "Result"), simulation = simulationsIdMap[[id]])
+      next()
+    }
+    # If the simulation run failed, show a warning
```
msevestre, line 195: Should not we set simulationResults[[id]] to NULL or sthg here ? to show that something is off?

msevestre (commented):
I think this is looking good. Just a quick question regarding returned value in case of SIlentMode

R/utilities-simulation.R
PavelBal: All values are NULL after initialization of `simulationResults` and are overwritten if the result exists.

PR #513, opened 2021-05-05
Concurrent simulation batch

PavelBal:


PavelBal:
Here is a proposal how to use simulation batches:

1) Create SimulationBatch(es), defining the simulation to run and parameter and initial values to vary
```
sim1 <- loadTestSimulation("simple", loadFromCache = TRUE)
sim2 <- loadTestSimulation("simple2", loadFromCache = TRUE)
parameters <- c("Organism|Liver|Volume", "R1|k1")
molecules <- "Organism|Liver|A"

simulationBatch1 <- createSimulationBatch(simulation = sim1,
    parametersOrPaths = parameters,
    moleculesOrPaths = molecules)
simulationBatch2 <- createSimulationBatch(simulation = sim2,
    parametersOrPaths = parameters,
    moleculesOrPaths = molecules)
```

2) Add run values to simulation batches. Currently, only one ran value set can be added per method call - if required, it could be extended to passing multiple parameter and initial values sets. Every added run values set gets an ID that can be later be used to get the right result.

```
#Ids of run values
ids <- c()
ids[[1]] <- simulationBatch1$addRunValues(parameterValues = c(1, 2), initialValues = 1)
ids[[2]] <- simulationBatch1$addRunValues(parameterValues = c(1.6, 2.4), initialValues = 3)
ids[[3]] <- simulationBatch2$addRunValues(parameterValues = c(4, 2), initialValues = 4)
ids[[4]] <- simulationBatch2$addRunValues(parameterValues = c(2.6, 4.4), initialValues = 5)
```

3) Run the batches. The results are grouped by the simulation batches (i.e., the output is a list with a list of `SimulationResult` per batch). Every entry of `SimulationResult` is named by the ID of of run values set.

`res <- runSimulationBatches(simulationBatches = list(simulationBatch1, simulationBatch2))
resultsOfFirstBatch <- res[[1]]
resultsOfSecondBatch <- res[[2]]`

4) Running the batches removes all added run values, so next run would result in no results.

R/error-checks.R
```diff
@@ -8,32 +8,48 @@ isSameLength <- function(...) {
...
 #' @param object An object or a list of objects
-#' @param type String  representation or Class of the type that should be checked for
+#' @param type String representation or Class of the type that should be checked for
+#' @param nullAllowed Boolean flag if \code{NULL} is accepted for the \code{object}. If \code{TRUE},
+#' \code{NULL} always returns \code{TRUE}, otherwise \code{NULL} returns \code{FALSE}. Default is \code{FALSE}
 #'
 #' @return TRUE if the object or all objects inside the list are of the given type.
 #' Only the first level of the given list is considered.
-isOfType <- function(object, type) {
+isOfType <- function(object, type, nullAllowed = FALSE) {
```
PavelBal, line 17: I extended the `isOfType`-function by `nullAllowed` so NULLs within a list of objects can also be allowed.

Otherwise:
`isOfType(list("as", "dsads", NULL, "sd"), "character", nullAllowed = TRUE)` returns `FALSE`.

R/error-checks.R
```diff
@@ -8,32 +8,48 @@ isSameLength <- function(...) {
...
+  }
   if (inheritType(object)) {
     return(TRUE)
   }
 
   object <- c(object)
-  all(sapply(object, inheritType))
+  # Have to check for the lenght because otherwise sapply() always returns list()
+  # which is always TRUE
+  if (length(object) > 0) {
```
PavelBal, line 39: Otherwise

`isOfType(list(), "AnyType")` is `TRUE`.

R/error-checks.R
```diff
@@ -8,32 +8,48 @@ isSameLength <- function(...) {
...
+  }
+  return(FALSE)
 }
 
+#' Check if the provided object is of certain type. If not, stop with an error.
+#'
+#' @param object An object or a list of objects
+#' @param type String representation or Class of the type that should be checked for
+#' @param nullAllowed Boolean flag if \code{NULL} is accepted for the \code{object}. If \code{TRUE},
+#' \code{NULL} is always valid, otherwise the error is thrown. Default is \code{FALSE}
```
PavelBal, line 50: Just added documentation.

R/messages.R
```diff
@@ -1,21 +1,14 @@
 messages <- list(
   errorWrongType = function(objectName, type, expectedType, optionalMessage = NULL) {
-    # Name of the calling function
-    callingFunctions <- sys.calls()
-    callingFunction <- sys.call(-length(callingFunctions) + 1)[[1]]
-
+    callingFunction <- .getCallingFunctionName()
```
PavelBal, line 3: I moved the code for checking for the name of the calling function to a separate function to reduce code duplication.

R/simulation-batch.R
```diff
@@ -19,36 +20,74 @@ SimulationBatch <- R6::R6Class(
     initialize = function(ref, simulation) {
       validateIsOfType(simulation, Simulation)
       super$initialize(ref)
-      self$simulation <- simulation
+      private$.simulation <- simulation
```
PavelBal, line 23: Simulation should not be changed after creating the batch.

R/simulation-batch.R
```diff
@@ -19,36 +20,74 @@ SimulationBatch <- R6::R6Class(
...
 
-    #' @description Set the parameter and initial values in the simulation and run the simulation
+    #' @description Add a set of parameter and start values for next execution.
+    #' @details Intended for the use with \code{runSimulationBatches}. The simulation batch is executed
+    #' with the sets of parameter and initial values that have been scheduled. The set of run values is cleared after successful run.
+    #'
     #' @param parameterValues Vector of parameter values to set in the simulation (default is `NULL`)
     #' @param initialValues Vector of initial values to set in the simulation  (default is `NULL`)
-    #' @return A `SimulationResults` object containing the result of the simulation run
-    run = function(parameterValues = NULL, initialValues = NULL) {
```
PavelBal, line 29: No `run` for the simulation batch! Batches should always be run with `runSimulationBatches` function.

R/simulation-batch.R
```diff
@@ -19,36 +20,74 @@ SimulationBatch <- R6::R6Class(
...
-      batchRunValues <- SimulationBatchRunValues$new(parameterValues, initialValues)
+      # Only one values set is allowed - no lists of values
+      if (is.list(parameterValues) || is.list(initialValues)) {
+        stop(messages$errorOnlyOneValuesSetAllowed("parameterValues, initialValues"))
+      }
 
-      results <- rClr::clrCall(self$ref, "Run", batchRunValues$ref)
-      SimulationResults$new(results, self$simulation)
+      batchRunValues <- SimulationBatchRunValues$new(parameterValues, initialValues)
+      rClr::clrCall(self$ref, "AddSimulationBatchRunValues", batchRunValues$ref)
```
PavelBal, line 66: Adding run values returns an ID.

R/simulation-batch.R
```diff
@@ -19,36 +20,74 @@ SimulationBatch <- R6::R6Class(
...
     simulation = function(value) {
       if (missing(value)) {
         private$.simulation
       } else {
-        private$.simulation <- value
+        private$throwPropertyIsReadonly("simulation")
+      }
+    },
+    #' @field runValuesIds Ids of the run values that will be executed on next run
+    runValuesIds = function(value) {
```
PavelBal, line 86: Returns the IDs of run values enqueued for the next run.

R/utilities-simulation.R
```diff
@@ -168,36 +168,25 @@ runSimulationsConcurrently <- function(simulations, simulationRunOptions = NULL,
...
-      simulationResults[[id]] <- SimulationResults$new(ref = rClr::clrGet(resultObject, "Result"), simulation = simulationsIdMap[[id]])
-      next()
-    }
-    # If the simulation run failed, show a warning
-    if (!silentMode) {
-      errorMessage <- rClr::clrGet(resultObject, "ErrorMessage")
-      warning(errorMessage)
-    }
-  }
+  # Ids of the results are Ids of the simulations
```
PavelBal, line 185: The logic for converting results of concurrent runner to `SimulationResults` is moved to a separate function as it is also used by `runSimulationBatches`.

msevestre:
@PavelBal The build is failing
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r

PavelBal:
@msevestre Yes, it is using the old Core... And if I replace the libraries, it fails because of PK-Sim.

msevestre:
ok let me check this out

msevestre:
I am update the lib now to make sure we are using the right version of everything. Let' see

msevestre:
I was trying to fix this and it was not letting me push lol.. I understand now

R/simulation-batch.R
```diff
@@ -19,36 +20,74 @@ SimulationBatch <- R6::R6Class(
...
+    #' @return Id of the values set that can be used to get the correct result from \code{runSimulationBatches}.
+    #' @export
+    #'
+    #' @examples
+    #' \dontrun{
+    #' sim <- loadSimulation("simple", loadFromCache = TRUE)
+    #' parameters <- c("Organism|Liver|Volume", "R1|k1")
+    #' molecules <- "Organism|Liver|A"
+    #' # Create two simulation batches. In this case, they are based on the same simulation
+    #' # and vary the same paramter and initial values.
```
msevestre, line 42: typo: parameters*

R/simulation-batch.R
```diff
@@ -19,36 +20,74 @@ SimulationBatch <- R6::R6Class(
...
+    #' moleculesOrPaths = molecules)
+    #' simulationBatch2 <- createSimulationBatch(simulation = sim,
+    #' parametersOrPaths = parameters,
+    #' moleculesOrPaths = molecules)
+    #' #Ids of run values
+    #' ids <- c()
+    #' ids[[1]] <- simulationBatch1$addRunValues(parameterValues = c(1, 2), initialValues = 1)
+    #' ids[[2]] <- simulationBatch1$addRunValues(parameterValues = c(1.6, 2.4), initialValues = 3)
+    #' ids[[3]] <- simulationBatch2$addRunValues(parameterValues = c(4, 2), initialValues = 4)
+    #' ids[[4]] <- simulationBatch2$addRunValues(parameterValues = c(2.6, 4.4), initialValues = 5)
```
msevestre, line 54: unclear from this example why you would need TWO simulation batches and not only one with two differents sets.

What is the difference?

R/utilities-simulation.R
```diff
@@ -244,17 +236,94 @@ createSimulationBatch <- function(simulation, parametersOrPaths = NULL, molecule
...
+#' simulationBatch2 <- createSimulationBatch(simulation = sim,
+#' parametersOrPaths = parameters,
+#' moleculesOrPaths = molecules)
+#' #Ids of run values
+#' ids <- c()
+#' ids[[1]] <- simulationBatch1$addRunValues(parameterValues = c(1, 2), initialValues = 1)
+#' ids[[2]] <- simulationBatch1$addRunValues(parameterValues = c(1.6, 2.4), initialValues = 3)
+#' ids[[3]] <- simulationBatch2$addRunValues(parameterValues = c(4, 2), initialValues = 4)
+#' ids[[4]] <- simulationBatch2$addRunValues(parameterValues = c(2.6, 4.4), initialValues = 5)
+#' res <- runSimulationBatches(simulationBatches = list(simulationBatch1, simulationBatch2))
```
msevestre, line 280: can it be an array e.g. c(simulationBatch1, simulationBatch2)

R/utilities-simulation.R
```diff
@@ -244,17 +236,94 @@ createSimulationBatch <- function(simulation, parametersOrPaths = NULL, molecule
...
+  # SimulationBatches can be created with the same simulation
+  # Each SimulationBatchRunValues has its own id, which will be the id of the result
+  resultsIdSimulationIdMap <- list()
+  # Map of simulations ids to simulations objects
+  simulationIdSimulationMap <- vector("list", length(simulationBatches))
+  # Iterate through all simulation batches
+  for (simBatchIdx in seq_along(simulationBatches)) {
+    simBatch <- simulationBatches[[simBatchIdx]]
+    # Put the simulation into sim pointer id <-> sim map
+    simId <- deparse(rClr::clrGetExtPtr(simBatch$ref))
```
msevestre, line 302: WHAT IS THIS? LEt's please have a batch return an id or sthg.

R/utilities-simulation.R
```diff
@@ -244,17 +236,94 @@ createSimulationBatch <- function(simulation, parametersOrPaths = NULL, molecule
...
+  # Run the batch with the ConcurrentSimulationRunner
+  results <- rClr::clrCall(simulationRunner, "RunConcurrently")
+  simulationResults <- .getConcurrentSimulationRunnerResults(results = results, resultsIdSimulationIdMap = resultsIdSimulationIdMap, simulationIdSimulationMap = simulationIdSimulationMap, silentMode = silentMode)
+
+  # output: list of lists of SimulationResults, one list per SimulationBatch
+  output <- lapply(names(simulationIdSimulationMap), function(simId) {
+    simulationResults[which(resultsIdSimulationIdMap == simId)]
+  })
+  # Not sure this is required
+  rClr::clrCall(simulationRunner, "Dispose")
```
msevestre, line 322: there is an issue here. If this object is returned via a Task, it might be singleton and should NOT Be disposed.

R/utilities-simulation.R
```diff
@@ -387,3 +456,41 @@ exportIndividualSimulations <- function(population, individualIds, outputFolder,
...
+#'
+#' @return A named list of \code{SimulationResults} objects with the names being the ids of simulations or
+#' simulation-batch values pairs they were produced by
+.getConcurrentSimulationRunnerResults <- function(results, resultsIdSimulationIdMap, simulationIdSimulationMap, silentMode) {
+  # Pre-allocate lists for SimulationResult
+  simulationResults <- vector("list", length(results))
+  # Set the correct order of IDs
+  names(simulationResults) <- names(resultsIdSimulationIdMap)
+
+  for (i in seq_along(results)) {
```
msevestre, line 478: foreach or something if i i never used

R/utilities-simulation.R
msevestre: nice

msevestre (changes requested):
I think the API is nice. I am wondering about a few things

1-the ptr
2-the dispose

R/error-checks.R
```diff
@@ -43,12 +54,17 @@ validateIsOfType <- function(object, type, nullAllowed = FALSE) {
...
+#' @param value \code{value} to search for in the \code{enum}
+#' @param nullAllowed If TRUE, \code{value} can be \code{NULL} and the test always passes.
+#' If \code{FALSE} (default), NULL is not accepted and the test fails.
 validateEnumValue <- function(value, enum, nullAllowed = FALSE) {
-  if (nullAllowed && is.null(value)) {
-    return()
-  }
-
   if (is.null(value)) {
+    if (nullAllowed) {
```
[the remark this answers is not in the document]
PavelBal: @msevestre You decide which approach to take.

R/simulation-batch.R
PavelBal: Was just an example with multiple batches. But as it is `dontrun` anyways, I update the example with two different simulations.

R/utilities-simulation.R
PavelBal: `c()` is always a list for non-atomic types. I prefer to use `list` in the code when I know it will be casted to a list.

I wish R were strong typed...

R/utilities-simulation.R
PavelBal: :D Yes then we need to add an ID field to `OSPSuite.R.Services.SettingsForConcurrentRunSimulationBatch` in core.

R/utilities-simulation.R
PavelBal: Ok, I don't have experience with this. Removing the dispose.

R/utilities-simulation.R
PavelBal: Done.

PavelBal (commented):
@msevestre There are some open questions left - e.g. SimBatch should return an ID must be first implemented in Core. Please see my answers.

R/utilities-simulation.R
msevestre: we might need this. A task should not be disposable. It might be something I need to discuss with @abdelr

msevestre: in short,if this is disposable, we should get a new instance everytime.

msevestre: Ok I checked. Dispose is fine and should be call in that case explicitly. Not a big deal if not but it will release stuff immediately.

PR #516, opened 2021-05-10
Fixes #515

### LINKED ISSUE #515: uniqueEntities: Error when passing only one entity
```
> sim <- loadTestSimulation("S1")
>   parameter <- getParameter(toPathString(c("Organism", "Liver", "Pericentral", "Volume")), sim)
> uniqueEntities(parameter)
Error in entities[[i]] : wrong arguments for subsetting an environment
```

PavelBal:


R/utilities-molecule.R
```diff
@@ -107,7 +107,7 @@ setMoleculeInitialValues <- function(molecules, values) {
 #' setMoleculeScaleDivisors(molecules, c(0.002, 0.003))
 #' @export
 setMoleculeScaleDivisors <- function(molecules, values) {
-  molecules <- c(molecules)
+  molecules <- enforceIsList(molecules)
```
msevestre, line 110: Isn't it equivalent? What is the benefits of this function?

PavelBal: In this case it is. I just hate it that `c()` might return either a vector (e.g. `c(1, 2 , 3)`) or a list (in case the values are not atomic).

msevestre: oh I see. Ok. should we maybe just rename the method to something less violent? asList or toList?

PR #518, opened 2021-05-11
Using low-level rCrl methods in getOutputValues

PavelBal:


PavelBal:
See https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/519

tests/testthat/test-utilities-output-selections.R
```diff
@@ -34,7 +34,7 @@ test_that("It can add single output by reference", {
   quantities <- addOutputs(parameter, sim)
   expect_equal(length(quantities), 1)
   expect_equal(length(outputSelections$allOutputs), 1)
-  expect_equal(quantities[[1]], parameter)
```
PavelBal, line 37: I had to change the test because now the output is "Quantity" and not a "Parameter", and the test would fail.

tests/testthat/test-utilities-simulation-results.R
```diff
@@ -17,7 +17,7 @@ test_that("It throws an error when no valid simulation results are provided", {
 })
 
 test_that("It returns an array of NA if specific result is not found and stopIfNotFound = FALSE", {
-  res <- getOutputValues(individualResults, "NoPath", stopIfNotFound = FALSE)
+  res <- getOutputValues(simulationResults = individualResults, quantitiesOrPaths = "NoPath", stopIfNotFound = FALSE)
```
PavelBal, line 20: Let's try to always use argument names in test to capture errors if the signature of the method changes.

PavelBal:
@msevestre You can review now. I am using the "fromPath" methods implemented in Core.

R/utilities-simulation-results.R
```diff
@@ -18,14 +18,18 @@
...
+#' the performance of the function.
+#'
 #' @export
 getOutputValues <- function(simulationResults,
                             quantitiesOrPaths = NULL,
                             population = NULL,
                             individualIds = NULL,
-                            stopIfNotFound = TRUE) {
+                            stopIfNotFound = TRUE,
+                            withMetaData = TRUE) {
```
msevestre, line 32: I really don't like this. Return value structure driven by input parameters. It feels wrong to me.
Why do you need this? I thought the changes brought down the time to almost nothing.

PavelBal:
You have to teach me how to correctly update package versions...

tests/testthat/test-utilities-output-selections.R
msevestre: I don't understand this. Changing the internal of a function should not change the output

msevestre (changes requested):

msevestre:
how about this?
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/README.md#how-to-update-dependencies-from-nuget

msevestre:
Let me do this for you:)

R/utilities-simulation-results.R
PavelBal: For the tested use-case, not returning the meta data (resp. returning NULL for meta data) reduces the time from 1 sec to 500 msec.

I am not sure, but I would say for >90% of the use cases this meta data is not used, and in performance-critical scenarios (e.g. PI) this argument would allow to almost completely avoid communication with .NET. With the default `TRUE`, the code that is already present would behave exactly as it did.

I must admit, I never understood why this meta-data is required in the first place. If it is for TLF, why cannot TLF use ospsuite-r for this?

PavelBal:
> Let me do this for you:)

Better so :D

tests/testthat/test-utilities-output-selections.R
PavelBal: Actually now the output is more consistent, because it always returns a Quantity (or a set of quantities). Before, it would return a quantity if the paths are passed, but could return a Parameter or a Molecule in case one of those are passed.

R/utilities-output-selections.R
```diff
@@ -17,24 +20,32 @@
...
+    paths <- quantitiesOrPaths
   }
+  paths <- unique(paths)
 
-  quantities <- uniqueEntities(quantities, compareBy = "path")
-  outputSelections <- simulation$outputSelections
-
-  for (quantity in quantities) {
-    outputSelections$addQuantity(quantity)
+  task <- ospsuite:::getContainerTask()
```
msevestre, line 40: why do you use the namespace method within the package? I'd remove this

R/utilities-output-selections.R
```diff
@@ -17,24 +20,32 @@
...
+  for (path in paths) {
+    rClr::clrCall(task, "AddQuantitiesToSimulationOutputFromPath", simulation$ref, enc2utf8(path))
   }
 
+  quantities <- NULL
+  if (returnQuantities) {
```
msevestre, line 46: Same comment as before. Here we are changing return values based on input values (and even more than the other one) which feels very wrong
I do not know if we `really` need to return the quantity from this methods. Maybe it should just have no return values since it has side effects . I 'd rather break this API rather than implementing this kind of logic to be honest

msevestre (changes requested):

R/utilities-output-selections.R
PavelBal: Oops, must be copy-paste.. Will remove

R/utilities-output-selections.R
PavelBal: Returning the quantities here takes 10 seconds in the steady-state example, which is not acceptable.

So either do not return them at all, or create a separate method (`addOuputsSilent` or something), which I do not like at all..

tests/testthat/test-utilities-output-selections.R
msevestre: I see.. hum

R/utilities-simulation-results.R
msevestre: THe idea is that you can plot your outputs in a one liner in TLF if you know what entities you are dealing with (dimension and unit for instance). I can leave with this micro optimization but in general, we need to be mindful that those input driven return values are VERY hard to comprehend. This is typically a big no no and I am not sure if this is something specific to R or not.

R/utilities-output-selections.R
msevestre: @Yuri05 I would vote for not returning the Outptus. If the user wants them, it's a one liner. I want to avoid complex APIs where not required.

R/utilities-simulation-results.R
PavelBal: I agree that in general this is not a good concept. Just in this case it is a feature that costs performance in a potentially performance-sensitive scenario, while it is not used in >90% of cases. I cannot imagine why someone would plot simulation results without loading ospsuite-r.

For TLF, a method that adds simulation results to a DataMapping can be implemented that queries the units paths, like

`DataMapping$addSimulationResults(simulationResults, paths)`

R/utilities-quantity.R
```diff
@@ -96,6 +96,34 @@ setQuantityValues <- function(quantities, values) {
...
+#' @param simulation Simulation uses to retrieve quantity instances from given paths.
+#' @examples
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' setQuantityValuesByPath("Organism|Liver|Volume", 1, sim)
+#'
+#' setParameterValuesByPath(list("Organism|Liver|Volume", "Organism|Liver|A"), c(2, 3), sim)
+#' @export
+setQuantityValuesByPath <- function(quantityPaths, values, simulation) {
```
PavelBal, line 114: @msevestre Implementation using the .NET method for setting values by path

R/utilities-parameter.R
```diff
@@ -120,11 +120,8 @@ setParameterValues <- function(parameters, values) {
 #' setParameterValuesByPath(c("Organism|Liver|Volume", "Organism|Volume"), c(2, 3), sim)
 #' @export
 setParameterValuesByPath <- function(parameterPaths, values, simulation) {
-  validateIsString(parameterPaths)
-  validateIsNumeric(values)
-  validateIsOfType(simulation, Simulation)
-  parameters <- sapply(parameterPaths, function(p) getParameter(p, simulation))
-  setParameterValues(parameters, values)
+  setQuantityValuesByPath(quantityPaths = parameterPaths, values = values,
```
PavelBal, line 123: `setParameterValuesByPath `can simply use `setQuantityValuesByPath`

PavelBal:
Open questions:

1) `withMetaData ` argument in `getOutputValues `
2) `addOutputs` - argument `returnQuantities` introduced - discard? Do not return quantities at all (would require a not in release notes)
2) New function `setQuantityValuesByPath`

R/messages.R
```diff
@@ -67,6 +67,12 @@ messages <- list(
   errorPKParameterNotFound = function(pkParameterName, allPKParameterNames) {
     paste0("PK-Parameter '", pkParameterName, "' not found.\nAvailable PK-Parameters are:\n", paste0(allPKParameterNames, collapse = ", "))
   },
+  errorEntityPathNotAbsolute = function(path) {
+    paste0(
+      "Only absolut paths (i.e. without the wildcard(s) `*`) are allowed, but the passed path is: ",
```
msevestre, line 72: but the `given` path maybe?

R/messages.R
```diff
@@ -91,6 +97,6 @@ formatNumerics <- function(numerics, digits = ospsuiteEnv$formatNumericsDigits,
 
 .getCallingFunctionName <- function() {
   callingFunctions <- sys.calls()
-  callingFunction <- sys.call(-length(callingFunctions) + 1)[[1]]
+  callingFunction <- sys.call(-length(callingFunctions) + 2)[[1]]
```
msevestre, line 100: wow..what is this?

R/utilities-parameter.R
```diff
@@ -120,11 +120,10 @@ setParameterValues <- function(parameters, values) {
 #' setParameterValuesByPath(c("Organism|Liver|Volume", "Organism|Volume"), c(2, 3), sim)
 #' @export
 setParameterValuesByPath <- function(parameterPaths, values, simulation) {
-  validateIsString(parameterPaths)
-  validateIsNumeric(values)
-  validateIsOfType(simulation, Simulation)
-  parameters <- sapply(parameterPaths, function(p) getParameter(p, simulation))
-  setParameterValues(parameters, values)
+  setQuantityValuesByPath(
+    quantityPaths = parameterPaths, values = values,
```
msevestre, line 124: I'd say either we put all parameters on the same line or one for each line. Here we are mixing and matching. Why?

R/utilities-quantity.R
```diff
@@ -96,6 +96,36 @@ setQuantityValues <- function(quantities, values) {
...
+  validateIsString(quantityPaths)
+  validateIsNumeric(values)
+  validateIsSameLength(quantityPaths, values)
+  validateIsOfType(simulation, Simulation)
```
msevestre, line 118: don't you want to use your `validatePathIsAbsolute` method?

R/messages.R
PavelBal: Done.

R/utilities-simulation-results.R
```diff
@@ -77,18 +80,29 @@ getOutputValues <- function(simulationResults,
...
+  task <- getContainerTask()
+  metaData <- NULL
+  if (addMetaData) {
+    metaData <- lapply(paths, function(path) {
+      unit <- NULL
+      dimension <- NULL
+      # Get the dimension and unit from path if the results are obtained. If the results
+      # are NA, the entity with such path does not exist
+      if (!all(is.na(values[[path]]))) {
+        unit <- rClr::clrCall(task, "BaseUnitNamesFromPath", simulationResults$simulation$ref, enc2utf8(path))[[1]]
```
msevestre, line 98: I am wondering if this function should not return only values for one path as opposed as for an array.
It looked suspicious to me before but seeing how it's used now...I think the implementation in .NET should only work with one path. Or do you expect to have wild cards?

msevestre (changes requested):

R/messages.R
PavelBal: Ok that didn't work, reverting.

R/utilities-parameter.R
PavelBal: Hmm, I relied on styler on that..

R/utilities-quantity.R
PavelBal: I implemented `validatePathIsAbsolute` before you implemented the check in .NET. I did not remove the validation as it might be used in other places later, but for this case it is covered by .NET, no?

R/utilities-parameter.R
msevestre: ok. if it's automated then I am fine

R/utilities-simulation-results.R
PavelBal: Good point, I have to check what happens when passing a path with wildcards. Will do it tomorrow.

PavelBal: According to the description, 

> #' When providing the paths, only absolute full paths are supported (i.e., no matching with '*' possible).

And the called .NET method is `public virtual float[] AllValuesFor(string quantityPath, params int[] individualIds)` of `SimulationResults`, which probably only accepts absolute paths.

Actually when using the wildcard, the path is not transformed to a regexp and the error is 

`getOutputValues: No results exists for path 'Organism|PeripheralVenousBlood|*|Plasma (Peripheral Venous Blood)' for individual IDs '0'! `



So maybe the funciton in .NET should always return only one value.

msevestre: ok. Going to update the .net side now

msevestre:
@PavelBal I updated to use single version of the methods . IsExplicitFormulaFromPath is never used anywhere so I am not sure where this is going to break

PavelBal:
@msevestre OK I think we are good to merge.

msevestre:
@PavelBal Be mindful of method rename and behavior. IsExplicitFormulasFromPath => IsExplicitFormulaByPath

PR #526, opened 2021-05-25
Add getAllStateVariablesPaths

PavelBal:


R/utilities-simulation.R
```diff
@@ -425,6 +425,22 @@ getAllParametersForSensitivityAnalysisMatching <- function(paths, simulation) {
...
+#' Get the paths of all state variable quantities of the simulation
+#'
+#' @param simulation \code{Simulation} object
+#' @details List of paths of all molecules in all compartments and all parameters that are
+#' state variables.
+#'
+#' @return A list of paths
+#' @export
+getAllStateVariablesPaths <- function(simulation) {
+  validateIsOfType(simulation, type = "Simulation")
```
msevestre, line 437: you can pass a ref to the R6 class directly here (e..g validateIsOfType(simulation, Simulation) which removes the use of "magic" string

R/utilities-simulation.R
```diff
@@ -425,6 +425,22 @@ getAllParametersForSensitivityAnalysisMatching <- function(paths, simulation) {
...
+  allMoleculesPaths <- getAllEntityPathsIn(container = simulation, entityType = Molecule)
```
msevestre, line 438: getAllMoleculePathsIn instead. Since we have a convenience method for this, why not use it?

msevestre (changes requested):

R/utilities-simulation.R
PavelBal: Done

R/utilities-simulation.R
PavelBal: Done

PR #527, opened 2021-05-27
Update core

msevestre:


packages.centOS7.config
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="OSPSuite.FuncParser.CentOS7" version="4.0.0.50"/>
-  <package id="OSPSuite.SimModel.CentOS7" version="4.0.0.42"/>
+  <package id="OSPSuite.SimModel.CentOS7" version="4.0.0.43"/>
```
msevestre, line 4: @Yuri05 Where is this binaries coming from? I don't see it on appveyor

Yuri05: At the moment I create CentOS nuget packages **manually**, then push them into https://github.com/Open-Systems-Pharmacology/OSPSuite.CentOS where they are published on AppVeyor nuget feed.

Can be improved with GitHub-Actions one day (or if AppVeyor adds the CentOS build environment https://github.com/appveyor/ci/issues/2333 )

PR #532, opened 2021-05-31
Fixes #531 version 4

### LINKED ISSUE #531: Target version 4 of R

msevestre:


DESCRIPTION
```diff
@@ -17,7 +17,7 @@ BugReports:
     https://github.com/open-systems-pharmacology/ospsuite-r/issues
 Depends: 
     R (>= 3.5),
-    rClr (>= 0.8.3)
+    rClr (>= 0.9.1)
```
msevestre, line 20: @PavelBal @Yuri05  Is this what we need to do?
Target 0.9.1 but say that we depend on R 3.5?
Appveyor will build with v4.0.x

msevestre: let's undo this

DESCRIPTION
```diff
@@ -9,15 +9,15 @@ Authors@R:
...
+Description: The ospsuite-R package provides the functionality of loading, manipulating, and simulating the simulations 
+  created in the Open Systems Pharmacology Software tools PK-Sim and MoBi.
 License: GPL-2 | file LICENSE
 URL: https://github.com/open-systems-pharmacology/ospsuite-r
 BugReports: 
     https://github.com/open-systems-pharmacology/ospsuite-r/issues
 Depends: 
     R (>= 3.5),
-    rClr (>= 0.9.1)
+    rClr (>= 0.8.3)
```
PavelBal: Let's use 0.9.1 as we now have releases of the version for R3 and R4, and 0.9.1 has a bug fix that 0.8.3 does not.

msevestre: YEs of course. Good idea

PR #543, opened 2021-07-01
Fixes #541 extend set parameter value by path

### LINKED ISSUE #541: extend `setParameterValuesByPath` function
Extend the function `setParameterValuesByPath` with an optional argument `ignoreNonExisting ` (or something like that).
The new argument should be **false** per default. 
If set to **true**: `setParameterValuesByPath `should skip non existing parameters with a warning and not exit with error.

Such a behaviour might be required in situations like [here](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/497)

msevestre:


tests/testthat/test-utilities-quantity.R
```diff
@@ -73,3 +73,11 @@ test_that("It throws an exception when setting values for a quantity that does n
   parameterPath <- "Organism|Liver|NOPE|Volume"
   expect_that(setQuantityValuesByPath(parameterPath, 100, sim), throws_error())
 })
+
+test_that("It does not throw an exception when setting values for a quantity that does not exist and the stopIfnotFound flag is set to false", {
+  sim <- loadTestSimulation("S1", loadFromCache = TRUE)
+  parameterPath <- "Organism|Liver|NOPE|Volume"
+  setQuantityValuesByPath(parameterPath, 100, sim, FALSE)
+  expect_false(is.null(sim))
```
msevestre, line 81: @Yuri05 We need an expect for this test to be found .I just want to make sure that the line before does not throw

Issue #544, opened 2021-07-01
Add dataSet wrapper around DataRepository


msevestre:
See https://github.com/Open-Systems-Pharmacology/OSPSuite-R/discussions/538

PR #546, opened 2021-07-02
WIP #544

msevestre:


msevestre:
This is just the beginning... and it should not be merged at all

R/data-repository.R
```diff
@@ -7,16 +7,20 @@ DataRepository <- R6::R6Class(
...
         if (is.null(private$.baseGrid)) {
-          private$.baseGrid <- DataColumn$new(private$wrapReadOnlyProperty("BaseGrid", value))
+          private$.baseGrid <- DataColumn$new(private$wrapProperty("BaseGrid", value))
         }
         return(private$.baseGrid)
       }
-
-      private$throwPropertyIsReadonly("baseGrid")
+      private$.baseGrid <- value
+      private$wrapProperty("BaseGrid", value$ref)
```
msevestre, line 23: Exposing now some of the Readonly properties

R/data-set.R
```diff
@@ -0,0 +1,80 @@
...
+    },
+    #' @field xValues Values stored in the xUnit dimension (not necessarily in the base unit of the dimension)
+    xValues = function(value){
+      if (missing(value)) {
+        return ( self$dataRepository$baseGrid$values)
+      }
+    }
+  ),
+  public = list(
+    dataRepository = NULL,
```
msevestre, line 38: Public for now so that I can debug stuff. It will be private once the implementation is stable

R/data-set.R
```diff
@@ -0,0 +1,80 @@
...
+  active = list(
+    #' @field name The name of the DataSet
+    name = function(value) {
+      if (missing(value)) {
+        return(self$dataRepository$name)
+      }
+      self$dataRepository$name <- value
+    },
+    #' @field xUnit Unit in which the xValues are defined
+    xUnit = function(value){
```
msevestre, line 19: This is not the real implementation. Just dummies for now for all methods

R/data-set.R
```diff
@@ -0,0 +1,80 @@
...
+    }
+  ),
+  private = list(
+    .xValues = NULL,
+    .yValues = NULL,
+    createDataRepository = function() {
+      # Create an empty data repository with a base grid and column
+      dataRepository <- DataRepository$new()
+      # Passing time for dimension for now
+      xValues <- DataColumn$new(rClr::clrNew("OSPSuite.Core.Domain.Data.BaseGrid", "xValues", getDimensionByName("Time")))
```
msevestre, line 62: I think it makes sense to initialize with some kind of defaults (time, and conc for instance)

msevestre: I will use our tasks to get the name instead of using magic string

msevestre:
@PavelBal Tagging @hannaei as well here for review

R/data-column.R
```diff
@@ -10,9 +10,9 @@ DataColumn <- R6::R6Class(
   inherit = DotNetWrapper,
   cloneable = FALSE,
   active = list(
-    #' @field values Returns the values defined in the column  (Read-Only)
+    #' @field values Returns the values defined in the column
     values = function(value) {
-      private$wrapReadOnlyProperty("ValuesAsArray", value)
+      private$wrapProperty("ValuesAsArray", value)
```
msevestre, line 15: I updated core to support setting values as well.

R/data-column.R
```diff
@@ -23,14 +23,31 @@ DataColumn <- R6::R6Class(
...
     displayUnit = function(value) {
-      private$wrapExtensionMethod(WITH_DISPLAY_UNIT_EXTENSION, "DisplayUnitName", "displayUnit", value)
+      if (missing(value)) {
+        return(private$wrapExtensionMethod(WITH_DISPLAY_UNIT_EXTENSION, "DisplayUnitName", "displayUnit", value))
+      }
+      dimension <- getDimensionByName(self$dimension)
+      # we use the ignore case parameter set  to true so that we do not have to worry about casing when set via scripts
+      unit <-  rClr::clrCall(dimension, "FindUnit", value, TRUE)
+      if(is.null(unit)){
+        stop(messages$errorUnitNotSupported(unit = value, dimension = self$dimension))
```
msevestre, line 35: Right now, an error will be thrown. This is OK in the DataColun object. In the DataSet one, we could be a bit more user friendly (see comment below)

R/data-column.R
```diff
@@ -23,14 +23,31 @@ DataColumn <- R6::R6Class(
...
-      return(private$.dimension)
+      if (missing(value)) {
+        if (is.null(private$.dimension)) {
+          private$.dimension <- private$wrapExtensionMethodCached(WITH_DIMENSION_EXTENSION, "DimensionName", "dimension", private$.dimension, value)
+        }
+        return(private$.dimension)
+      }
+      # updating the dimension
+      rClr::clrSet(self$ref, "Dimension", getDimensionByName(value))
+      private$.dimension <- NULL
```
msevestre, line 49: For efficient usage, we are caching dimension and unit. We need to reset the cache obviously when updating the dimension

R/data-set.R
```diff
@@ -0,0 +1,141 @@
...
+        return(self$dataRepository$name)
+      }
+      self$dataRepository$name <- value
+    },
+    #' @field xDimension Dimension in which the xValues are defined
+    xDimension = function(value) {
+      if (missing(value)) {
+        return(private$.xValues$dimension)
+      }
+      private$setColumnDimension(private$.xValues, value)
```
msevestre, line 23: I have define a method for each `set`a ction so that we can call it for x, y and later error

R/data-set.R
```diff
@@ -0,0 +1,141 @@
...
+    #' @param dataRepository Instance of the \code{DataRepository} object to wrap
+    #' @return A new `DotNetWrapper` object.
+    initialize = function(dataRepository = NULL) {
+      self$dataRepository <- dataRepository %||% private$createDataRepository()
+      private$initializeCache()
+    },
+    #' @description
+    #' Print the object to the console
+    #' @param ... Rest arguments.
+    print = function(...) {
```
msevestre, line 76: @hannaei and @PavelBal 
Here the print method should be enhanced to show the data in a user friendly fashion. I'll leave that to you

R/data-set.R
```diff
@@ -0,0 +1,141 @@
...
+      # no need to update anything if we are setting the same values
+      if(column$displayUnit == unit){
+        return()
+      }
+
+      # save the values in their display unit before updating
+      values <- private$getColumnValues(column)
+
+      #now we need to update dimension and display unit
+      column$displayUnit <- unit
```
msevestre, line 118: HERE: If the unit is not supported, we could easily update the dimension by using the method getDimensionForUnit. That way, the user would never have to worry about updating the dimension when updating the unit.

R/data-set.R
```diff
@@ -0,0 +1,141 @@
...
+      yValues <- DataColumn$new(rClr::clrNew("OSPSuite.Core.Domain.Data.DataColumn", "yValues", getDimensionByName(ospDimensions$`Concentration (mass)`), xValues$ref))
+
+      dataRepository$addColumn(xValues)
+      dataRepository$addColumn(yValues)
+      return(dataRepository)
+    },
+    initializeCache = function() {
+      private$.xValues <- self$dataRepository$baseGrid
+      # TODO we need to be a bit more careful here
+      private$.yValues <- self$dataRepository$allButBaseGrid[[1]]
```
msevestre, line 138: There is a more to do. Specifically regarding the error column.

msevestre (commented):
@hannaei @PavelBal @Yuri05 
Please review. I think this is in a good place for @hannaei or @PavelBal  to continue now. I have added capabilities in core to update dimension and unit

R/data-set.R
PavelBal: @msevestre I would not prefer to automatically change the dimension, as I see it as a sanity check here. Simply changing the unit only affects what actual values the data have, but not the meaning of the data. Changing the dimension actually means that the data means something else. What do you think?

PavelBal:
@msevestre Should we remove "utilities-data-mapping" and completely replace it by "utilities-data-set"? E.g. `loadDataRepositoryFromPKML` would become `loadDataSetFromPKML` and create a data set object.

PavelBal:
Should we maybe fork this so we can merge PRs? I do not want to have a huge PR without intermediate states.

R/data-repository.R
```diff
@@ -56,12 +58,60 @@ DataRepository <- R6::R6Class(
...
       private$printClass()
       invisible(self)
+    },
+    #' @description
+    #' Adds a new entry to meta data list or changes its value if the name is already present.
+    #' If \code{value} is \code{NULL}, the entry with corresponding name is deleted from meta data set.
+    #'
+    #' @param name Name of new meta data list entry
+    #' @param value Value of new meta data list entry
+    addMetaData = function(name, value) {
```
msevestre, line 90: @PavelBal Add and remove meta data as separate function in data repo.

msevestre: there was this comment
    #' If \code{value} is \code{NULL}, the entry with corresponding name is deleted from meta data set.

That I don't want to have. Let's create two methods instead

R/data-repository.R
```diff
@@ -56,12 +58,60 @@ DataRepository <- R6::R6Class(
...
+      if (length(name) != 1) {
+        stop(messages$errorMultipleMetaDataEntries())
+      }
+      validateIsString(name)
+      validateIsString(value)
+      dataRepositoryTask <- getNetTask("DataRepositoryTask")
+      rClr::clrCall(dataRepositoryTask, "AddMetaData", self$ref, name, value)
```
msevestre, line 97: is there a way to coherce a parameter to string in R. If yes, then we can remove the need for value to be a string

R/data-set.R
```diff
@@ -0,0 +1,277 @@
...
+        return(private$.getColumnValues(private$.yColumn))
+      }
+      private$throwPropertyIsReadonly("yValues")
+    },
+
+    #' @field yErrorType Type of the error - geometric or arithmetic.
+    #' When changing from arithmetic to geometric error, the values are considered in as fraction (1 = 100%).
+    #' When changing from geometric to arithmetic, the values are set to the same unit as \code{yErrorUnit}.
+    yErrorType = function(value) {
+      private$.createErrorColumnIfMissing()
```
msevestre, line 74: when accessing Error properties, I create the column if it's not there. The way it was implemented @PavelBal , you would modify the original repo if loaded from file and I don't think this is pure

R/utilities.R
```diff
@@ -67,3 +72,15 @@ toList <- function(object) {
...
+
+
+#' Retrieves the name of the constant in the specified enumeration that has the specified value.
+#'
+#' @inheritParams rClr::clrGetEnumNames
+#' @param enumValue The value of a particular enumerated constant in terms of its underlying type. Typically an integer.
+#'
+#' @return A string containing the name of the enumerated constant in enumType whose value is enumValue; or null if no such constant is found.
+netEnumName <- function(enumType, enumValue){
+  netTypeObj <- rClr::clrGetType(enumType)
```
msevestre, line 84: @PavelBal also fixed this

tests/testthat/test-data-set.R
```diff
@@ -0,0 +1,144 @@
...
+  dataSet <- DataSet$new()
+  expect_equal(dataSet$yValues, numeric(0))
+})
+
+test_that("empty y error", {
+  dataSet <- DataSet$new()
+  expect_equal(dataSet$yErrorValues, numeric(0))
+})
+
+test_that("it can print a data set", {
```
msevestre, line 99: @PavelBal Also fixed this

msevestre (commented):
@PavelBal I think this is good to review ...

R/data-repository.R
PavelBal: `as.character()` works e.g. for numerics but not for class objects, so maybe leave the check here.

R/data-set.R
PavelBal: I don't really understand... Now the original repo is modified the first time any property of yError is called, no?

R/utilities-data-set.R
```diff
@@ -0,0 +1,46 @@
+
+makeColumn <- function(dataSet, property) {
```
msevestre, line 2: This function is intended to be private it seems. Let make it start with a "."

R/utilities-data-set.R
```diff
@@ -0,0 +1,46 @@
...
+# }
+
+#' Converts a list of DataSet objects to a data.frame
+#'
+#' @param dataSets A list of DataSet objects or a single DataSet
+#'
+#' @return DataSet objects as dataframe with columns name, xValue, yValue, yErrorValues,
+#' xDimension, xUnit, yDimension, yUnit, yErrorType, yErrorUnit, yMolWeight
+#' @export
+dataSetToDataFrame <- function(dataSets) {
```
msevestre, line 28: we need tests for this method, at least just to make sure that this does not crash with updated DataSet class

msevestre:
@PavelBal DLL updated to support the RemoveColumn

PR #564, opened 2021-07-26
Wip data set

PavelBal:


R/data-set.R
```diff
@@ -1,7 +1,15 @@
...
+DataErrorType <- enum(c(
+  "ArithmeticStdDev",
+  "GeometricStdDev"
+))
+
 #' @title DataSet
 #' @docType class
-#' @description  A wrapper around DataRepository exposing convenience methods to use and manipulate dataSets
-#' (typically observed data) containing an X column, a Y column and potentially an Error columns
+#' @description  A class for storage of numerical x- and y-value pairs and optional error for y-values.
```
PavelBal, line 11: I would prefer to use as few `DataRepository` in user-exposed class as possible.

R/data-set.R
```diff
@@ -11,38 +19,39 @@ DataSet <- R6::R6Class(
...
-      self$dataRepository$name <- value
+      private$.dataRepository$name <- value
     },
     #' @field xDimension Dimension in which the xValues are defined
     xDimension = function(value) {
       if (missing(value)) {
         return(private$.xValues$dimension)
       }
-      private$setColumnDimension(private$.xValues, value)
+      private$.setColumnDimension(private$.xValues, value)
```
PavelBal, line 31: I suggest using a dot in the names of private functions as well.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
       private$.yValues$displayUnit <- value
     },
-    #' @field yValues Values stored in the yUnit dimension (not necessarily in the base unit of the dimension)
+    #' @field yValues Values stored in the yDimension dimension (not necessarily in the base unit)
     yValues = function(values) {
       if (missing(values)) {
-        return(private$getColumnValues(private$.yValues))
+        return(private$.getColumnValues(private$.yValues))
+      }
+      validateIsSameLength(self$xValues, values)
```
PavelBal, line 68: We have to find a proper way to synchronize the lengths of base grid and data columns, it's pure pain!

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+      private$.setColumnValues(private$.yValues, values)
+    },
+
+    #' @field yErrorType Type of the error - geometric or arithmetic.
+    #' When changing from arithmetic to geometric error, the values are considered in as fraction (1 = 100%).
+    #' When changing from geometric to arithmetic, the values are set to the same unit as \code{yErrorUnit}.
+    yErrorType = function(value) {
```
PavelBal, line 75: Returning a string representation of the error type.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+      private$.yErrorValues$displayUnit <- value
+    },
+    #' @field yErrorValues Values of error stored in the yErrorUnit unit
+    yErrorValues = function(values) {
+      if (missing(values)) {
+        return(private$.getColumnValues(private$.yErrorValues))
       }
 
-      private$setColumnValues(private$.yValues, values)
+      validateIsSameLength(self$yValues, values)
```
PavelBal, line 97: Same here - must synchronize the lengths of error values and y values columns.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+      private$.setColumnValues(private$.yErrorValues, values)
+    },
+
+    #' @field metaData Returns a named list of meta data defined for the data set.
+    metaData = function(value) {
```
PavelBal, line 102: We should be able to get meta data but also add entries. This is the proposed solution where `metaData` just returns a list of all entries.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+    #'
+    #' @param name Name of new meta data list entry
+    #' @param value Value of new meta data list entry
+    addMetaData = function(name, value) {
+      if (length(name) != 1) {
+        stop(messages$errorMultipleMetaDataEntries())
+      }
+      validateIsString(name)
+
+      # TODO
```
PavelBal, line 132: Not implemented yet - requires the last Core .dll. And I am not sure how the deletion will work.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+    },
+
     #' @description
     #' Print the object to the console
     #' @param ... Rest arguments.
     print = function(...) {
-      private$printClass()
+      private$.printClass()
```
PavelBal, line 139: Something still failing in print().

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
 
       #now we need to update dimension and display unit
       column$displayUnit <- unit
 
-      private$setColumnValues(column, values)
+      private$.setColumnValues(column, values)
     },
-    createDataRepository = function() {
-      # Create an empty data repository with a base grid and column
+    .setErrorType = function(errorType){
```
PavelBal, line 193: Not tested yet at all!

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+      #If error type does not change, do nothing
+      if (errorType == self$yErrorType){
+        invisible(self)
+      }
+      browser()
```
PavelBal, line 198: Ups, will remove this in the next commit.

tests/testthat/test-data-column.R
```diff
@@ -27,3 +35,7 @@ test_that("it throws an error when setting a unit thast does not exist in the di
   baseGrid$dimension <- ospDimensions$Time
   expect_error(baseGrid$displayUnit <- ospUnits$Amount$mmol)
 })
+
+test_that("it can print a data column", {
```
PavelBal, line 39: We should also always check print(), we haven't been thinking about it.

tests/testthat/test-data-set.R
```diff
@@ -60,3 +59,58 @@ test_that("it can update the dimension of the yValues and this does not change t
...
+  expect_equal(dataSet$xValues, c(1, 2, 3, 4, 5))
+  expect_equal(dataSet$xUnit, ospUnits$Time$`week(s)`)
+})
+
+test_that("it can set the yValues if yUnit is different from base unit", {
+  dataSet <- DataSet$new()
+  dataSet$xValues <- c(1, 2, 3, 4, 5)
+  dataSet$yUnit <- "mg/l"
+  dataSet$yValues <- c(1, 2, 3, 4, 5)
+  #TODO this test fails because of numerical error in the conversion if default tolerance is used
```
PavelBal, line 76: Conversion between units introduces some numerical error...

tests/testthat/test-data-set.R
```diff
@@ -60,3 +59,58 @@ test_that("it can update the dimension of the yValues and this does not change t
...
+  expect_equal(dataSet$yValues, c(1, 2, 3, 4, 5), tolerance = 0.0001)
+  expect_equal(dataSet$yUnit, "mg/l")
+})
+
+test_that("it can change yErrorType when no yError is defined", {
+  dataSet <- DataSet$new()
+  expect_equal(dataSet$yErrorType, DataErrorType$ArithmeticStdDev)
+  dataSet$yErrorType <- DataErrorType$GeometricStdDev
+
```
PavelBal, line 85: Failing because getting yValues or yErrorValues does not work if the values are not set.

tests/testthat/test-data-set.R
```diff
@@ -60,3 +59,58 @@ test_that("it can update the dimension of the yValues and this does not change t
...
+  dataSet$yValues <- c(1, 2, 3, 4, 5)
+  #TODO this test fails because of numerical error in the conversion if default tolerance is used
+  expect_equal(dataSet$yValues, c(1, 2, 3, 4, 5), tolerance = 0.0001)
+  expect_equal(dataSet$yUnit, "mg/l")
+})
+
+test_that("empty y values", {
+  #TODO FAILING!
+  dataSet <- DataSet$new()
+  dataSet$yValues
```
PavelBal, line 98: This is what I am talking about.


```
>   dataSet <- DataSet$new()
>   dataSet$yValues
 Error in clrCallStatic(clrFacadeTypeName, "GetFieldOrProperty", objOrType,  : 
  Type:    System.ArgumentNullException
Message: Value cannot be null.
Parameter name: source
Method:  System.Collections.Generic.IEnumerable`1[TResult] Select[TSource,TResult](System.Collections.Generic.IEnumerable`1[TSource], System.Func`2[TSource,TResult])
Stack trace:
   at System.Linq.Enumerable.Select[TSource,TResult](IEnumerable`1 source, Func`2 selector)
   at OSPSuite.Core.Extensions.ArrayExtensions.ToDoubleArray(IEnumerable`1 floats) 
8.
clrCallStatic(clrFacadeTypeName, "GetFieldOrProperty", objOrType, 
    name) 
7.
rClr::clrGet(self$ref, propertyName) at dot-net-wrapper.R#35
6.
private$wrapProperty("ValuesAsArray", value) at data-column.R#15
5.
(function (value) 
{
    private$wrapProperty("ValuesAsArray", value)
})() at error-checks.R#92
4.
validateIsNumeric(values) at utilities-units.R#133
3.
toUnit(quantityOrDimension = column$dimension, values = column$values, 
    targetUnit = column$displayUnit) at data-set.R#163
2.
private$.getColumnValues(private$.yValues) at data-set.R#66
1.
(function (values) 
{
    if (missing(values)) {
        return(private$.getColumnValues(private$.yValues)) ... 

```

PavelBal:
@msevestre please review. Still lots of tests to write, I stopped because I don't know what is the best way to deal with

1) Empty yValues
2) Changing the length of xValues - should we automatically create new (empty) yValues and yErrorValues based on the new base grid?

R/data-set.R
PavelBal: Make sure that x and y values always have the same length. Will be covered by setting xValues yValues yErrorValues to readonly and implementing a separate method to set values.

tests/testthat/test-data-set.R
PavelBal: No its fine to adjust the tolerance.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+        errorTypeEnumVal <- rClr::clrGet(dataInfo, "AuxiliaryType")
+        return(netEnumName("OSPSuite.Core.Domain.Data.AuxiliaryType", errorTypeEnumVal))
+      }
+      private$.setErrorType(value)
+    },
+    #' @field yErrorUnit Unit in which the yErrorValues are defined. For arithmetic error, the unit must be valid
+    #' for \code{yDimension}. For geometric error, the unit must be valid for \code{Dimensionless}.
+    yErrorUnit = function(value) {
+      if (missing(value)) {
+        return(private$.yErrorValues$displayUnit)
```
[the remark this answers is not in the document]
PavelBal: I chose the same naming as for xValues and yValues - there we also have the active field xValues and yValues and the private fields .xValues/.yValues. But indeed, the private fields are columns, so maybe rename them to .xColumn, .yColumn, .yErrorColumn.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
+      # TODO
+    },
+
     #' @description
     #' Print the object to the console
     #' @param ... Rest arguments.
     print = function(...) {
-      private$printClass()
+      private$.printClass()
+      private$printLine("X dimension", c(self$xDimension))
```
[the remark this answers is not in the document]
PavelBal: Erm probably  a leftover from something... I have to adjust the print method.

R/data-set.R
```diff
@@ -51,76 +60,166 @@ DataSet <- R6::R6Class(
...
     .xValues = NULL,
     .yValues = NULL,
-    setColumnValues = function(column, values) {
+    .yErrorValues = NULL,
+    .setColumnValues = function(column, values) {
       # values are set in the display unit. We need to make sure we convert them to the base unit
       valuesInBaseUnit <- toBaseUnit(quantityOrDimension = column$dimension, values = values, unit = column$displayUnit)
       column$values <- valuesInBaseUnit
-      invisible()
+      invisible(self)
```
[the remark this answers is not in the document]
PavelBal: Following the guidelines from here https://adv-r.hadley.nz/r6.html#method-chaining

`invisible()` simply returns `NULL`, which does not make much sense, does it?

R/utilities.R
```diff
@@ -72,3 +72,15 @@ toList <- function(object) {
...
+
+
+#' Retrieves the name of the constant in the specified enumeration that has the specified value.
+#'
+#' @inheritParams rClr::clrGetEnumNames
+#' @param enumValue The value of a particular enumerated constant in terms of its underlying type. Typically an integer.
+#'
+#' @return A string containing the name of the enumerated constant in enumType whose value is enumValue; or null if no such constant is found.
+netEnumName <- function(enumType, enumValue){
+  netTypeObj <- rClr::clrGetType("OSPSuite.Core.Domain.Data.AuxiliaryType")
```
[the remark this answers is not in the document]
PavelBal: Damn, that should be the argument `enumType` and not the string!

Issue #431, opened 2020-09-25
RStudio crashing after saving workspace - OSPSuite-R


Chrizsly:
Dear all,

When working with the R package OSPsuite and using functions e.g. loadSimulation or runSimulation, and the workspace in R is saved, RStudio keeps on crashing. Running the same piece of code in the R Gui works just fine. 

Proposed workaround: don't save the current workspace in R, but refer to the documentation (https://docs.open-systems-pharmacology.org/working-with-r/r-introduction#running-a-simulation) on how to store and re-load simulation results. This is especially helpful for time-intense population simulations. 

Thanks and best,

msevestre:
@Chrizsly THis is correct. Our data model is using .NET object under the hood that cannot be serialized by R Studio
This should definitely be in the documentation! Thanks for pointing this out

@Chrizsly Do you feel like creating a PR for the doc?

PR #572, opened 2021-08-06
Update README.md

PavelBal:
Fixes https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/431

Add information about RStudio crashing when trying to load the workspace containing .NET objects.


Yuri05:
@PavelBal 
I would also split the required packages into 2 sublists: packages available on Cran and not available on Cran. Then the sentence below ("R6 on Cran") can be removed.
grafik

PavelBal:
@Yuri05 Updated, please check.

Yuri05:
nice!

PR #574, opened 2021-08-08
Fixes #561 lloq data set

### LINKED ISSUE #561: DataSet: LLOQ
- getLLOQ
- setLLOQ

msevestre:


R/data-set.R
```diff
@@ -129,6 +129,18 @@ DataSet <- R6::R6Class(
 
       rClr::clrSet(dataInfo, "MolWeight", toBaseUnit(quantityOrDimension = ospDimensions$`Molecular weight`, values = value, unit = ospUnits$`Molecular weight`$`g/mol`))
     },
+    #' @field lloq LLOQ value in yUnit associated with the yValues
+    lloq = function(value){
```
msevestre, line 133: @PavelBal very similar to what was implemented for Molweight. Could be refactored if we have a 3rd parameter like this but for now, I am ok with duplicating code

R/data-set.R
```diff
@@ -129,6 +129,18 @@ DataSet <- R6::R6Class(
...
+       dataInfo <- rClr::clrGet(private$.yColumn$ref, "DataInfo")
+      if (missing(value)) {
+        lloq <- rClr::clrGet(dataInfo, "LLOQAsDouble")
+        if (is.null(lloq)) {
+          return(NULL)
+        }
+      return(toUnit(quantityOrDimension = private$.yColumn$dimension, values = lloq, targetUnit = private$.yColumn$displayUnit))
+      }
+      rClr::clrSet(dataInfo, "LLOQAsDouble", toBaseUnit(quantityOrDimension =private$.yColumn$dimension, values = value, unit =  private$.yColumn$displayUnit))
```
PavelBal, line 142: Add `validateIsNumeric(value)` pls.

R/data-set.R
PavelBal: Looks good. I implemented it as a property of DataColumn class (could not finalize as could not set due to float vs double...)

```
    #' @field LLOQ Lower Limit Of Quantification.
    #' In no LLOQ is defined, the value is \code{NULL}
    LLOQ = function(value){
      browser()
      dataInfo <- rClr::clrGet(self$ref, "DataInfo")
      if (missing(value)) {
        return(rClr::clrGet(dataInfo, "LLOQ"))
      }

      validateIsNumeric(value)
      rClr::clrSet(dataInfo, "LLOQ", value)
      },
```

and then in DataSet 

```
    #' @field LLOQ Lower Limit Of Quantification for \code{yValues}.
    #' In no LLOQ is defined, the value is \code{NULL}
    LLOQ = function(value){
      if (missing(value)) {
        # Value in base unit
        lloq <- private$.yColumn$LLOQ
        if (is.null(lloq)) {
          return(NULL)
        }
        return(toUnit(quantityOrDimension = private$.yColumn$dimension, values = lloq, targetUnit = private$.yColumn$displayUnit))
      }

      private$.yColumn$LLOQ <- toBaseUnit(quantityOrDimension = private$.yColumn$dimension, values = value, unit = private$.yColumn$displayUnit)
      },
```

but I actually don't have any preferences. Please add a validation for numeric (see next comment) and then can merge.

PavelBal (changes requested):

msevestre:
@PavelBal I think what you are proposing makes sense (e..g in DataColumn). I'll do the same for MolWeight otherwise it won't be consistent

msevestre:
@PavelBal Updated. Check it out

PR #579, opened 2021-08-09
Load data set from PKML

### LINKED ISSUE #575: DataSet: load from PKML
Wrapper for `loadDataRepositoryFromPKML` and create DataSet.

`loadDataSetFromPKML()`

PavelBal:
Fixes #575


R/utilities-data-repository.R
```diff
@@ -8,8 +8,7 @@
 #'
 #' obsData <- loadDataRepositoryFromPKML(filePath)
 #' metaData <- obsData$metaData
-#' @export
-loadDataRepositoryFromPKML <- function(filePath) {
+.loadDataRepositoryFromPKML <- function(filePath) {
```
PavelBal, line 11: I propose not to export this function. The class DataRepository is also not exported, so user interaction with it should be as minimal as possible.

R/utilities-pk-analysis.R
```diff
@@ -64,21 +64,22 @@ importPKAnalysesFromCSV <- function(filePath, simulation) {
...
-      Value = readr::col_double(),
-      Unit = readr::col_factor()
-    )
-    pkResultsDataFrame <- readr::read_csv(pkParameterResultsFilePath, locale = readr::locale(encoding = "UTF-8"), comment = "#", col_types = colTypes)
-    return(pkResultsDataFrame)
-  },
-  finally = {
-    file.remove(pkParameterResultsFilePath)
-  }
+  dataFrame <- tryCatch(
```
PavelBal, line 67: Just ran styler here.

R/utilities-data-repository.R
msevestre: Why is the DataRepository class not exported ? Aren't they all exported by default?

msevestre (changes requested):
I am not sure why you want to hide the DataRepository. One can create a DataSet from a Data repository so I think it should be available. Can you explain?

R/utilities-data-repository.R
PavelBal: Nope, you have to add @export to class definition.

PavelBal:
Don't you think there is too much "choice" for the user if we expose both DataSet and DataRepository? We wrote the DataSet around DataRepository, and I actually do not see a use case using DataRepository directly. If creating from scratch, it is much easier to use DataSet. If loading from pkml, `loadDataSetFromPKML` does the job.

msevestre:
@PavelBal Ok fair enough. But then we need to export method to create dataSet from a file etc.. .I don't see it in the code that you just created. Maybe you forgot to add a file?

R/utilities-data-set.R
```diff
@@ -8,6 +8,22 @@
...
+#' The pkml files are typically exported from PK-Sim or MoBi.
+#'
+#' @param filePath Full path of pkml file containing the observed data to load
+#'
+#' @examples
+#' filePath <- system.file("extdata", "obs_data.pkml", package = "ospsuite")
+#'
+#' obsData <- loadDataSetFromPKML(filePath)
+#' @export
+loadDataSetFromPKML <- function(filePath) {
```
msevestre, line 21: oh it's here :) lol

msevestre (commented):
ok passt

PR #580, opened 2021-08-09
Added method for exporting DataSet as PKML

### LINKED ISSUE #578: DataSet: save as pkml
Save the DataRepositry as PKML. Would implement it as a method of `DataSet`

PavelBal:
Fixes #578


R/data-set.R
```diff
@@ -197,6 +197,18 @@ DataSet <- R6::R6Class(
       }
     },
 
+    #' @description
+    #' Save the \code{DataSet} to a pkml file that can be loaded by MoBi
+    #' @param filePath Path to the location (incl. file name)
+    saveToPKML = function(filePath){
```
msevestre, line 203: In order to stay consistent with what we have done for other objects, shouldn't we create a utility-data-set or sthg and define this method there instead? Also this is were you could create some help methods around dataSet such as createDataSet(x=, y= etc..)

PavelBal: Ok, then "saveDataSetToPKML" or "export..."? We are using "export" for simulation results etc, but "save" sounds somehow more intuitive for me..

msevestre: saveDataSetToPKML
We can also create aliases for other method (so that we are consistent and always have save..xx)

PavelBal: @msevestre We would have to expose the `DataRepository` of the `DataSet` in order to save it via a utility function...

msevestre: Hontestly I do not have an issue with exposing the underlying DataREpository as public readonly {get;}. We can create the object tusign a DataRepo as parameter so why not?

PavelBal: ok, I will expose it (readonly).

msevestre: any update on this? Id' like to merge asap

PavelBal:
@msevestre Updated.

PR #589, opened 2021-08-24
Data importer

PavelBal:
This PR implements the ImporterConfiguration and requires the Core version from this PR: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/1250

Importing data sets in R requires an excel file and a configuration. There are three ways to get a configuration in R:

1) Load an existing configuration (e.g. the one created in PK-Sim or MoBi)
2) Create a configuration from scratch
3) Create a configuration for a file/sheet (basically what PK-Sim tries to do when starting an import process). This is not implemented yet and will be done in a separate PR.


PavelBal:
The failing tests should work again when the Core PR is merged.

PavelBal:
Each configuration has at least a time and a measurement columns, so these are created when creating an empty configuration. An error column can be added or removed manually.

R/utilities-data-import-configuration.R
```diff
@@ -0,0 +1,2 @@
+createConfigurationForFile <- function(filePath) {
```
PavelBal, line 1: Will be implemented later.

R/utilities-pk-analysis.R
```diff
@@ -64,21 +64,22 @@ importPKAnalysesFromCSV <- function(filePath, simulation) {
...
-      Value = readr::col_double(),
-      Unit = readr::col_factor()
-    )
-    pkResultsDataFrame <- readr::read_csv(pkParameterResultsFilePath, locale = readr::locale(encoding = "UTF-8"), comment = "#", col_types = colTypes)
-    return(pkResultsDataFrame)
-  },
-  finally = {
-    file.remove(pkParameterResultsFilePath)
-  }
+  dataFrame <- tryCatch(
```
PavelBal, line 67: Run styler

tests/data/dataImporterConfiguration.xml
```diff
@@ -0,0 +1,30 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
PavelBal, line 1: Added a couple of configurations created in PK-Sim for testing.

R/data-importer-configuration.R
```diff
@@ -0,0 +1,356 @@
...
+      if (missing(value)) {
+        return(rClr::clrGet(column, "ColumnName"))
+      }
+      validateIsString(value)
+      rClr::clrSet(column, "ColumnName", enc2utf8(value))
+    },
+
+    #' @field timeUnit If \code{timeUnitFromColumn} is \code{FALSE}, unit of the values in time column
+    #' If \code{timeUnitFromColumn} is \code{TRUE}, name of the column with units of the values in time column
+    timeUnit = function(value) {
```
PavelBal, line 23: Time unit (the same is valid for measurement unit) - it is either a fixed unit that the user can change, or the name of the column that defines the units.

PR #598, opened 2021-08-30
update vignette for efficient calculations

PavelBal:
@msevestre Nice description, indeed! I corrected some typos and added some explanation to the vignette, let me know what do you think.

### DIFF
--- a/vignettes/efficient-calculations.Rmd
+++ b/vignettes/efficient-calculations.Rmd
@@ -18,11 +18,11 @@ knitr::opts_chunk$set(
 There are a few ways to run simulations using this package. 
 The easiest way is, as expected, not necessarily the most efficient one. So depending on the use case, a different approach than what is outlined in [Run Simulations](run-simulation.html) might be more suited.
 
-This vignettes aims at describing some scenarios and the suggested methods to efficiently run simulations.
+This vignette aims at describing some scenarios and the suggested methods to efficiently run simulations.
 
 ## Life cycle of a simulation run
 
-In general, a user loads a simulation, update some parameter values, calculates the simulation and retrieves some results. 
+In general, a user loads a simulation, updates some parameter values, calculates the simulation and retrieves the results. 
 Here are some steps inherent to each simulation:
 
 1. Loading a simulation
@@ -37,13 +37,13 @@ Depending on the simulation of interest, some steps might take much longer than 
 ## Running a single simulation once
 
 This is by far the easiest use cases. A simulation is loaded from a pkml file and the outputs are calculated. In this use case, step 4 and 5 are automatically lumped together in one single call
-See [Run Simulations](run-simulation.html) for more details. 
+(See [Run Simulations](run-simulation.html) for more details).
 
 
 ## Running a single simulation multiple times and varying some parameters/initial values between each run
 
 This is already a use case where the optimal and most efficient way depends drastically on the simulation at hand.
-Let's assume that we want to calculate the outputs of a simulation for different doses
+Let's assume that we want to calculate the outputs of a simulation for different doses.
 
 A simple implementation could follow the following idea, where a simulation is loaded once, the dose value is updated and the simulation is calculated for each dose of interest. 
 ```{r SequentialRun, eval=TRUE}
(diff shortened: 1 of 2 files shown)


vignettes/efficient-calculations.Rmd
```diff
@@ -100,64 +100,81 @@ results <- runSimulations(simulations = list(sim100, sim200, sim300, sim400, sim
 # Results in now a list of SimulationResults
 ```
 
-This implementation is also fairly straight forward. We load the simulation multiple times from the same file (note the usage of `loadFromCache=FALSE` to ensure that we always get a new instance of a simulation and not the same instance) and we set the dose. Then, we call the `runSimulations` function with the list of simulations. The engine will run these simulations in parallel, which means that the overall execution time should be slightly more than 1h, as each run would be executed **at the same time**.
+This implementation is also fairly straight forward. We load the simulation multiple times from the same file (note the usage of `loadFromCache = FALSE` to ensure that we always get a new instance of a simulation and not the same instance) and we set the dose. Then, we call the `runSimulations` function with the list of simulations. The engine will run these simulations in parallel, which means that the overall execution time should be slightly more than 1h, as each run would be executed **at the same time**.
 
 
-## Running one or more simulations multiple times in a loop
```
PavelBal, line 106: I have the feeling that this was not the right caption? I did not see a section where you describe a loop. I propose a new caption.

vignettes/efficient-calculations.Rmd
```diff
@@ -100,64 +100,81 @@ results <- runSimulations(simulations = list(sim100, sim200, sim300, sim400, sim
...
+and define the parameter values that can be changed between runs. Note that, as opposed to 
+the `Simulation` object, we do not have access to any parameter/species values nor can we change
+the outputs or simulation time.
+```{r BatchRunInit, eval=TRUE}
 library(ospsuite)
 simFilePath <- system.file("extdata", "Aciclovir.pkml", package = "ospsuite")
 
-# We create two different instances of the same simulation to illustrate the usage.
+# We load the simulation for which the batches will be created
 sim1 <- loadSimulation(simFilePath, loadFromCache = FALSE)
```
PavelBal, line 119: Actually, no two sims needed here.

vignettes/efficient-calculations.Rmd
```diff
@@ -100,64 +100,81 @@ results <- runSimulations(simulations = list(sim100, sim200, sim300, sim400, sim
...
+# for the second batch, we will vary Molecular Weight
 simBatch2 <- createSimulationBatch(
-  simulation = sim2, 
+  simulation = sim1, 
   parametersOrPaths = c("Aciclovir|Molecular weight")
   )
 
 simulationBatches <- list(simBatch1, simBatch2)
+```
+
```
PavelBal, line 136: I split the code into smaller chunks with some info in between.

vignettes/efficient-calculations.Rmd
```diff
@@ -100,64 +100,81 @@ results <- runSimulations(simulations = list(sim100, sim200, sim300, sim400, sim
...
 simBatch2$addRunValues(parameterValues = 300)
 simBatch2$addRunValues(parameterValues = 400)
+```
 
+So far, we created 2 simulation batches, one with 3 parameter sets and the 
+other one with 4. That means that 3 runs will be enqueued for simBatch1 and 
+4 will be enqueued for simBatch2. Each run gets a unique id that can later be used
+to correctly assign simulation results to the simulated set of parameters.
 
-# So far, we created 2 simulation batches, one with 3 parameter sets and the
```
PavelBal, line 146: I find it a bit too technical, with memory allocation etc..

vignettes/run-simulation.Rmd
```diff
@@ -76,38 +76,40 @@ csvResultsPath <- system.file("extdata", "SimResults.csv", package = "ospsuite")
...
-library(ospsuite)
-
-# Load the simulation
-simFilePath <- system.file("extdata", "Aciclovir.pkml", package = "ospsuite")
-sim <- loadSimulation(simFilePath)
-
-simulationResults <- runSimulations(simulations = sim)
-print(simulationResults)
+The results in this case will be a named list of `SimulationResults`-object , i.e. one for each simulation.
+The names of the entries of the list are the ids of the simulation to which the results
```
PavelBal, line 103: Added some info about ids.

vignettes/efficient-calculations.Rmd
msevestre: I would only use batch if it  was in a loop such as an algorithm (PI, sensitivity etc..) .For a finite number of run, I am not sure that this is really required.....

vignettes/efficient-calculations.Rmd
msevestre: Yes this was just to say that you can have multiple simulations and not only one simulation

vignettes/efficient-calculations.Rmd
```diff
@@ -167,11 +184,13 @@ simBatch2$addRunValues(parameterValues = 200)
...
 
-# Once the simulation batches instances are not need anymore, they can be removed 
+# Once the simulation batches instances are not needed anymore, they can be removed 
 # from the environment and the allocated memory cleared. 
 # This will be done automatically when the R session is terminated.
 rm(simBatch1)
 rm(simBatch2)
 ```
 
-Usage of `SimulationBatch` is recommended for advanced scenarios where simulations expected to be run hundreds of times and where each second that can be spared will impact the performance significantly.
```
msevestre, line 177: I would not do this as it is MD and therefor will be formatted automatically. Why do you think we need to add CR manually?

msevestre:
Fine with me

Issue #576, opened 2021-08-09
DataSet: load from excel


PavelBal:
Using importer + configurations.

PR #599, opened 2021-08-30
Data set from excel

### LINKED ISSUE #597: DataImporterConfiguration: get/set NamingConvention

PavelBal:
Fixes #597, #576

This PR allows creation of DataSets from an excel with a valid configuration.


R/utilities-data-set.R
```diff
@@ -95,3 +95,45 @@ dataSetToDataFrame <- function(dataSets) {
...
+#'
+#' @examples
+#' \dontrun{
+#' configurationPath <- "../dataImporterConfiguration_noSheets.xml"
+#' xlsFilePath <- "../CompiledDataSet_oneSheet.xlsx"
+#' importerConfiguration <- DataImporterConfiguration$new(configurationPath)
+#'
+#' dataSets <- loadDataSetsFromXls(xlsFilePath = xlsFilePath, importerConfiguration = importerConfiguration, importAllSheets = TRUE)
+#' }
+loadDataSetsFromXls <- function(xlsFilePath, importerConfiguration, importAllSheets = FALSE) {
```
msevestre, line 122: I would rename it to loadDataSetsFromExcel or at least loadDataSetsFromXLS (CSV, PKML etc..)

R/utilities-data-set.R
```diff
@@ -95,3 +95,45 @@ dataSetToDataFrame <- function(dataSets) {
...
+  validateIsString(xlsFilePath)
+  validateIsOfType(importerConfiguration, DataImporterConfiguration)
+  validateIsLogical(importAllSheets)
+
+  dataImporterTask <- getNetTask("DataImporterTask")
+  rClr::clrSet(dataImporterTask, "IgnoreSheetNamesAtImport", importAllSheets)
+  dataRepositories <- rClr::clrCall(dataImporterTask, "ImportExcelFromConfiguration", importerConfiguration$ref, xlsFilePath)
+  dataSets <- lapply(dataRepositories, \(x){
```
msevestre, line 130: \(x). What does this do?

msevestre (commented):
I think the binaries (ospsuite.core etc.. )do not match the version in package.json
Where are those coming from?

R/utilities-data-set.R
PavelBal: Will rename to "E.xcel"

R/utilities-data-set.R
PavelBal: Oooops I got used to it.
Since R4, anonymous functions (like inside a lapply call) can be written as \(x) instead of function(x).
So instead of 

`lapply(valuesArray, function(x){x + 1})`

one can write

`lapply(valuesArray, \(x){x + 1})`

PavelBal:
No I just took the binaries that I compiled locally from the develop-branch so it would work. They should be updated again and the package,json adjusted again.

packages.config
```diff
@@ -14,4 +14,6 @@
   <package id="Microsoft.Extensions.Logging.Abstractions" version="3.1.0"/>
   <package id="Autofac" version="4.9.4"/>
   <package id="MathNet.Numerics" version="4.12.0"/>
+  <package id="NPOI" version="2.5.1"/>
```
msevestre, line 17: I have added those two entries here

msevestre (commented):
\ does not work for me. Ihave replaced it with function. I think this is also not required to gain  7 chars here.

Issue #604, opened 2021-09-08
dataSetToDataFrame: wrong results for multiple data sets


PavelBal:
```
configurationPath <- getTestDataFilePath("dataImporterConfiguration_noSheets.xml")
xlsFilePath <- getTestDataFilePath("CompiledDataSet_oneSheet.xlsx")
importerConfiguration <- DataImporterConfiguration$new(configurationPath)
dataSets <- loadDataSetsFromExcel(xlsFilePath = xlsFilePath, importerConfiguration = importerConfiguration, importAllSheets = TRUE)

dataSetsFrame <- dataSetToDataFrame(dataSets)
```

msevestre:
@PavelBal What's the isssue? What do you get and what do you expect?

PavelBal:
Definitely not what is expected ;) I am on it this moment.

msevestre:
fixed with #606

PR #606, opened 2021-09-08
Fix dataSetToDataFrame

### LINKED ISSUE #603: DataSet: add "name" to print
Second line of the print function should print the name of the data set.

PavelBal:
Fixes #603; #604

### DIFF
--- a/R/utilities-data-set.R
+++ b/R/utilities-data-set.R
@@ -1,4 +1,5 @@
-.makeDataFrameColumn <- function(dataSet, property, metaDataName = NULL) {
+.makeDataFrameColumn <- function(dataSets, property, metaDataName = NULL) {
+  columForDataSet <- function(dataSet){
   # check length of entry for a certain property of this data set, i.e. if it exists
   if (is.null(metaDataName)) {
     len <- length(dataSet[[property]])
@@ -17,6 +18,11 @@
   } else {
     dataSet[[property]]
   }
+  }
+
+  unlist(lapply(dataSets, function(x) {
+    columForDataSet(x)
+  }), use.names = FALSE)
 }
 
 #' Loads data (typically observed data) from a PKML file and creates a \code{DataSet} from it.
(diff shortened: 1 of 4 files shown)


R/utilities-data-set.R
```diff
@@ -55,38 +55,64 @@ saveDataSetToPKML <- function(dataSet, filePath) {
...
-  yUnit <- unlist(mapply(.makeDataFrameColumn, dataSets, "yUnit"))
-  yErrorUnit <- as.character(unlist(mapply(.makeDataFrameColumn, dataSets, "yErrorUnit")))
-  xDimension <- unlist(mapply(.makeDataFrameColumn, dataSets, "xDimension"))
-  yDimension <- unlist(mapply(.makeDataFrameColumn, dataSets, "yDimension"))
-  yErrorType <- as.character(unlist(mapply(.makeDataFrameColumn, dataSets, "yErrorType")))
-  yMolWeight <- unlist(mapply(.makeDataFrameColumn, dataSets, "molWeight"))
-  xValue <- unlist(mapply(.makeDataFrameColumn, dataSets, "xValues"))
-  yValue <- unlist(mapply(.makeDataFrameColumn, dataSets, "yValues"))
-  yErrorValues <- unlist(mapply(.makeDataFrameColumn, dataSets, "yErrorValues"))
+  name <- unlist(lapply(dataSets, function(x) {
```
PavelBal, line 70: mapply and sapply and so produce weird results sometimes (in case of data sets loaded from excel the output was a matrix, in the case of manually created data sets it was a vector...), `lapply` is the only function that _always_ will return a list, so lets use it where possible.

Adding `use.naames = FALSE` to `unlist` speeds up the whole thing and should be used when we do not need a named list.

PavelBal: @hannaei FYI

R/utilities-data-set.R
```diff
@@ -55,38 +55,64 @@ saveDataSetToPKML <- function(dataSet, filePath) {
...
+    .makeDataFrameColumn(x, "yErrorValues")
+  }), use.names = FALSE)
+  lloq <- unlist(lapply(dataSets, function(x) {
+    .makeDataFrameColumn(x, "LLOQ")
+  }), use.names = FALSE)
 
   df <- data.frame(
-    name, xValue, yValue, yErrorValues, xDimension, xUnit, yDimension,
-    yUnit, yErrorType, yErrorUnit, yMolWeight
+    name, xValues, yValues, yErrorValues, xDimension, xUnit, yDimension,
```
PavelBal, line 108: Renamed column names (using plural) and added a LLOQ column.

msevestre: if feels like you could write a method
```
makeColums <- function(dataSets, columnName) {
  unlist(lapply(dataSet, function(x){
.makeDatqaFrameColumn(x, columnsName)
}
}

```
or sthg like this

msevestre (commented):
see suggestion to avoid code duplicatin

PavelBal:
@msevestre done.
Also the build is not failing any more.

PR #607, opened 2021-09-08
587 get seed

msevestre:


DESCRIPTION
```diff
@@ -90,8 +90,6 @@ Collate:
     'table-formula.R'
     'user-defined-pk-parameter.R'
     'utilities-container.R'
-    'utilities-data-import-configuration.R'
-    'utilities-data-importer-configuration.R'
```
msevestre, line 94: @PavelBal I am not sure what these files were about? They were empty or containing an empty function

R/create-population.R
```diff
@@ -3,35 +3,28 @@
...
   populationFactory <- rClr::clrCallStatic("PKSim.R.Api", "GetPopulationFactory")
-  netPopulation <- rClr::clrCall(populationFactory, "CreatePopulation", populationCharacteristics$ref)
+  results <- rClr::clrCall(populationFactory, "CreatePopulation", populationCharacteristics$ref)
+  netPopulation <- getPropertyValue(results, "IndividualValuesCache")
+  seed <- getPropertyValue(results, "Seed")
   population <- Population$new(netPopulation)
 
   individualCharacteristics <- NULL
 
-  # NOTE THIS IS A WORKAROUND UNTIL THE CODE IN PKSIM IS UPDATED
```
msevestre, line 20: Also fixed this in PKSim

R/utilities-data-set.R
```diff
@@ -1,23 +1,23 @@
...
-  if (is.null(metaDataName)) {
-    len <- length(dataSet[[property]])
-  } else {
-    len <- length(dataSet[[property]][[metaDataName]])
-  }
-
-  if (len == 0) {
-    rep(NA, length(dataSet$xValues))
-  } else if (len == 1) {
+  columForDataSet <- function(dataSet) {
```
msevestre, line 2: just running styler here

tests/testthat/test-data-importer-configuration.R
```diff
@@ -1,6 +1,8 @@
 context("New DataImporterConfiguration")
 
 test_that("it can create a new data importer configuration", {
+  skip_on_os("linux") # TODO enable again as soon as npoi works under linux
```
msevestre, line 4: this was missing everywhere too

tests/testthat/test-data-importer-configuration.R
```diff
@@ -1,6 +1,8 @@
 context("New DataImporterConfiguration")
 
 test_that("it can create a new data importer configuration", {
+  skip_on_os("linux") # TODO enable again as soon as npoi works under linux
+
   importerConfiguration <- DataImporterConfiguration$new()
```
msevestre, line 6: I honestly do not like the fact that so much happens on new. A constructor should be very fast and **NEVER** crash. 
I would like to see a method on that object  for importing or creating. Right now this is all implicit and it is not what I would expect

PavelBal (approved):

PR #617, opened 2021-09-14
Fixes #553 integer unit

### LINKED ISSUE #553: Integer variables not handled by some functions
Sometimes observed data are assumed as integer and not numeric variables by R, which crashes the function `toUnit` with the error message below. It could be helpful to enforce integer variables into numeric within functions such as `toUnit` (using `as.numeric` ?)

```
ospsuite::toUnit("Time", as.integer(c(1,2,3)), "h")

Error in rClr::clrCall(dimensionTask, "ConvertToUnit", dimension, targetUnit,  : 
  Type:    System.MissingMethodException
Message: Could not find a suitable instance method ConvertToUnit on type OSPSuite.R.Services.DimensionTask for method parameters String, String, Int32[]
Method:  Void ThrowMissingMethod(System.Type, System.String, System.String, System.Type[])
Stack trace:
   at Rclr.ReflectionHelper.ThrowMissingMethod(Type classType, String methodName, String modifier, Type[] types)
   at Rclr.ClrFacade.InternalCallInstanceMethod(Object obj, String methodName, Boolean tryUseConverter, Object[] arguments)
```

msevestre:


msevestre:
Easy fix I believe. I trust that it does not break anything since all tests are passing....

PR #622, opened 2021-09-21
Fixes #621 minimum r version

### LINKED ISSUE #621: Add minimum R version
see https://github.com/Open-Systems-Pharmacology/docs/issues/156

msevestre:


DESCRIPTION
```diff
@@ -16,7 +16,7 @@ URL: https://github.com/open-systems-pharmacology/ospsuite-r
 BugReports: 
     https://github.com/open-systems-pharmacology/ospsuite-r/issues
 Depends: 
-    R (>= 3.5),
+    R (>= 3.6),
```
msevestre, line 19: This was wrong here also @IndrajeetPatil

IndrajeetPatil (approved):

DESCRIPTION
Yuri05, line 19: Why? It works fine under 3.5 - what is the reason to require 3.6?

msevestre: the readme says it works under 3.6 :) I assumed this was accurate?!?

msevestre: seem to remember some issue with rclr and 3.5. But maybe I am wrong

msevestre: I don't think this is going to work here as we need different version of rclr based on the version of R

msevestre: see https://github.com/Open-Systems-Pharmacology/OSPSuite-R#installation

msevestre: we install the 4.x version specifically.

msevestre: here https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/appveyor.yml#L40

IndrajeetPatil: Oh, I see! Then we should wait and ascertain that the CI indeed fails :)

Okay, outside the CI context, I am not really sure how to check if the package works with `3.5`.

msevestre: ah ah yes!

Looks like at least it will nicely mess up the version number lol
image

msevestre:
3.5 seems to be working indeed. So I will update to 3.5 in the readme and docs

msevestre:
Well tests are failling with 3.5
Going to bump to 3.6 if they pass with this version!

msevestre:
```
Running the tests in 'tests/testthat.R' failed.
Last 13 lines of output:
  Component "Organ": Modes: character, numeric
  Component "Organ": Attributes: < target is NULL, current is list >
  Component "Organ": target is character, current is factor
  Component "Compartment": Modes: character, numeric
  Component "Compartment": Attributes: < target is NULL, current is list >
  Component "Compartment": target is character, current is factor
  
  Warning - Could not find quantity with path 'Organism|Liver|NOPE|Volume'.
  == testthat results  ===========================================================
  [ OK: 849 | SKIPPED: 0 | WARNINGS: 3 | FAILED: 2 ]
  1. Failure: It can convert a data set with metaData (@test-utilities-data-set.R#59) 
  2. Failure: It can convert a list of data sets (@test-utilities-data-set.R#79) 
  
  Error: testthat unit tests failed
  Execution halted
```

msevestre:
ok works with 3.6 and higher. I don't see why we should support an older version anyways. So min is now 3.6. :)

PR #623, opened 2021-09-21
test the package on devel, release, and oldrel

IndrajeetPatil:
Instead of just testing on the old release of `4.0.5`, we should also be checking on `4.1.1` (release) and `4.2` (devel).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,11 @@ environment:
   NOT_CRAN: true
   KEEP_VIGNETTES: true
   R_ARCH: x64
-  R_VERSION: "4.0.5"
+    matrix:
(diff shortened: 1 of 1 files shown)


IndrajeetPatil:
So if I understand this correctly, `{rClr}` package will work with the major R version for which it was built? 
For example, if it was built for `4`, it will work with `4.0`, `4.1`, `4.2`, etc., but not `3.0` or `5.0`?

PavelBal:
@IndrajeetPatil rClr built for R4 will work with R4 but not R3, rClr built with R3 will work with R3 but not R4 - this is all we know :D No idea why, and cannot predict how it will behave with R5.

PR #624, opened 2021-09-22
add a static website for the package

IndrajeetPatil:
Adds a `{pkgdown}` website.

After the PR is merged, whoever can change the Settings to the following:

image

And then add the URL here:

image

If you wish to have a look at what the website looks like _locally_, just run the following the package **root directory**:

```r
pkgdown::build_site()
```


msevestre:
@IndrajeetPatil How do we update the content? I guess this needs to be done by hand now right?

msevestre:
http://www.open-systems-pharmacology.org/OSPSuite-R/

why is it using http://www.open-systems-pharmacology.org?

msevestre:
oh it's becuase our website is using github page and is http://www.open-systems-pharmacology.org

msevestre:
how do we navigate to subpage ? e.g. createPopulation etc?

IndrajeetPatil:
All function docs are under reference: http://www.open-systems-pharmacology.org/OSPSuite-R/reference/index.html

For example, if you want to see `createPopulation`, that will be here:
http://www.open-systems-pharmacology.org/OSPSuite-R/reference/createPopulation.html

msevestre:
done

PR #630, opened 2021-09-30
Set quantity values with units

### LINKED ISSUE #629: Allow setting parameter/molecule/quantity values with units
Currenty: `setQuantityValues <- function(quantities, values)`

Proposal: `setQuantityValues <- function(quantities, values, units = NULL)`

and with `units = NULL`, base units are used.

PavelBal:
Fixes #629


R/utilities-molecule.R
```diff
@@ -81,11 +82,38 @@ getMolecule <- function(path, container, stopIfNotFound = TRUE) {
...
+#'
+#' @examples
+#'
+#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' setMoleculeValuesByPath("Organism|VenousBlood|Plasma|CYP3A4", 1, sim)
+#'
+#' setMoleculeValuesByPath(c("Organism|VenousBlood|Plasma|CYP3A4", "Organism|ArterialBlood|Plasma|CYP3A4"), c(2, 3), sim, units = c("µmol", "mmol"))
+#' @export
+setMoleculeValuesByPath <- function(moleculePaths, values, simulation, stopIfNotFound = TRUE, units) {
```
PavelBal, line 109: We had this for parameters but not for molecules.

R/utilities-units.R
```diff
@@ -280,9 +280,10 @@ getUnitsEnum <- function() {
     x <- getUnitsForDimension(dimension = dimension)
     return(enum(replace(x, x == "", "Unitless")))
   })
-  names(units) <- sapply(dimensions,function(str){
-    str <- gsub(pattern = "[(]",replacement = "[",x = str)
-    str <- gsub(pattern = "[)]",replacement = "]",x = str)})
+  names(units) <- sapply(dimensions, function(str) {
```
PavelBal, line 283: Just styler

R/utilities-quantity.R
```diff
@@ -113,19 +125,31 @@ setQuantityValues <- function(quantities, values) {
 #'
 #' setParameterValuesByPath(list("Organism|Liver|Volume", "Organism|Liver|A"), c(2, 3), sim)
 #' @export
-setQuantityValuesByPath <- function(quantityPaths, values, simulation, stopIfNotFound = TRUE) {
+setQuantityValuesByPath <- function(quantityPaths, values, simulation, stopIfNotFound = TRUE, units = NULL) {
```
msevestre, line 128: looking good. stopInfNotFound is in general OUR LAST parameter. Here we are changing the signatures `logic` .
shouldn't we put `units` before `stopIfNotFound`  If people have used this flag, one can argue that they would have named it explicitly to set it to FALSE

msevestre:
@PavelBal I have fixed the code to PR is working now. Let me know your thoughts on the order of parameters.

R/utilities-quantity.R
PavelBal: I couldn't decide what is better - keep the logic by putting `stopIfNotFound` at the end, or maintain compatibility with code that uses the function without specifying argument names. But I think the method has not been used extensively so far, so let's keep the logic. I will update.

PavelBal:
@msevestre `stopIfNotFound` is now always the last parameter. Ready to merge unless you have other comments.

PR #633, opened 2021-09-30
626 errors in validate

msevestre:


R/error-checks.R
```diff
@@ -41,14 +41,20 @@ isOfType <- function(object, type, nullAllowed = FALSE) {
 #' @param nullAllowed Boolean flag if `NULL` is accepted for the `object`. If `TRUE`,
 #' `NULL` is always valid, otherwise the error is thrown. Default is `FALSE`
 validateIsOfType <- function(object, type, nullAllowed = FALSE) {
+  if (type == "integer") {
```
msevestre, line 44: this is to ensure that `2` and integer are a match...it was not the case

PavelBal (approved):
Looking good

PR #641, opened 2021-10-04
Update dlls

msevestre:


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: ospsuite
 Title: R package to manipulate OSPSuite Models
-Version: 10.0.0
+Version: 10.0.70
```
msevestre, line 4: @Yuri05 @IndrajeetPatil I checked what was the last version on appveyor so that we would get the expected version here...Not required once we automate the deploy

msevestre (commented):
@IndrajeetPatil We will always have changes since it uses some local files
image

PR #665, opened 2021-11-17
minor 80-width doc formatting

IndrajeetPatil:
I am making this (non-code) PR just to check if there is something broken in the vignettes or it is just my local version of PK-Sim that is creating problems.


IndrajeetPatil:
Yeah, looks like a local issue.

At any rate, this PR does fix a minor issue I noticed for SimulationPKAnalyses with printing method.

Yuri05 (approved):

PR #666, opened 2021-11-17
closes #664

### LINKED ISSUE #664: About DEPRECATED tag for `runSimulation()`
This function seems to be used on multiple occasions in the package, so I am confused as to why it has a deprecated tag 🤔 

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/f19aef1b0572bccfd89e6e1da85946fb721e0d41/R/utilities-simulation.R#L85

Or are we planning to remove it soon?

IndrajeetPatil:
Can be merged whenever we are ready to make this breaking change.


PavelBal:
Wow-wow, not that fast :D I would wait at least a couple of months.

IndrajeetPatil:
Haha

Sure thing. We can keep it pending for as long as we want.

Still needs to fix some issues with vignettes, though.

msevestre:
I don't like this.
First iteration shoukld be that we throw a warning when using this methods saying that this method is deprecated and will be removed in v12. Then we remove it

PR #688, opened 2021-12-10
remove rd docs for finalize methods

IndrajeetPatil:
#682 moved `finalize` from public to private methods, but their Rd documentation was still there, which is unnecessary since these are private methods.

This PR removes that documentation.

Additional, formatting changes to make code at a few locations a bit more tidy and docs to 80-character width.

### DIFF
--- a/R/data-set.R
+++ b/R/data-set.R
@@ -16,6 +16,7 @@ DataSet <- R6::R6Class(
   inherit = DotNetWrapper,
   cloneable = FALSE,
   active = list(
+
     #' @field name The name of the DataSet
     name = function(value) {
       if (missing(value)) {
@@ -25,6 +26,7 @@ DataSet <- R6::R6Class(
     },
 
     #' @field dataRepository The underlying DataRepository object
+
     dataRepository = function(value) {
       if (missing(value)) {
         return(private$.dataRepository)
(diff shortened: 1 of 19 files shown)


IndrajeetPatil:
Merging it since there is no code-related changes. Asked for a review only because CI wasn't working when the PR was initially made.

R/data-set.R
```diff
@@ -125,33 +146,57 @@ DataSet <- R6::R6Class(
...
     LLOQ = function(value) {
       if (missing(value)) {
         # Value in base unit
         lloq <- private$.yColumn$LLOQ
         if (is.null(lloq)) {
           return(NULL)
         }
-        return(toUnit(quantityOrDimension = private$.yColumn$dimension, values = lloq, targetUnit = private$.yColumn$displayUnit))
+
+        return(
```
msevestre, line 182: I think return only is weird and not consistent with the rest. See below

R/data-set.R
```diff
@@ -125,33 +146,57 @@ DataSet <- R6::R6Class(
...
     #' @field molWeight Molecular weight of the yValues in g/mol
+
     molWeight = function(value) {
       if (missing(value)) {
         molWeight <- private$.yColumn$molWeight
         if (is.null(molWeight)) {
           return(NULL)
         }
-        return(toUnit(quantityOrDimension = ospDimensions$`Molecular weight`, values = molWeight, targetUnit = ospUnits$`Molecular weight`$`g/mol`))
+        return(toUnit(
```
msevestre, line 158: This is better like this

R/data-set.R
IndrajeetPatil: I have reverted this change.

R/data-set.R
IndrajeetPatil: I have reverted this change.

msevestre: Well one of this was good lol.
;-)

IndrajeetPatil: LOL, sorry misunderstood.

I think we need to have a coding standard for this, instead of me changing the formatting haphazardly. 

Specifically, at what point we prefer to have one argument per line: when it is over 80 characters?, when there are more than 3 arguments? etc.

msevestre: I would love the styler to do it for us. Personal preferences is irrelevant. As long as a machine formats the  code. Can we have the styler do if for us?

msevestre: Do it*

IndrajeetPatil: > I would love the styler to do it for us

There has been [this](https://github.com/r-lib/styler/issues/247) issue pending since 4 years and yet to be resolved. If it is ever resolved, it might do this for us.

Issue #674, opened 2021-12-06
API proposal for bridging `{ospsuite}` and `{tlf}` packages


IndrajeetPatil:
Based on discussion with @msevestre, @PavelBal, @pchelle, @Yuri05   

Adopt a functional approach to create plots using `{tlf}` with a few arguments and sensible internal defaults. 

The primary argument would be an R6 object encapsulating both simulated and observed data from `{ospsuite}`.

![plotProposal](https://user-images.githubusercontent.com/11330453/144883501-d519efdf-b4e4-4b49-a914-b6951b70e59d.jpg)

As to which arguments these plots should have remains to be decided (see the comment below with possible arguments).

@PavelBal Sorry creating issue now since need to go to sleep soon 😅

IndrajeetPatil:
Possible arguments:

- xUnit
- yUnit

msevestre:
Here my pseudo code from this morning

```


cb <- CombinedData$new(SimulationResults, DataSet)...needs to be defined
cb.Color red


IN OSPSUITE-R

plotTimeProfile(cb, xUnit="min", yUnit="mg/l",   plotConfiguration){

	1. create TLF dataObject (using, cb converting values from baseUnit to xUnit and yUnit)
	2. Create MetaData on the fly (baseGrid, x, concentration y) specify the unit (xUnit, and yUnit)
	3. Create your PlotConfiguration(TLF) using the properties
	4. observedDataMapping and observedDate (from cb) unit conversion also happpening
	5. color <- red

	return tlf::plotTimeProfile(data, metadata....)


plotPredVsObserved(cb,){


}

IndrajeetPatil:
Steps needed:

- [ ] implement `DataCombined` objects (containing simulation results, observed data, and grouping information)
- [ ] implement `plot*()` functions

IndrajeetPatil:
Plots needed:

- [ ] histogram
- [ ] boxplot
- [ ] DDI plot
- [ ] tornado plot
- [ ] PK ratio plot
- [ ] concentration profile plot
- [ ] observed versus predicted plot
- [ ] residuals versus predicted plot

PavelBal:
- `DataCombined` should allow to set scale factors and offsets for data (x and y values)
- `DataCombined` should have a method `toDataFrame` that returns a data frame with all data sets and additional information specified (like dimension, unit, offsets, scale factors, meta data)

Following plot types are prio one:
[...]

PR #690, opened 2021-12-10
Add `DataCombined` object

IndrajeetPatil:
This is Step-1 of https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/674

- [x] add datasets to tests folder for testing
- [x] add method for converting `SimulationResults` object to dataframe
- [x] introduces `DataCombined` object and related methods
- [x] add grouping 
- [x] allow setting scale factors and offsets
- [x] support sequential update
- [x] add tests for `DataCombined`
- [ ] add tests for utility functions
- [ ] add comments to tests
- [x] improve `simulationResultsToDataFrame()` method to extract molecular weights
- [x] modularize and refactor code
- [x] check to see if additional dependencies are really necessary
- [x] only single internal dataframe to contain everything
- [x] add comments to explain the tidy workflow
- [x] make entry in NEWS about all related changes
- [x] format code, spell check
- [x] ~~add examples for the manual~~
- [x] ~~add vignette to demonstrate its usage and all important features~~ 
      (someone else, probably a regular user of the package, should take care of doing this in a later PR)
- [x] address peer review comments
- [ ] move utility functions to utilities package
- [ ] CRAN release for the utilities package
- [ ] remove the utilities from ospsuite


R/data-combined.R
```diff
@@ -0,0 +1,163 @@
...
+
+  public = list(
+
+    ## constructor method ---------------
+
+    #' @description
+    #' Initialize a new instance of the class.
+    #' @return A new instance of `DataCombined` object.
+
+    # empty constructor because we prefer that the users add simulationResults
```
msevestre, line 53: Why do you prefer this? Can't you make them optional parameters at least?

IndrajeetPatil: @PavelBal made a point that otherwise the signature of the constructor will grow to an unwieldy length because each individual method's parameters will need to be included.

I am fine either way.

PavelBal: @msevestre I have made a proposal for the API of `DataCombined`, please review it here: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/674

msevestre:
Sorry I didn't see that this PR was a draft yesterday. I'll continue commenting when the draft mode is removed.

IndrajeetPatil:
CI is failing with the following error, and I am not sure how to fix that:

```r
     -----------------------------------
ERROR: dependency 'rClr' is not available for package 'ospsuite'
* removing 'C:/Users/appveyor/AppData/Local/Temp/1/Rtmp8I3pqT/Rinst13fc673e19ba/ospsuite'
      -----------------------------------
ERROR: package installation failed
```

IndrajeetPatil:
Let's see if `git cherry-pic c459b84797785876c67952db4cd81c6210d2fc71` does the trick.

IndrajeetPatil:
Yupp, CI working again, but the tests are failing now LOL

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,741 @@
...
+
+# both `DataSet` and `SimulationResults` provided -------------
+
+test_that("dataCombined - both dataSet and SimulationResults provided", {
+  skip_if_not_installed("R6")
+
+  expect_true(R6::is.R6Class(DataCombined))
+
+  # load the simulation
+  sim <- loadSimulation(file.path(getwd(), "..", "data", "MinimalModel.pkml"))
```
PavelBal, line 23: Check out the file "helper-for-tests" and how it can be used e.g. in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/6021fc93197eb3551009072aff77975606cf1568/tests/testthat/test-simulation.R#L4

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,741 @@
...
+  simResults <- importResultsFromCSV(
+    simulation = sim,
+    filePaths = file.path(getwd(), "..", "data", "Stevens_2012_placebo_indiv_results.csv")
```
PavelBal, line 26: Also here - use `getTestDataFilePath` to generate the path.

R/data-combined.R
```diff
@@ -0,0 +1,625 @@
...
+    initialize = function() {
+    },
+
+    ## setter methods ---------------
+
+    #' @description
+    #' Add simulated data.
+    #' @return `DataCombined` object containing simulated data.
+
+    addSimulationResults = function(simulationResults,
```
PavelBal, line 122: I would prefer to have the order of arguments similar to that in `getOutputValues`, namely

`simulationResults, quantitiesOrPaths, population, individualIds, names, groups`

R/data-combined.R
```diff
@@ -0,0 +1,625 @@
...
+
+      # make sure that length of groups and names is the same
+      # additionally since these lists are going to have strings as elements,
+      # convert NULLs to NAs
+      groups <- .validateListArgs(groups, simulationResults$allQuantityPaths)
+      names <- .validateListArgs(names, quantitiesOrPaths %||% simulationResults$allQuantityPaths)
+
+      # save the original object as it is; useful for `$toDataFrame()` method
+      # styler: off
+      private$.simulationResults              <- simulationResults
```
PavelBal, line 150: It does not make sense like this, because we can add multiple simulation results (one after the other). Actually, once the results have been added, we do not need them any more, so no need to store in the object?

R/data-combined.R
```diff
@@ -0,0 +1,625 @@
...
+    #' A dataframe of simulated and/or observed data (depending on instances of
+    #' which objects have been added to the object).
+    #'
+    #' Note that the order in which you enter different object matters. If you
+    #' first enter observed data and simulated data later, the rows will also be
+    #' ordered in the same way.
+    #'
+    #' @return A dataframe.
+
+    toDataFrame = function() {
```
PavelBal, line 276: I am not comfortable with the fact that the method's name is "toDataFrame", but it returns a tibble. I would prefer a data.frame.

R/data-combined.R
```diff
@@ -0,0 +1,625 @@
...
+    # since the groups are always going to be strings
+    x <- NA_character_
+  } else {
+    x
+  }
+}
+
+.validateListArgs <- function(arg = NULL, expectedLength) {
+  if (!is.null(arg)) {
+    validateIsSameLength(expectedLength, arg)
```
PavelBal, line 615: You need `validateIsOfLength ` here.

PavelBal (changes requested):

PavelBal:
- Let's have a field` $names `to get the names of all data in the `DataCombined`-object

- Why do we need "paths" in the df? I do not think it is required, let's remove the column unless we have a use case for it.


----------------
```
- #Add all simulation results
dataCombined <- DataCombined$new()
dataCombined$addSimulationResults(simulationResults = simResults)
df <- dataCombined$toDataFrame()

#Add them again
dataCombined$addSimulationResults(simulationResults = simResults)
df <- dataCombined$toDataFrame()
```

If adding data with the same name without specifying the names, it should overwrite the previous - otherwise we have non-unique names!! Now we have the data double in the data frame.

-----------
- df entries are sorted by x values? It makes sense to sort them by names, I think..
Behavior (of sorting) is different between adding with names and without.
-------------
###
Cannot add simulation results where one result is assigned to the group and the other not

```
dataCombined <- DataCombined$new()
dataCombined$addSimulationResults(simulationResults = simResults, quantitiesOrPaths = c("Organism|Lumen|Stomach|Dapagliflozin|Gastric emptying", "Organism|Lumen|Stomach|Metformin|Gastric retention"),
                                  names = list("Dapa GE", "Dapa GR"),
                                  groups = list("GE liquid", NULL))
```
								  
								  
Actually, cannot specify a group at all ( even with
 ```
dataCombined$addSimulationResults(simulationResults = simResults, quantitiesOrPaths = c("Organism|Lumen|Stomach|Dapagliflozin|Gastric emptying", "Organism|Lumen|Stomach|Metformin|Gastric retention"),
                                  names = list("Dapa GE", "Dapa GR"),
                                  groups = list("GE liquid", NULL))
```


- Check for usecase - add data with name "Group1", but WITHOUT a group, and add another data with name "SomeName", and assign it to the group "Group1". Then should not be considered in one group!!!

File: R/data-combined.R (hunk 6 above)
Comment (IndrajeetPatil) [reply R/data-combined.R:276]: There are a number of reasons why we should prefer returning a tibble dataframe, as opposed to a classical dataframe, and I would like to get into them during a meeting, but for now, I just want to emphasize that this is not false advertising 😅 

``` r
is.data.frame(tibble::tibble())
#> [1] TRUE
```

<sup>Created on 2022-01-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

PavelBal: I actually don't know how R can return `TRUE` for this. Tibbles and data.frames have different subsetting syntax, I see it as a huge problem.

IndrajeetPatil: Yeah, true. But, the good news is that, in most cases, the tibble subsetting behavior is what the users will find intuitive or unsurprising, though.

You can find an extensive list of differences between tibble and dataframe subsetting behaviors here:
https://tibble.tidyverse.org/articles/wip/subset.html

IndrajeetPatil:
@PavelBal I have addressed most of your points in recent commits. About remaining points:

> df entries are sorted by x values? It makes sense to sort them by names, I think..

Currently, we are not doing any sorting at all. Should we be returning a dataframe sorted alphabetically by `name` column?

> Behavior (of sorting) is different between adding with names and without.

I can't reproduce this.

I also can't reproduce any of your other points. Let's have a call today to talk more about this.

IndrajeetPatil:
> Check for use case - add data with name "Group1", but WITHOUT a group, and add another data with name "SomeName", and assign it to the group "Group1". Then should not be considered in one group!!!

As far as I can see, this is going to be impossible to avoid because for datasets for which no grouping is specified, we use their names as grouping column value, and so for the edge case you provided, they _will_ have the same grouping.

The only way I see out of this is to error when users enter group names that are already present in dataset names.
Is that an acceptable solution for you?

IndrajeetPatil:
To discuss:

- [ ] Should there be a `$print()` method? If yes, what should it print?
- [ ] Which datasets to use for examples in the documentation? I couldn't find anything appropriate in `inst/extdata` folder. Can a few of the datasets from `tests/data` folder moved to `inst/extdata`? The largest file (`Stevens_2012_placebo_indiv_results.csv`) is 1.2 MB.

IndrajeetPatil:
To discuss with @PavelBal and @msevestre: 

- [ ] Should there be a `$print()` method? If yes, what should it print? (I think we should print only the names of the datasets and the current group map, i.e., basically just print the `$groupMap` binding output.)
- [ ] Which datasets to use in manual examples and vignettes? I couldn't find anything appropriate in `inst/extdata` folder. Can a few of the datasets from `tests/data` folder moved to `inst/extdata`? The largest file (`Stevens_2012_placebo_indiv_results.csv`) is 1.2 MB.
- [ ] Should we have the same `names` argument across three different methods, where it means different things? I've named them like this because Pavel wanted them so. But I personally think that when the argument provides alternate names for datasets (in `$addSimulationResults()` and `$addDatasets()`), the argument can be called `newNames`, while when it is used to filter certain datasets for data transformation (in `$setDataTransformations()`), it can be called `includeNames`.  
- [ ] Active bindings are all getter methods. Should they have a naming pattern of `get*()`?

R/data-combined.R
```diff
@@ -0,0 +1,807 @@
...
+          optionalMessage = "Data sets are assigned to groups when adding via `$addSimulationResults()` or `$addDataSets()` methods."
+        ))
+      }
+    },
+
+    #' @field dataTransformations A dataframe specifying which offsets and scale
+    #'   factor values were specified by the user for each dataset.
+
+    # just a way to access whatever was specified
+    dataTransformations = function(value) {
```
PavelBal, line 388: Should be `dataTransfors` to be consistent with the setter.

PavelBal (changes requested):

PavelBal:
`dataTransforms` are not reset when calling `setDataTransforms` with default values:

```
dataCombined <- DataCombined$new()
dataCombined$addDataSets(dataSets = dataSet)
dataCombined$setDataTransforms(names = "Stevens_2012_placebo.Placebo_distal", xOffsets = -1, yOffsets = 100, xScaleFactors = 0.1, yScaleFactors = 10)
dataCombined$dataTransformations

# A tibble: 1 x 6
  name                                dataType xOffsets xScaleFactors yOffsets yScaleFactors
  <chr>                               <chr>       <dbl>         <dbl>    <dbl>         <dbl>
1 Stevens_2012_placebo.Placebo_distal observed       -1           0.1      100            10

df <- dataCombined$toDataFrame()
```

now calling `setDataTransforms` with defaults:

```
dataCombined$setDataTransforms(names = "Stevens_2012_placebo.Placebo_distal")
dataCombined$dataTransformations

# A tibble: 1 x 6
  name                                dataType xOffsets xScaleFactors yOffsets yScaleFactors
  <chr>                               <chr>       <dbl>         <dbl>    <dbl>         <dbl>
1 Stevens_2012_placebo.Placebo_distal observed        0             1        0             1


df_ori <- dataCombined$toDataFrame()
```

df_ori should have the untransformed values, but they are transformed

R/data-combined.R
```diff
@@ -0,0 +1,807 @@
...
+    #'
+    #' A dataframe.
+    #'
+    #' The length (number of columns) and number of rows of a dataframe will
+    #' change depending on instances of which objects and how many objects have
+    #' been provided to the object. During the object's lifetime, this method
+    #' will return an updated version of a dataframe that combines observed and
+    #' simulated data contained in provided objects.
+    #'
+    #' For the same reasons, the dataframe might contain some or all of the
```
PavelBal, line 289: This is not entirely correct. The columns depend on the meta data stored in the `DataSet` objects. The number or names of meta data are not restricted (e.g. Source, Sheet, etc are meta data).

R/data-combined.R
```diff
@@ -0,0 +1,807 @@
...
+
+    ## print method -----------------
+
+    #' @description
+    #' Print the object to the console
+
+    print = function() {
+      private$printClass()
+
+      # TODO: what do we want to print here?!
```
PavelBal, line 340: At least the names of the data sets and the groups. Maybe just the "groupMap"?

PavelBal:
When adding `DataSet`, all meta data columns should be at the end of the data frame. E.g. "Group Id" got mixed in between "name" and "xValues"
image


I prefer the following sorting:

`name - group - dataType - xValues - xDimension - xUnit - yValues - yErrorValues - yDimension - yUnit - yErrorType - yErrorUnit - molWeight 
` are always present
- and then all meta data

R/data-combined.R
IndrajeetPatil: I am actually going the other way: changing the method name to `setDataTransformations`. 

_Transforms_ sounds awkward. Better to be verbose here.

PavelBal:
lol, setting data transforms removes all other data sets from the data frame:


Here, adding 6 data sets. The generated df has all six data sets included
```
> dataCombined <- DataCombined$new()
> dataCombined$addDataSets(dataSets = dataSet)
> dataCombined$names
[1] "Stevens_2012_placebo.Placebo_distal"   "Stevens_2012_placebo.Placebo_proximal" "Stevens_2012_placebo.Placebo_total"    "Stevens_2012_placebo.Sita_dist"        "Stevens_2012_placebo.Sita_proximal"   
[6] "Stevens_2012_placebo.Sita_total"      
> df <- dataCombined$toDataFrame()
unique(df$name)
[1] "Stevens_2012_placebo.Placebo_distal"   "Stevens_2012_placebo.Placebo_proximal" "Stevens_2012_placebo.Placebo_total"    "Stevens_2012_placebo.Sita_dist"        "Stevens_2012_placebo.Sita_proximal"   
[6] "Stevens_2012_placebo.Sita_total"  
```

Now setting data transforms for one data set. The resulting df has only this data set included

```
> dataCombined$setDataTransforms(names = "Stevens_2012_placebo.Placebo_distal", xOffsets = -1, yOffsets = 100, xScaleFactors = 0.1, yScaleFactors = 10)
> df <- dataCombined$toDataFrame()
> unique(df$name)
[1] "Stevens_2012_placebo.Placebo_distal"
```

PavelBal:
The column "molecularWeight" should always be present, as it might be required for unit conversion (or you store molecular weight somewhere else...)

For simulation results, use `simResults$simulation$molWeightFor()` to get the molecular weight of the output (if defined).

IndrajeetPatil:
> For simulation results, use simResults$simulation$molWeightFor() to get the molecular weight of the output (if defined).

This needs to be implemented in `simulationResultsToDataFrame()`, and not in this object. Once it's there, it will automatically be included in the `DataCombined` output.

I will handle this in a separate PR.

IndrajeetPatil:
> lol, setting data transforms removes all other data sets from the data frame:

Ah, that's what I thought you were saying that this function should do. I will change its behavior so that all the datasets are retained, irrespective of which ones selected for transforming.

msevestre:
Sorry I thought this was still a WIP. Can you rename the title so that WIP is gone and I'll review

IndrajeetPatil:
@msevestre Sorry, I am just waiting for Pavel to give a green signal as a user that everything makes sense, and then you can review the code and the implementation. Will ping you one more time once this happens.

R/data-combined.R
IndrajeetPatil: I've updated the description with a copy-paste of what you have written here.

PavelBal:
`yErrorValues` column is missing when adding simulation results. Simulation results never have errors, but this leads to inconsistent outputs. Maybe populate them with `NA`?

PavelBal:
When adding simulation results:
- Column `xDimension` is missing
- the column `yUnit` appears before `yDimension` (as opposed to when adding `DataSet`)
- Add column `yErrorType` and population with `NA`
- Column `molWeight` is missing (see another comment how to get it)

PavelBal:
> This needs to be implemented in simulationResultsToDataFrame(), and not in this object. Once it's there, it will automatically be included in the DataCombined output.
> 
> I will handle this in a separate PR.

The model file did not contain the mw, I have updated the pkml.

When you call `molWeightFor`, it returns the value in the base unit - which is `kg/µmol`. This is not the unit the user should expect, so we have to convert it first to the common unit `g/mol`: 

mw <- simResults$simulation$molWeightFor("Organism|Lumen|Stomach|Dapagliflozin|Gastric emptying")
mw <- ospsuite::toUnit(quantityOrDimension = ospDimensions$`Molecular weight`, values = mw, targetUnit = ospUnits$`Molecular weight`$`g/mol`)

R/data-combined.R
```diff
@@ -0,0 +1,860 @@
...
+      # for method chaining
+      invisible(self)
+    }
+  ),
+
+  # active bindings ---------------------------------------------------
+
+  active = list(
+
+    #' @field names A vector of unique names of `DataSet` objects and/or
```
PavelBal, line 312: "Unique names of data set contained in the `DataCombined` object" or something like this... We do not store `DataSet` or simulation results after adding.

R/data-combined.R
```diff
@@ -0,0 +1,860 @@
...
+
+      # for method chaining
+      invisible(self)
+    },
+
+    #' @description
+    #'
+    #' Transform raw data with required offsets and scale factors.
+    #'
+    #' @param xOffsets,yOffsets,xScaleFactors,yScaleFactors Either a numeric
```
PavelBal, line 215: Add a description how transformations are applied. I.e. for x/yValues, first the offset is added and then the value is multiplied by factor; for yErrorValues only multiplying by factor.

PavelBal (commented):
You can invite Michael for review after addressing those issues.

R/data-combined.R
```diff
@@ -0,0 +1,860 @@
...
+#'
+#' Additionally, it allows:
+#'
+#' - Grouping different (simulated and/or observed) datasets with grouping
+#' labels. If no grouping is specified, name of the dataset is used as a
+#' grouping label.
+#'
+#' - Transforming data (with given offsets and scale factors).
+#'
+#' @param simulationResults Object of type `SimulationResults` produced by
```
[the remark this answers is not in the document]
IndrajeetPatil: > @param tags that appear before the class definition are automatically inherited by all methods, if needed

Ref: https://github.com/r-lib/roxygen2/issues/996#issuecomment-572588902

This is why some parameters are defined before the class definition, while others (like `xOffsets`, `yOffsets`, etc.) are defined before the class.

At any rate, this is what it looks like only in the source code. For users, parameters will only be displayed before respective methods, so it should not be overwhelming for them.

R/simulationResultsToDataFrame.R
```diff
@@ -54,83 +54,50 @@ simulationResultsToDataFrame <- function(simulationResults,
...
-  if (is.character(cols) && length(cols) == 1) {
-    if (cols == "all") {
-      # If all, take all
-      cols <- names(data)
-    } else {
-      # Surely, a regex
-      cols <- grep(cols, names(data), value = TRUE)
-    }
+  # if molecular weight is not present, extract it and add as a new column
+  if (!"molWeight" %in% names(df)) {
```
PavelBal, line 73: `molWeight` is never part of `SimulationResults`, so don't have to check this.

And this function belongs to "utilities-simulation-results", I think.

IndrajeetPatil:
@msevestre The PR is now ready for your review! 🥵

msevestre:
@IndrajeetPatil @PavelBal I am going to review first for code and then I'll try the API. So my comments will come in multiple batches. Sorry about this but there is just way too much for me to review at once

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      if (is.vector(simulationResults)) {
+        stop(messages$errorOnlyOneSupported())
+      }
+
+      # validate the object type (`NULL` elements are not accepted)
+      validateIsOfType(simulationResults, SimulationResults, nullAllowed = FALSE)
+
+      # validate argument type and length
+      groups <- validateVectorArgs(groups, length(quantitiesOrPaths %||% simulationResults$allQuantityPaths), type = "character")
+      names <- validateVectorArgs(names, length(quantitiesOrPaths %||% simulationResults$allQuantityPaths), type = "character")
```
msevestre, line 182: I seee that you have a method that validate length and type... why not

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+        private$.simResultsToDF(
+          simulationResults = simulationResults,
+          quantitiesOrPaths = quantitiesOrPaths,
+          population        = population,
+          individualIds     = individualIds,
+          names             = names,
+          groups            = groups
+        )
+      )
+      private$.dataCombined <- private$.extractXYData(private$.dataCombined)
```
msevestre, line 202: I have seen the exact same code before. Create a private method that will always set those fields (groupmap, names, extractXY data) at the same time.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      private$.groupMap     <- private$.extractGroupMap(private$.dataCombined)
+      private$.names        <- private$.extractNames(private$.dataCombined)
+      # styler: on
+
+      # set up data transformations
+      self$setDataTransformations(names)
```
msevestre, line 208: This was also done at the same time. Maybe it belongs to this new method?

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+
+    #' @description
+    #'
+    #' Transform raw data with required offsets and scale factors.
+    #'
+    #' @param xOffsets,yOffsets,xScaleFactors,yScaleFactors Either a numeric
+    #'   scalar or a list of numeric quantities specifying offsets and scale
+    #'   factors to apply to raw values. The default offset is `0`, while
+    #'   default scale factor is `1`, i.e., the data will not be modified. If a
+    #'   list is specified, it should be the same length as `names` argument.
```
msevestre, line 222: I do not see in the code where you check this.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+
+  active = list(
+
+    #' @field names A vector of unique names of datasets contained in the
+    #'   `DataCombined` class instance.
+
+    # just a way to access whatever was specified
+    names = function(value) {
+      if (missing(value)) {
+        private$.names
```
msevestre, line 328: return early. If you use the return() key word here, you can remove the else and therefor avoid nesting

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+        ))
+      }
+    },
+
+    #' @field groupMap A dataframe specifying which datasets have been grouped
+    #'   together and the name and the nature (observed or simulated?) of the
+    #'   data. If a dataset was not assigned to any group, this is denoted by
+    #'   `NA` in the dataframe.
+
+    # just a way to access whatever was specified
```
msevestre, line 342: lol.. what?

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+    #
+    # If no grouping is specified for a dataset, the group will be `NA`.
+    #
+    # While visualizing such data, the datset name can instead be used as a
+    # grouping variable, but this association will happen in the plotting
+    # function itself, and not in the output produced by the current object.
+
+    .addGroupCol = function(data, groups = NULL) {
+      if (!is.null(groups)) {
+        data <- data %>%
```
msevestre, line 463: This is the first line of code where I would really need to see comment to understand what it does. Grouping, nesting, ungrouping, mutate, unnest etc.. 
from the function comment, I am expecting this to add a column to data. but it does clearly something else
Also data is a df right? should we call it df? or dataFrame ? Thougths?

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+          dplyr::group_by(name) %>%
+          tidyr::nest() %>%
+          dplyr::ungroup() %>%
+          dplyr::mutate(group = groups) %>%
+          tidyr::unnest(cols = c(data))
+      } else {
+        data <- data %>% dplyr::mutate(group = NA_character_)
```
msevestre, line 470: Why is this use case even allowed? Adding a NA group?

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      }
+
+      return(data)
+    },
+
+    # update the combined dataframe in place
+
+    .updateDF = function(dataCurrent = NULL, dataNew = NULL) {
+      # if there is already data, add new data at the bottom
```
msevestre, line 479: This is a fantastic comment. That is exactly the stuff that I love.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      # since the `toDataFrame()` method arranges the data by dataset name,
+      # this will not introduce any order effects, i.e., it doesn't matter
+      # if the user entered `DataSet` objects first and then `SimulationResults`
+      # objects, or vice versa
+      if (!is.null(dataCurrent) && !is.null(dataNew)) {
+        # check for duplicated datasets between the current and the new dataframes
+        dupDatasets <- intersect(unique(dataCurrent$name), unique(dataNew$name))
```
msevestre, line 486: I don't understand this. How can we have duplicate?

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+    # first group the dataset by its unique identifier, which is name, and then
+    # nest the rest of the dataframe in a list column, and then
+    # ungroup the dataframe as mutating column is not going to be by group, and then
+    # assign new `names` vector to the existing `name` column, and then
+    # unnest the list column
+
+    .renameDatasets = function(data, names = NULL) {
+      if (!is.null(names)) {
+        data <- data %>%
+          dplyr::group_by(name) %>%
```
msevestre, line 569: I think I need to understand the structure of the data before commenting any further. I am going to check unit tests

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      data %>%
+        dplyr::select(group, name, dataType) %>%
+        dplyr::distinct() %>%
+        dplyr::arrange(group, name)
+    },
+
+    # extract unique and sorted dataset names from the combined dataframe
+
+    .extractNames = function(data = NULL) {
+      if (!is.null(data)) {
```
msevestre, line 596: In all thos functions, exist early. if (is.Null) return NULL. This will remove a lot of nesting

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+
+    # create copies of original data before changing the respective columns
+    # this helps to revert or change the applied transformations if needed
+
+    .extractXYData = function(data = NULL) {
+      if (is.null(data)) {
+        return(NULL)
+      }
+
+      data <- dplyr::mutate(data,
```
msevestre, line 614: mutate creates a copy?

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      }
+
+      # Update private fields for the new setter call
+      #
+      # - extract dataframe and append it to the combined dataframe
+      # - extract original XY data into separate columns
+      # - extract group map
+      # - extract dataset names
+
+      # styler: off
```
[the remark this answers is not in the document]
PavelBal: Yes we did.

R/data-combined.R
```diff
@@ -0,0 +1,163 @@
...
+  # private fields and methods -----------------------------------
+
+  private = list(
+    .dataSet           = NULL,
+    .simulationResults = NULL
+  ),
+
+  # other object properties --------------------------------------
+
+  lock_objects = TRUE,
```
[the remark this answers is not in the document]
IndrajeetPatil: Those flags in RStudio turn into a nicely laid out content structure. In a file with a class that spans over 800 lines of code, it's crucial for one to easily toggle back and forth between different parts of the file. That's why I have these flags everywhere. They have nothing to do with increasing the readability of the code. 

Screenshot 2022-01-21 16 00 11

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+#' # print the object
+#' myCombDat
+#' @docType class
+#' @export
+
+DataCombined <- R6::R6Class(
+  classname = "DataCombined",
+  inherit = Printable,
+
+  # public fields and methods ------------------------------------
```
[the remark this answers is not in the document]
IndrajeetPatil: See my comment above. Once again, this has nothing to do with readability and everything to do with convenience of working with files with a large amount of code that needs to be quickly traversed.

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,53 @@
+test_that("it can set and retrieve the dimension of a data column", {
+  skip_if_not_installed("R6")
```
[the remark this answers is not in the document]
IndrajeetPatil: I have removed these conditions. 

But note that what these instructions are telling is not to run the tests if the needed package is not available, which is reasonable.

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,53 @@
...
+
+  # load the simulation
+  sim <- loadSimulation(file.path(getwd(), "..", "data", "MinimalModel.pkml"))
+  simResults <- importResultsFromCSV(
+    simulation = sim,
+    filePaths = file.path(getwd(), "..", "data", "Stevens_2012_placebo_indiv_results.csv")
```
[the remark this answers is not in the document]
IndrajeetPatil: Yes, I am already using it. Not sure why this Outdated file popped up for your review.

R/data-combined.R
IndrajeetPatil: Sorry, poor phrasing on my part. I have changed this to the following:

```r
        # check if the new dataset(s) entered are already present in the
        # internal combined dataframe
        dupDatasets <- intersect(unique(dataCurrent$name), unique(dataNew$name))

        # if this is the case, then replace the older datasets with the newer
        # versions of the same datasets
        #
        # e.g. someone can all `$addSimulationResults(dataSet1)` and then again
        # call `$addSimulationResults(dataSet1)` with the same class instance.
        # In this case, dataframe created in the latter call will replace the
        # one created in the former call
        if (length(dupDatasets) > 0L) {
          dataCurrent <- dplyr::filter(dataCurrent, !name %in% dupDatasets)
        }
```

R/data-combined.R
IndrajeetPatil: Good point! I was doing this in some functions, while not in others. I am doing this consistently now.

R/data-combined.R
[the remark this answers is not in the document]
msevestre: I understand but I think this could hide a design problem. There is absolutely no reason for any class to have 800 lines of code unless it is machine generated code.
The fact that this is required to be able to navigate the codes can (does not have to) mean that this class does too much. Of course, there is a  lot of validation stuff that does not belong there. Once this is removed, how big is the file. Do you still need this region?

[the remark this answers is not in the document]
IndrajeetPatil: 800 LOC involves hundreds of lines of comments, empty lines, functions that might be moved to the utilities package.
So the actual code base is much smaller than 800 LOC.

> Do you still need this region?

Before I answer this, I am actually curious as to why you are so opposed to having comments that act as flags in an IDE and give the file a navigation structure. How is that detrimental to anything?

R/data-combined.R
IndrajeetPatil: Good point!

R/data-combined.R
IndrajeetPatil: We want the returned dataframe to consistently contain the `group` column. If the entire group column is `NA`, the user immediately understands that they haven't set up any groupings. 

Plus, this way, the dataframe doesn't change its length (no. of columns) depending on whether grouping was provided or not.

R/data-combined.R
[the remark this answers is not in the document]
msevestre: I am not opposed to it. I am traumatized with code using region maybe
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.UI/Views/ProteinExpression/ExpressionDataView.cs#L115
 
900 lines of code, lots of region, lots of comments, no one knows how to fix this because the code does a million unrelated tasks.

So my first instinct, as an old fashioned developer granted, is to frown upon code where regions are needed in order to navigate the code. It is not because the IDE supports it that is is a *good* thing. 

I do not say that this is bad or good. I am just saying that that relying on region to be able to navigate in the code *could* mean that the code does too much. <= Do you agree with this statement?

R/data-combined.R
IndrajeetPatil: I am punting this to @PavelBal, who wanted this scenario to be supported.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+    initialize = function() {
+    },
+
+    ## setter methods ---------------
+
+    #' @description
+    #' Adds observed data.
+    #' @return `DataCombined` object containing observed data.
+
+    addDataSets = function(dataSets, names = NULL, groups = NULL) {
```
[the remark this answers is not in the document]
IndrajeetPatil: Since you are the second person who has said this, I have changed how parameters are specified now.

Only the common parameters live before the class definition, while others before class methods.

R/data-combined.R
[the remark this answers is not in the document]
msevestre: and here some of the reasons why we do not use region in our code
https://softwareengineering.stackexchange.com/questions/53086/are-regions-an-antipattern-or-code-smell

I do not know the R world, and maybe this is different and much welcomed. My goal is to start a discussion to make sure we have at least talked about those different way of organizing code (As you said, this is purely about code organization)

R/data-combined.R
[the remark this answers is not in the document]
msevestre: That view above is pretty cool indeed.

msevestre:
@IndrajeetPatil Please do not close discussion when you answer a question otherwise I cannot find them again. Like I'd like to find the one regarding the ValidateArgs xx method. The link is broken. 
Good point in general .Who should close the discussion lol

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      # validate the object type (`NULL` elements are not accepted)
+      # if list of instances is provided, `purrr::walk` will check each element
+      if (is.list(dataSets)) {
+        purrr::walk(.x = dataSets, .f = ~ validateIsOfType(.x, "DataSet", nullAllowed = FALSE))
```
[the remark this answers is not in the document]
IndrajeetPatil: No, not at all. `purrr::walk()` won't change anything about the arguments supplied to it. 

Here is a simple example:

``` r
library(ospsuite.utils)

x <- list("x", "y", 1)

purrr::walk(x, ~ validateIsOfType(.x, type = "character"))
#> Error in validateIsOfType(.x, type = "character"): tryCatch: argument '.x' is of type 'numeric', but expected 'character'!

x
#> [[1]]
#> [1] "x"
#> 
#> [[2]]
#> [1] "y"
#> 
#> [[3]]
#> [1] 1
```

<sup>Created on 2022-01-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      } else {
+        validateIsOfType(dataSets, "DataSet", nullAllowed = FALSE)
+      }
+
+      # validate argument type and length
+      groups <- validateVectorArgs(groups, objCount(dataSets), type = "character")
```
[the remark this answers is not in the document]
IndrajeetPatil: Are you working with outdated versions of these files by any chance? 

Because these methods don't live in the class any more. They are currently internal package functions. For example, here are docs for `objCount`, which make it clear why we need it:

image

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      # instead and simplify to a character vector
+      if (is.list(dataSets) && is.null(names(dataSets))) {
+        names(dataSets) <- purrr::map_chr(dataSets, .f = ~ purrr::pluck(.x, "name"))
+      }
+
+      # if alternate names are provided for datasets, use them instead
+      # for `NULL` elements in a list, the original names will be used
+      if (!is.null(names) && is.list(dataSets)) {
+        # lengths of alternate names and objects should be same
+        names <- validateVectorArgs(names, length(names(dataSets)), type = "character")
```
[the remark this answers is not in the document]
IndrajeetPatil: The custom validation function is doing much more than that, and that's why I am using that one over `valideIsOfLength`:

image

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: But we *are* checking their types right away. It's just that the validation is happening in the same order as the argument.

So first `dataSets` argument is validated, then `names`, and then `groups`.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+    #' @return `DataCombined` object containing simulated data.
+
+    addSimulationResults = function(simulationResults,
+                                    quantitiesOrPaths = NULL,
+                                    population = NULL,
+                                    individualIds = NULL,
+                                    names = NULL,
+                                    groups = NULL) {
+      # list or a vector of `SimulationResults` class instances is not allowed
+      if (is.vector(simulationResults)) {
```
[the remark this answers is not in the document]
IndrajeetPatil: `is.vector()` will cover both `c(simResults1, simResults2, ...)` and `list(simResults1, simResults2, ...)` possibilities.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+        validateIsOfType(dataSets, "DataSet", nullAllowed = FALSE)
+      }
+
+      # validate argument type and length
+      groups <- validateVectorArgs(groups, objCount(dataSets), type = "character")
+
+      # anticipate that although a list of `DataSet` objects might be entered,
+      # they might not have names associated with them in the argument list
+      #
+      # in such cases, go inside each element of the list (`purrr::map()`) and
```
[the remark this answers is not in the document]
IndrajeetPatil: I was using map_chr, which is just a specific variant of map. I've updated the comment.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+                                    population = NULL,
+                                    individualIds = NULL,
+                                    names = NULL,
+                                    groups = NULL) {
+      # list or a vector of `SimulationResults` class instances is not allowed
+      if (is.vector(simulationResults)) {
+        stop(messages$errorOnlyOneSupported())
+      }
+
+      # validate the object type (`NULL` elements are not accepted)
```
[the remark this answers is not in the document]
IndrajeetPatil: Slowly removing redundant comments.

R/data-combined.R
IndrajeetPatil: Relic of something that was supposed to be removed. Removed now.

IndrajeetPatil:
Oops. I have unresolved everything. Hope how you can find your respective comment.

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: I have changed the position of parameter documentation now to the way you were expecting.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+
+    ## constructor method ---------------
+
+    #' @description
+    #' Initialize a new instance of the class.
+    #'
+    #' @return A new (empty) instance of `DataCombined` class.
+
+    # empty constructor
+    initialize = function() {
```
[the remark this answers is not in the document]
IndrajeetPatil: > Remove it if it's not use

Fair enough. Removed the constructor.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+    ## setter methods ---------------
+
+    #' @description
+    #' Adds observed data.
+    #' @return `DataCombined` object containing observed data.
+
+    addDataSets = function(dataSets, names = NULL, groups = NULL) {
+      # validate the object type (`NULL` elements are not accepted)
+      # if list of instances is provided, `purrr::walk` will check each element
+      if (is.list(dataSets)) {
```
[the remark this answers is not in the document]
IndrajeetPatil: Are you sure you are looking at the latest version of the files?

Because what you are referring to was present in the files a couple of days ago. Since then I have changed the file and we no longer allow anything in a list that is not a DataSet class instance.

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: > Why don't you allow the user to add data set and simulation results while instancing the object?

This is why (for more, see this [comment](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/674#issuecomment-992434941))

image

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: I have removed these. 

Your opposition to this alignment reminds me of the following passage from *The Art of Readable Code* (p. 39) 

image

Can you please explain why you are opposed to it?

R/data-combined.R
IndrajeetPatil: How about to cover the following context:

I run a simulation and call `myCombDat$addSimulationResults(simResults1)`.

But then I realized I did something wrong while running the simulation, so I rerun it. In this case, it will be convenient to just do the same call again with the assurance that the new results will replace the old ones.

Does that make sense? I am guessing this as a non-user. Maybe Pavel has a much more compelling case to make.

R/data-combined.R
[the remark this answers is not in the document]
msevestre: oh I see. But validateIsOfType accepts a list as well as parameter. SO I think you don't need the "walk" at all in that case ? Or am I mistaken?

R/data-combined.R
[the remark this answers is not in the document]
msevestre: I love those examples BTW. How do you do this?

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: > validateIsOfType accepts a list as well as parameter

I don't think so. I've already created an issue about this:
https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/issues/21

It will be nice if it could also handle a list.

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: Using `{reprex}` package: https://reprex.tidyverse.org/

R/data-combined.R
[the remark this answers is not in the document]
msevestre: lol... I don't think I have programmer fears...but I have **consistency OCD**. If I can make it work without having to think about it, please let's use it. If I have to mentally do sthg beside the styler, then no. Can we modify the styler to do it, then yes. I am very flexible but also very lazy. If it cannot be automated, it won't be consistent. Ready to bet on this :)

R/data-combined.R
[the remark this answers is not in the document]
msevestre: I thougth c(simResults1, simResults2, ...) was a list... (even though it is initialized as a vector).
So why do we use is.list in other places then?

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: Those are just the default arguments for the R6 class constructor:
https://r6.r-lib.org/reference/R6Class.html

I am listing them explicitly to protect against the possibility that the defaults might change in the future, and also to remind ourselves of the options we are agreeing to.

image

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+      groups <- validateVectorArgs(groups, objCount(dataSets), type = "character")
+
+      # anticipate that although a list of `DataSet` objects might be entered,
+      # they might not have names associated with them in the argument list
+      #
+      # in such cases, go inside each element of the list (`purrr::map()`) and
+      # extract (`purrr::pluck()`) the name from the object itself and use them
+      # instead and simplify to a character vector
+      if (is.list(dataSets) && is.null(names(dataSets))) {
+        names(dataSets) <- purrr::map_chr(dataSets, .f = ~ purrr::pluck(.x, "name"))
```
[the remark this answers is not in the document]
IndrajeetPatil: Good point!

I have refactored the code here that avoids changing the inputs.

R/data-combined.R
IndrajeetPatil: No, poor phrasing on my part. 

Mutate creates new columns, and I am saving copies of raw data in these new columns before they are transformed in place using provided offsets and scale factors.

R/data-combined.R
```diff
@@ -0,0 +1,676 @@
...
+    #'   class.
+    #'
+    #' @description
+    #' Adds observed data.
+    #'
+    #' @return `DataCombined` object containing observed data.
+    addDataSets = function(dataSets, names = NULL, groups = NULL) {
+      # if list of instances is provided, `purrr::walk()` will check each element
+      # `purrr::walk()` will in no way modify objects provided to it
+      if (is.list(dataSets)) {
```
PavelBal, line 85: validateIsOfType(dataSets, DataSet) works perfectly, let's use it.

IndrajeetPatil: Strange that `validateIsOfType` works with `list` input of some type like `DataSet`, but not with others.

https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/issues/21

R/data-combined.R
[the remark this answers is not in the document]
PavelBal: Same question as Michael... `c(simResults1, simResults2, ...)` will always be `is.list = true`?

PavelBal:
> Good point in general .Who should close the discussion lol

The one who opens it. So the reviewer can check that it has been addressed properly.

IndrajeetPatil:
We badly need to do this: https://github.com/Open-Systems-Pharmacology/Suite/issues/103

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,1072 @@
...
+)
+
+# import observed data (will return a list of DataSet objects)
+dataSet <- loadDataSetsFromExcel(
+  xlsFilePath = getTestDataFilePath("CompiledDataSetStevens2012.xlsx"),
+  importerConfiguration = DataImporterConfiguration$new(getTestDataFilePath("ImporterConfiguration.xml"))
+)
+
+# same as dataSet, but with deliberately changed entries for testing
+dataSet2 <- loadDataSetsFromExcel(
```
PavelBal, line 19: Instead of creating a new file, you can modify `DataSet` objects by calling `setValues`.

IndrajeetPatil:
> So the reviewer can check that it has been addressed properly.

I'd argue that the reviewer should check the commits made after the comments are resolved to see if they were comments were addressed 🤷

R/data-combined.R
msevestre: I think this is the vector vs list vs simple type vs refs.
There is sthg weird going on for sure with integner, number list (primitive types)
Maybe this is because a vector of DataSet is a list of DataSet and our weird  obj <- c(obj) works as expected

R/data-combined.R
```diff
@@ -0,0 +1,676 @@
...
+      private$.names <- private$.extractNames(private$.dataCombined)
+
+      # set up data transformations
+      self$setDataTransformations(names)
+
+      # for method chaining
+      invisible(self)
+    },
+
+    #' @param simulationResults Object of type `SimulationResults` produced by
```
PavelBal, line 120: Can't you use @inherit to inherit the documentation from `getOutputValues`? This would ensure that the documentation is always up to date and save some lines of code.

IndrajeetPatil: I had tried this. Unfortunately, this is not supported for `R6`. 

In general, `{roxygen2}` doesn't have a good support for R6,as can be surmised from [dozens of pending issues](https://github.com/r-lib/roxygen2/issues?q=is%3Aopen+is%3Aissue+label%3A%22R6+%3Asix%3A%22).

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,1072 @@
...
+    myCombDat$dataTransformations,
+    structure(
+      list(
+        name = "Stevens_2012_placebo.Placebo_total",
+        xOffsets = 0,
+        yOffsets = 0,
+        xScaleFactors = 1,
+        yScaleFactors = 1
+      ),
+      class = c("tbl_df", "tbl", "data.frame"),
```
msevestre, line 73: You are testing the framework here. This is not required.. If this changes for some reasons, your test will fail. Do not test the internal of stuff that you do not OWN. 
Testing line 67 to 71 is what you need

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,1072 @@
...
+      row.names = c(NA, -1L)
+    )
+  )
+
+
+  # checking dataframe methods
+  df <- myCombDat$toDataFrame()
+  expect_s3_class(df, "data.frame")
```
msevestre, line 81: There has to be a method like is.data.frame or sthg like this. Why do we care that this is a s3 class?

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,1072 @@
...
+          "Organism|Lumen|Stomach|Metformin|Gastric retention distal",
+          "Organism|Lumen|Stomach|Metformin|Gastric retention proximal",
+          "Organism|Lumen|Stomach|Metformin|Gastric retention"
+        ),
+        xOffsets = c(0, 0, 0, 0, 0),
+        yOffsets = c(0, 0, 0, 0, 0),
+        xScaleFactors = c(1, 1, 1, 1, 1),
+        yScaleFactors = c(1, 1, 1, 1, 1)
+      ),
+      class = c("tbl_df", "tbl", "data.frame"),
```
msevestre, line 119: same here.

R/data-combined.R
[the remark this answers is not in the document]
msevestre: >the possibility that the defaults might change in the future,

That would really surprise me lol. I don't think we need to list them here but documenting it somewhere is good. For me, it looks like we are overriding the default behavior which is not the case. No need to change

msevestre:
@IndrajeetPatil I am always reviewing on github diff (uptodate by default) or after a pull locally. Not sure how it can be out of sync but I'll refresh my page next time. Sorry about this

R/data-combined.R
msevestre: sure

R/data-combined.R
```diff
@@ -0,0 +1,680 @@
...
+        #
+        # retaining the offset and scale factor parameters might confuse the
+        # user about whether the transformations are supposed to be carried out
+        # by the user using these values or these transformations have already
+        # been carried out
+        #
+        # internal copies of raw data and redundant paths columns also need to
+        # be left out
+        dplyr::select(
+          -dplyr::matches("^paths$"),
```
msevestre, line 655: Can you explain the path and original? I am assuming they are added somewhere programatically but I can't find it
also what happens  if some meta data have the name original in them? they are going to be removed. It is too much of a common name to remove it like this. Maybe our internal keyword should be encapsulated in marker (typically succession of one or two chars such as &%original&%. And you define constants for these keywords  so that you do not have to remembner what this is

R/data-combined.R
```diff
@@ -0,0 +1,680 @@
...
+          -dplyr::matches("offsets$|scalefactors$"),
+          -dplyr::contains("original")
```
msevestre, line 657: this is the one that I also don't get

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,1072 @@
...
+  expect_equal(unique(df$name), names(dataSet)[[1]])
+
+  # only SimulationResults input
+
+  # selecting all paths
+  myCombDat2 <- DataCombined$new()
+  myCombDat2$addSimulationResults(simResults)
+
+  # data transformations should already be setup with the defaults
+  expect_equal(
```
msevestre, line 103: I am not sure if this is how tests are written in R so I am just commenting from otjher languages point of view.
WE typically try to not test too much at the same time. So one test_that should have only one (a few) expect

If you are changing the context (new dataSets, adding removing etc.. ) you are testing something else. 

You can also use the keyword context() to group tests that belong together. Then in your test that, you can be more explicit about what you are testing. 

Example: Instead of a comment "data transformations should already be setup with the defaults"

This shoulld be the name of the test

test_that("data transformations should already be setup with the defaults")

does it make sense?

R/data-combined.R
```diff
@@ -0,0 +1,680 @@
...
+    },
+
+    ## print method -----------------
+
+    #' @description
+    #' Print the object to the console
+
+    print = function() {
+      # group map contains names and nature of the datasets and grouping details
+      private$printLine("DataCombined", addTab = FALSE)
```
PavelBal, line 295: First line of `print` should be

`private$printClass()`

Then you don't need `printLine("DataCombined")`

IndrajeetPatil: This is actually deliberate on my part. 

I'll double-check, but I am pretty sure that the alignment looked off if I used `private$printClass()` and this is why I used this workaround.

R/data-combined.R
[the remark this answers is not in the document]
PavelBal: I would say, having the region markers for "public", "active", and "private" are ok because it is sometimes indeed not easy to navigate between them.

"setter methods", "getter methods", "print method" (sure it is a print method, if it is called `print`), "private methods" etc just add lines of code and introduce "too much structure" (or too many hierarchy levels), so I would get rid of them.

R/data-combined.R
```diff
@@ -0,0 +1,680 @@
...
+    #'
+    #' In the returned dataframe, the following columns will always be present:
+    #'
+    #' name - group - dataType - xValues - xDimension - xUnit - yValues -
+    #' yErrorValues - yDimension - yUnit - yErrorType - yErrorUnit - molWeight
+    #'
+    #' @note
+    #'
+    #' The molecular weight (in `molWeight` column) is in `g/mol` units.
+
```
PavelBal, line 271: Why do you leave an additional empty line (or even multiple empty lines) between the documentation and the function? This is not the default behavior of "generate roxygen code"?

I liked our "old" style better, this style makes the documentation look a bit "detached" from the function and also introduces additional lines of code.

PavelBal: And it is not consistent... so definitely not something automated.

R/data-combined.R
```diff
@@ -0,0 +1,680 @@
...
+        return(data)
+      }
+
+      data %>%
+        tidyr::nest(data = -name) %>%
+        dplyr::mutate(name = names) %>%
+        tidyr::unnest(cols = c(data))
+    },
+
+    # update the combined dataframe in place
```
PavelBal, line 460: A blank line before a comment is good, a line after is not necessary.

R/data-combined.R
PavelBal: Then `private$printClass()` should be fixed ;) We should use general functions if we already have them, instead of having "custom solutions" in every file.

R/data-combined.R
```diff
@@ -65,13 +65,8 @@
 DataCombined <- R6::R6Class(
   classname = "DataCombined",
   inherit = Printable,
-
```
PavelBal: You don't have to remove everything :D 

# public fields and methods ------------------------------------

# active bindings ---------------------------------------------------

# private -----------------------------------

# private fields-----------------------------------

are definitely useful.

R/utilities-data-combined.R
```diff
@@ -0,0 +1,164 @@
...
+#' @keywords internal
+#' @noRd
+
+objCount <- function(x) {
+  # `is.vector()` can handle both atomic vectors and lists, i.e.
+  # both `is.vector(c(1, 2))` and `is.vector(list(1, 2))` will be `TRUE`
+  if (is.vector(x)) {
+    l <- length(x)
+  } else {
+    l <- length(list(x))
```
msevestre, line 33: I am wondering about this. SInce x is not a vector, list(x) will have one item no? Isn't the value always 1?
Code can be simplified by returning early

R/utilities-data-combined.R
```diff
@@ -0,0 +1,164 @@
...
+#' @examples
+#'
+#' validateVectorArgs(list(1, 2, NA, NULL), 4L, "numeric")
+#' validateVectorArgs(c(1, 2, NA, NA_complex), 4L, "numeric")
+#'
+#' @keywords internal
+#' @noRd
+
+validateVectorArgs <- function(arg = NULL, expectedLength = NULL, type) {
+  # return early if argument was not specified
```
msevestre, line 68: I think it's an error? Does it make sense to have args NULL?

R/utilities-data-combined.R
```diff
@@ -0,0 +1,164 @@
...
+  # Note that `purrr::map()` will return a list
+  arg <- purrr::map(arg, ~ toMissingOfType(.x, type))
+
+  # validate the type of arguments
+
+  # `nullAllowed = TRUE` is necessary because `NULL` in vector arguments is
+  # used to specify no change for the corresponding dataset
+  # `purrr::walk()` is needed below because validation helper functions
+  # are called only for their side effects
+  purrr::walk(.x = arg, .f = ~ validateIsOfType(.x, type, nullAllowed = TRUE))
```
msevestre, line 89: WE should put a TODO here to replace once validateIsOfType with a list is fixed

R/utilities-data-combined.R
```diff
@@ -0,0 +1,164 @@
...
+#'
+#' @examples
+#'
+#' validateVectorArgs(list(1, 2, NA, NULL), 4L, "numeric")
+#' validateVectorArgs(c(1, 2, NA, NA_complex), 4L, "numeric")
+#'
+#' @keywords internal
+#' @noRd
+
+validateVectorArgs <- function(arg = NULL, expectedLength = NULL, type) {
```
msevestre, line 67: Naming might be a bit misleading because the function actually returns a modified values of args potentially whereas all other validate functions have no return values... I am not good at naming things lol but I would not have expected a returned value (also the examples are not showing this)

R/utilities-data-combined.R
```diff
@@ -0,0 +1,164 @@
...
+#' flattenList(list(1, 2, 3, NA), type = "numeric")
+#' flattenList(list(TRUE, FALSE, NA), type = "integer")
+#'
+#' @return An atomic vector of desired type.
+#'
+#' @keywords internal
+#' @noRd
+
+flattenList <- function(x, type) {
+  if (is.list(x)) {
```
msevestre, line 119: Isn't it also an error if this is not a list?

R/utilities-data-combined.R
```diff
@@ -0,0 +1,164 @@
...
+#'
+#' flattenList(list(1, 2, 3, NA), type = "numeric")
+#' flattenList(list(TRUE, FALSE, NA), type = "integer")
+#'
+#' @return An atomic vector of desired type.
+#'
+#' @keywords internal
+#' @noRd
+
+flattenList <- function(x, type) {
```
msevestre, line 118: Trying to understand this method

> toto <- c(1,2,3)
> flattenList(toto)
[1] 1 2 3
> toto
[1] 1 2 3
> flattenList(dataSet)
Error in flattenList(dataSet) : 
  argument "type" is missing, with no default

It behaves differently with a vector of integer and a list of objects (note that I did not specify a type in the first place and it worked)

Why do I have to specify a type to flatten a list?

R/utilities-data-combined.R
PavelBal: There is nothing wrong with `validateIsOfType`, only `validateIsInteger`
https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/issues/21

R/utilities-data-combined.R
```diff
@@ -0,0 +1,164 @@
...
+  if (is.list(x)) {
+    x <- switch(type,
+                "character" = purrr::flatten_chr(x),
+                "numeric" = ,
```
msevestre, line 122: what does the, do in this case? e..g there is no specific purr method? Should not we call the default? or is this going to call the one for double?

R/utilities-simulation-results.R
```diff
@@ -195,3 +195,105 @@ importResultsFromCSV <- function(simulation, filePaths) {
...
+#' # convert to a dataframe
+#' simulationResultsToDataFrame(results)
+#' @return
+#'
+#' SimulationResults object as data.frame with columns IndividualId, Time, paths,
+#' simulationValues, unit, dimension, TimeUnit.
+#'
+#' @export
+
+simulationResultsToDataFrame <- function(simulationResults,
```
msevestre, line 223: Very neat.  Some very cool stuff going on in this method

R/utilities-data-combined.R
IndrajeetPatil: In R's `switch` statement, if nothing is specified, the switch trickles downwards to find the value. 

So 

```r
"numeric" = ,
 "real" = ,
 "double" = purrr::flatten_dbl(x),
```

is the same as writing

```r
```r
"numeric" = purrr::flatten_dbl(x),
 "real" =  purrr::flatten_dbl(x),
 "double" = purrr::flatten_dbl(x),
```
```

Comment (IndrajeetPatil) [reply R/utilities-data-combined.R:122]: So DRYed it up ;)

File: R/data-combined.R [hunk 64]
```diff
@@ -0,0 +1,628 @@
...
+#' - Grouping different (simulated and/or observed) datasets with grouping
+#' labels.
+#'
+#' - Transforming data (with given offsets and scale factors).
+#'
+#' @param groups A string or a list of strings assigning the data set to a
+#'   group. If an entry within the list is `NULL`, the corresponding data set is
+#'   not assigned to any group (and the corresponding entry in the dataframe
+#'   will be an `NA`). If provided, `groups` must have the same length as
+#'   `dataSets` and/or `simulationResults`. If no grouping is specified for any
```
msevestre, line 19: This is not accurate. It is the same lenght as simulationResults$quantityPath

```
> length(simResults)
[1] 13
> myCombDat$addSimulationResults(simulationResults = simResults, groups = list("distal"))
 Error in validateIsOfLength(arg, expectedLength) : 
$: Object should be of length '5', but is of length '1' instead. myCombDat: Object should be of length '5', but is of length '1' instead. addSimulationResults: Object should be of length '5', but is of length '1' instead.
```

msevestre:
This looks good to me .I think we need to merge this asap. Instead of creating new tasks in this branch, let's create issue so that we can have small and cohesive PR and commits. 

Two points that I want to make
1: @IndrajeetPatil You removed ALL regions although this was not requested. I think there is some sense to have a tree structure in your file for searching if it helps. We went from a deeply nested structure to nothing. Maybe there is a good happy medium
2: I think the code is fascinating to read because it's very well documented and explained. Very well done

I would suggest to address all other questions in separate issues  (that's why I started to create some). We can some final tweaks to the API etc.. by creating issues for each changes.

PavelBal (approved):

PavelBal:
Agree with Michael. Let's merge and continue working in issues.

IndrajeetPatil:
Please don't merge this yet. I still need to address a few comments before this can be merged. 

It should take me not more than a couple of days next week to address the rest of the unresolved code review comments. I won't be able to sleep properly if PR with unresolved comments is merged 😅 

Thanks.

msevestre:
Don't lose sleep over this @IndrajeetPatil 
Have a safe trip

R/utilities-data-combined.R
IndrajeetPatil: No. 

If a list, flatten it.
If a vector, just return it untouched since it's already flat.

That said, this should produce an error if non-vector. I have added a check for that.

R/utilities-data-combined.R
IndrajeetPatil: > It behaves differently with a vector of integer and a list of objects

Precisely. That's the whole point of it. 

If a list, it flattens it out to an atomic vector, in which case you need `type` because atomic vector needs to be of certain type.
If a vector, just return it untouched since it's already flat.

R/utilities-data-combined.R
IndrajeetPatil: Hmm, good point.

How about `cleanVectorArgs()`?

R/utilities-data-combined.R
IndrajeetPatil: I have already replaced this with `validateIsOfType()`. 
We don't have pure integer arguments anywhere, so that bug doesn't affect this feature.

R/utilities-data-combined.R
IndrajeetPatil: In this case, yes.

If `arg` is `NULL`, we use the defaults. E.g. `groups` or `names` are `NULL`, that's a valid argument.

If `expectedLength` is `NULL`, it means that the function is being used only to check the type and not the length.

R/utilities-data-combined.R
IndrajeetPatil: Good point. Done!

R/data-combined.R
IndrajeetPatil: Added them back in.

R/data-combined.R
```diff
@@ -0,0 +1,867 @@
...
+        # lengths of alternate names and objects should be same
+        names <- validateVectorArgs(names, length(names(dataSets)), type = "character")
+
+        # if any of the alternate names `NULL`, use original names
+        names <- ifelse(is.na(names), names(dataSets), names)
+      }
+
+      # Update private fields for the new setter call
+      #
+      # - extract dataframe and append it to the combined dataframe
```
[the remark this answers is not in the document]
IndrajeetPatil: I have tried to improve the quality of the comments and removed some obvious comments.

R/data-combined.R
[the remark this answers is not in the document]
IndrajeetPatil: No. `c(...)` function will always produce an atomic vector, while `list(...)` will create a non-atomic vector, aka list.

``` r
is.list(c(1, 2))
#> [1] FALSE
```

<sup>Created on 2022-02-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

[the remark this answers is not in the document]
msevestre: This is not true

> is.list(c(1, 2))
[1] FALSE
> is.list(c(sim1, sim2))
[1] TRUE

R/data-combined.R
IndrajeetPatil: Good point. 

I'have added comments to explain the _why_ and _how_ for what's going on in `.addGroupCol()` function.

I've also changed the return logic.

R/data-combined.R
IndrajeetPatil: I've added `TODO` marker here to revisit this in case this is supported by _roxygen2_.

tests/testthat/test-data-combined.R
```diff
@@ -0,0 +1,1072 @@
...
+dataSet2 <- loadDataSetsFromExcel(
+  xlsFilePath = getTestDataFilePath("CompiledDataSetStevens2012v2.xlsx"),
+  importerConfiguration = DataImporterConfiguration$new(getTestDataFilePath("ImporterConfiguration.xml"))
+)
+
+# empty initialization ----------------------------
+
+test_that("dataCombined - initialization", {
+  # check class generator
+  expect_true(R6::is.R6Class(DataCombined))
```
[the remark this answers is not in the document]
IndrajeetPatil: I've removed these tests.

R/data-combined.R
IndrajeetPatil: I've changed the docs to reflect that.

IndrajeetPatil:
Interesting. 

Okay, removing `is.vector()` checks for R6 objects in favor `is.list()`.

``` r
library(R6)

is.list(R6::R6Class())
#> [1] FALSE
is.list(c(R6::R6Class()))
#> [1] TRUE
```

<sup>Created on 2022-02-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

IndrajeetPatil:
Points to address from meeting:

- [ ] Use more specific way to remove original columns ("original" is too risky since this metadata can contain this in their names)

IndrajeetPatil:
This is ready to be merged!

Thank you both for your insightful and critical code review for this PR, which has significantly improved its quality!

Learned a lot about how one ought to be reviewing PRs, both at high-concept-level and low-code-level.

msevestre:
I am doing it!!! Woot woot

PR #693, opened 2021-12-11
closes #660

### LINKED ISSUE #660: Use `{ospsuite.utils}`
Should not we start doing this? @IndrajeetPatil

IndrajeetPatil:
- removes all functions which now reside in [utilities package](https://cran.r-project.org/web/packages/ospsuite.utils/ospsuite.utils.pdf)
- removes associated tests
- reexports the previously exported functions, so this is **NOT** a breaking change
- adds IP as author


IndrajeetPatil:
Need to wait until binaries are built for the package on CRAN.

msevestre:
>reexports the previously exported functions, so this is NOT a breaking change


Cool. How does this work?
I am ok with a breaking change since we are bumping the version from 10 to 11 BTW and I would like all packages to use the utils implementation

IndrajeetPatil:
> How does this work?

Reexports basically just import and export a function from another package's namespace so that it is available in the current package's namespace.

> I am ok with a breaking change 

Since this PR is only about relying on the utilities package, I didn't want any deletions in NAMESPACE.
I can make another PR with a breaking change where these functions are no longer exported from `{ospsuite}`.

IndrajeetPatil:
@pchelle and @abdullahhamadeh You can also make the same change for `{tlf}` and `{ospsuite.reportingengine}` packages.

Yuri05:
>@pchelle and @abdullahhamadeh You can also make the same change for {tlf} and {ospsuite.reportingengine} packages.

For tlf and RE I suggest to do this AFTER the qualification workflow in the reporting engine is released (or at least beta-released)

Yuri05 (approved):

pchelle (approved):

PavelBal (approved):

NAMESPACE
```diff
@@ -119,6 +120,16 @@ export(uniqueEntities)
 export(updatePKParameter)
 export(validateDimension)
 export(validateUnit)
+import(ospsuite.utils)
 import(rClr)
 import(stringr)
+importFrom(ospsuite.utils,"%||%")
```
msevestre, line 126: Why is this required? (e.g. why is line 123 not enough?)

IndrajeetPatil: This brings the `NAMESPACE` of `{ospsuite.utils}` in `{ospsuite}` package, so that you can use exported functions inside this package, BUT these functions will not be available to the user (because they are not exported).

```r
import(ospsuite.utils)
```

In order to re-export a function, it first needs to explicitly imported and then exported. This is why you see `importFrom` tags **only** for those functions which need to be exported:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/006c31500d19a03f51bfd9d294dfcac3ae263209/NAMESPACE#L126-L134)

msevestre: ok. I don't think we should export those methods at all. But this is not very important. Merging

msevestre (approved):

PR #695, opened 2021-12-13
closes #426

### LINKED ISSUE #426: Not displaying some documented functions
Using `help(package = ospsuite)`, you can get the list of all the documented methods and objects in the package, even if they are not exported.
Adding `#' @keyword internal` prevent the display for the user, but the function documentation is still accessible if the user ask specifically for it (e.g. `help(ApiConfig, package = ospsuite)`)

IndrajeetPatil:


Yuri05 (approved):

R/simulation.R
```diff
@@ -4,6 +4,7 @@ MODEL_CORE_SIMULATION_EXTENSIONS <- "OSPSuite.Core.Domain.ModelCoreSimulationExt
 #' @docType class
 #' @description  An OSPSuite simulation
 #' @format NULL
+#' @keywords internal
```
msevestre, line 7: What does internal means? Why aren't we exposing any of those classes for users to use if they wish?

R/standard-path.R
```diff
@@ -16,13 +16,13 @@ StandardContainer <- enum(c(
...
-  Height = toPathString(StandardContainer$Organism, "Height"),
-  Weight = toPathString(StandardContainer$Organism, "Weight"),
-  BMI = toPathString(StandardContainer$Organism, "BMI"),
-  BSA = toPathString(StandardContainer$Organism, "BSA"),
-  GestationalAge = toPathString(StandardContainer$Organism, "Gestational age"),
-  OntogenyFactorAlbumin = toPathString(StandardContainer$Organism, "Ontogeny factor (albumin)"),
+  Age                      = toPathString(StandardContainer$Organism, "Age"),
+  Height                   = toPathString(StandardContainer$Organism, "Height"),
+  Weight                   = toPathString(StandardContainer$Organism, "Weight"),
+  BMI                      = toPathString(StandardContainer$Organism, "BMI"),
```
msevestre, line 22: It this done via styler oder by hand?
Styler yes! By hand no

R/simulation.R
IndrajeetPatil: That's a good question, and whoever wrote these functions can answer it. I don't know why they are not exported currently. 

Note that since these functions haven't had `#' @export` tags so far, these functions have thus far been **not** available to the users.

Of course, if you want, I can add export tags to all `R6` classes we currently have, then all of them will become available in `NAMESPACE`.

IndrajeetPatil: I personally think we should be exporting all of these classes, but wasn't sure if it was deliberately left internal, so I didn't export them.

R/standard-path.R
PavelBal: Naaaah not doing it! One space all the time!

PavelBal:
Sorry @IndrajeetPatil, let's discuss first before making such changes. I have no idea why you add @internal sometimes or export other functions.

IndrajeetPatil:
>  I have no idea why you add @internal sometimes or export other functions.

I ask the same question to myself when I look at the package NAMESPACE 😅

R/simulation.R
msevestre: Aren't they exported by default? We can create those instances from REportingEngine.... that's why we assumed that this was the default behavior. 
I agree with you. All classes should be exported, unless they are not intended for public use but I do not see any at this moment
So what is the default behavior of an R6 class in a package? Internal, exporter or sthg else?

IndrajeetPatil: **Everything** that doesn't have `#' @export` roxygen2 tag is internal. 

So anything (R6 classes, functions, lists, etc.) that doesn't have this tag is currently only internal and not user visible.

msevestre: What do you mean with user visible.Clearly those classes can be used from other packages.. (OSPSuite.ReportingEngine or even OSPSUite.PArameterIdentification). Is this a matter of documentaton not available? Sorry if this is obvious but it is not for me. I can use those classes in a script even though they are internal

IndrajeetPatil: The non-exported objects will still be accessible via the `:::` operator, but those objects will not be in the package `NAMESPACE`.

This distinction is not that important if you never intend to get the package on CRAN, but it is otherwise.

Objects which are in the `NAMESPACE` are the only objects users are supposed to use, and the developer provides no guarantee that the non-exported functions will continue to work in the future releases. For example, if I decide to remove an object which was never exported, it is **not** considered a breaking change because it was never supposed to be used and was for internal purposes only.

For example, the `Printable` object is not exported by `{ospsuite}`, so when `{esqlabsR}` package needs to use this object, it [must use](https://github.com/esqLABS/esqlabsR/blob/a17c845025145b3bf51f3d59f24f8df466718fca/R/Plotable.R#L8) the `:::`, and not `::` operator because `Printable` is not in `{ospsuite}` package's `NAMESPACE`.

msevestre: I see. Still. you can use the object from a returned value (sim <- loadSimulation(...)) and use sim in your code (no :: or :::) here. I agree with you. Almost all R6 class should be exporter, specifically if they are part of exported API methods.

PR #705, opened 2021-12-17
fixes #594

### LINKED ISSUE #594: ImporterConfiguration: createForFile
Implement creating a configuration for a specific file. Requires extension of core to specify the excel sheet.

hannaei:
Not ready for merging - still unsure about the issue with dimensions. Also for the sheet "UnitsFromColumn" from file "CompiledDataSet.xlsx" the column "measurementUnit" is mapped as the measurement column, this is probably undesired behaviour.


tests/testthat/test-utilities.R
```diff
@@ -8,3 +8,21 @@ test_that("It returns the correct name of an enum entry", {
 test_that("It returns NULL if the entry does not exist", {
   expect_equal(netEnumName(enumType = enumType, 5L), NULL)
 })
+
+context("createConfigurationForFile")
```
Yuri05, line 12: I think the tst should be moved to `test-data-importer-configuration.R` where all other data importer configurtion tests are stored.

tests/testthat/test-utilities.R
```diff
@@ -8,3 +8,21 @@ test_that("It returns the correct name of an enum entry", {
...
+filePath <-getTestDataFilePath("CompiledDataSet.xlsx")
+
+test_that("It can create a DataImporterConfiguration from a XLS file", {
```
Yuri05, line 15: please deactivate the test under Linux, because now data importer does not work under Linux due to npoi problem
(s. e.g. https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/e9148614cf2085caef1cde0f7a915aaf5f76e954/tests/testthat/test-data-importer-configuration.R#L3-L4 )

Yuri05:
@hannaei As this PR is still work in progress, I converted it to draft. Click on "Ready for review" button when you think all open questions are clarified

PavelBal:
@hannaei Create a file `utilities-data-importer-configuration` and move the function there.

R/data-importer-configuration.R
```diff
@@ -70,7 +70,7 @@ DataImporterConfiguration <- R6::R6Class(
           return(NULL)
         }
         dimension <- rClr::clrGet(mappedColumn, "Dimension")
-        return(rClr::clrGet(dimension, "DisplayName"))
+        return(ospsuite.utils::ifNotNull(dimension, rClr::clrGet(dimension, "DisplayName")))
```
PavelBal, line 73: @hannaei Is there a use case where dimension can be null? Let's discuss this.

R/data-importer-configuration.R
```diff
@@ -114,45 +114,45 @@ DataImporterConfiguration <- R6::R6Class(
...
-      column <- private$.errorColumn
+      column <- private$.errorColumn()
       if (missing(value)) {
         return(ifNotNull(column, rClr::clrGet(column, "ColumnName")))
       }
       # If value is NULL, remove the error column
       if (is.null(value)) {
         rClr::clrCall(private$.dataImporterTask, "RemoveError", self$ref)
-        private$.errorColumn <- NULL
+        # private$.errorColumn <- NULL
```
PavelBal, line 140: You have to explain to me why you commented it out

R/data-importer-configuration.R
```diff
@@ -316,14 +318,21 @@ DataImporterConfiguration <- R6::R6Class(
...
+    },
+    .errorColumn = function() {
+      return(rClr::clrCall(private$.dataImporterTask, "GetError", self$ref))
+    },
     .addErrorColumn = function() {
       rClr::clrCall(private$.dataImporterTask, "AddError", self$ref)
-      column <- rClr::clrCall(private$.dataImporterTask, "GetError", self$ref)
-      private$.errorColumn <- column
-      mappedColumn <- rClr::clrGet(column, "MappedColumn")
+      # column <- rClr::clrCall(private$.dataImporterTask, "GetError", self$ref)
```
PavelBal, line 332: Remove comments?

PavelBal (changes requested):

R/data-importer-configuration.R
hannaei: E.g. when the wrong column is matched as the measurement column (like in the example I mentioned above in my first comment) and therefore no unit can be retrieved, dimension will be null.

PavelBal:
@hannaei Add this to `DataImporterConfiguration` in `active` section:

    #' @field ref Underlying .NET reference
    ref = function(value) {
      if (missing(value)){
        return(private$.ref)
      }
      private$.ref <- value
    },

PavelBal:
@msevestre What kind of magic do I have to perform to solve that EOL conflicts? The conflicting files are only because of EOL...

R/data-importer-configuration.R
```diff
@@ -8,9 +8,17 @@ DataImporterConfiguration <- R6::R6Class(
   inherit = DotNetWrapper,
   cloneable = TRUE,
   active = list(
+    #' @field ref Underlying .NET reference
```
PavelBal, line 11: This is an example where we have to change the `ref` of the object, so we overload the method.

msevestre:
@PavelBal I can fix this for you

R/data-importer-configuration.R
```diff
@@ -257,10 +264,6 @@ DataImporterConfiguration <- R6::R6Class(
       }
       super$initialize(ref)
       private$.dataImporterTask <- importerTask
-
-      private$.timeColumn <- rClr::clrCall(importerTask, "GetTime", ref)
```
PavelBal, line 261: We don't cache the columns any more in the class, they are always retrieved from .NET object. Less error prone.

R/data-importer-configuration.R
msevestre: I don't see how this is a good idea!

WHY ?

R/data-importer-configuration.R
```diff
@@ -316,14 +319,18 @@ DataImporterConfiguration <- R6::R6Class(
   ),
   private = list(
     .dataImporterTask = NULL,
-    .timeColumn = NULL,
-    .measurementColumn = NULL,
-    .errorColumn = NULL,
+    .timeColumn = function() {
```
PavelBal, line 322: No cache, always call rClr.

R/utilities-data-importer-configuration.R
```diff
@@ -0,0 +1,43 @@
...
+  importerConfiguration <- DataImporterConfiguration$new()
+  importerTask <- getNetTask("DataImporterTask")
+  if (is.null(sheet)) {
+    ref <- rClr::clrCall(importerTask, "CreateConfigurationFor", filePath)
+  } else {
+    ref <- rClr::clrCall(importerTask, "CreateConfigurationFor", filePath, sheet)
+  }
+  importerConfiguration$ref <- ref
+  # set timeColumn dimension and unit to default ("Time" and "h") if it could not be read from file
+  if (importerConfiguration$timeUnit == "?") {
```
PavelBal, line 35: This must be done here because the user cannot set the dimension of x values.

PavelBal (approved):

R/data-importer-configuration.R
PavelBal: Otherwise we must change the constructor. Currently it gets a path to a configuration file, but not a `ref`. So either we change it to receiving a `ref` - but in this case creating from a file must be implemented through a function - or we allow to set ref from ouside.

OR the constructor can receive a `ref` or a path to the file - which I do not like at all.

msevestre: i need to see how this is being used. I don't understand. I'll check it out. Please do not merge

IndrajeetPatil:
Before I review this, can this feature branch please catch up with the develop branch?

I see some changes who are undoing changes in develop. For example, copy or Printable was removed but it's back in this PR.

PavelBal:
@IndrajeetPatil merged

tests/testthat/helper-for-tests.R
```diff
@@ -1,22 +1,27 @@
...
-  actionWithFile(newFile)
-  file.remove(newFile)
-}
+
+getTestDataFilePath <- function(fileName) {
+  dataPath <- file.path(getwd(), "..", "data", fsep = .Platform$file.sep)
+  file.path(dataPath, fileName, fsep = .Platform$file.sep)
+}
+
+getExtDataFilePath <- function(fileName) {
```
PavelBal, line 7: Another helping function so we can create paths to files stored in `extdata`, soo we can avoid duplication of files for tests and examples.

IndrajeetPatil:
Builds are still failing because one example seems to fail to run:

```r
> xlsFilePath <- system.file("extdata", "CompiledDataSet.xlsx", package = "ospsuite")
> importerConfiguration <- createImporterConfigurationForFile(xlsFilePath)
Error in rClr::clrCall(importerTask, "CreateConfigurationFor", filePath) : 
  Type:    OSPSuite.Infrastructure.Import.Core.UnsupportedFileTypeException
Message: The type of file that you are trying to open is not currently supported
Method:  OSPSuite.Infrastructure.Import.Core.IDataSourceFile For(System.String)
Stack trace:
   at OSPSuite.Infrastructure.Import.Core.DataSourceFileParser.For(String path)
   at OSPSuite.Infrastructure.Import.Services.Importer.LoadFile(IReadOnlyList`1 columnInfos, String fileName, IReadOnlyList`1 metaDataCategories)
   at OSPSuite.Infrastructure.Import.Services.AbstractDataImporter.ConfigurationFromData(String dataPath, IReadOnlyList`1 columnInfos, IReadOnlyList`1 metaDataCategories, String sheetName)
   at OSPSuite.R.Services.DataImporterTask.CreateConfigurationFor(String dataPath, String sheetName)
Calls: createImporterConfigurationForFile -> <Anonymous>
Execution halted
```

R/data-importer-configuration.R
msevestre: ok so there is definitely no need for this This is used once, where the DataImportConfiguraiton object is instantiated. 
Like we do for many other objects, pass it in the constructor so that no one will ever create this object by hand but always use the service

R/utilities-data-importer-configuration.R
```diff
@@ -0,0 +1,43 @@
...
+#'   importerConfiguration = importerConfiguration,
+#'   importAllSheets = TRUE
+#' )
+#' @export
+createImporterConfigurationForFile <- function(filePath, sheet = NULL) {
+  validateIsString(filePath)
+
+  importerConfiguration <- DataImporterConfiguration$new()
+  importerTask <- getNetTask("DataImporterTask")
+  if (is.null(sheet)) {
```
msevestre, line 28: Do this first. Get the ref and then instantiate the importerConfiguration

R/utilities-data-importer-configuration.R
```diff
@@ -0,0 +1,43 @@
...
+    ref <- rClr::clrCall(importerTask, "CreateConfigurationFor", filePath)
+  } else {
+    ref <- rClr::clrCall(importerTask, "CreateConfigurationFor", filePath, sheet)
+  }
+  importerConfiguration$ref <- ref
+  # set timeColumn dimension and unit to default ("Time" and "h") if it could not be read from file
+  if (importerConfiguration$timeUnit == "?") {
+    column <- rClr::clrCall(importerTask, "GetTime", importerConfiguration$ref)
+    mappedColumn <- rClr::clrGet(column, "MappedColumn")
+    rClr::clrSet(mappedColumn, "Dimension", getDimensionByName(enc2utf8(ospDimensions$Time)))
```
msevestre, line 38: feels like this code belongs int he importerConfiguration object itself not in the service

R/data-importer-configuration.R
```diff
@@ -21,7 +29,7 @@ DataImporterConfiguration <- R6::R6Class(
     #' @field timeUnit If `timeUnitFromColumn` is `FALSE`, unit of the values in time column
     #' If `timeUnitFromColumn` is `TRUE`, name of the column with units of the values in time column
     timeUnit = function(value) {
-      column <- private$.timeColumn
+      column <- private$.timeColumn()
       mappedColumn <- rClr::clrGet(column, "MappedColumn")
```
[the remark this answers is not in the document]
msevestre: Last, the service is reference in the domain object which is NOT something we do normally. You have a utility class that does all the serialization etc... normally. So it is a bit surprising to me to see one method in utility, bus save etc.. in the domain object.

tests/testthat/helper-for-tests.R
PavelBal: I am talking only for files that are actually duplicated. We sometimes use the same files for test and examples.

Of course, we still should have files that are only for testing. This function should not replace `getTestDataFilePath`.

msevestre: ok

R/data-importer-configuration.R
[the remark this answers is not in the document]
PavelBal: I agree that we might want to refactor `DataImporterConfiguration`, but let's create an issue for that and do it in a separate PR. The here implemented feature does not rely on the internal implementation of `DataImporterConfiguration`.

Code looks similar but has small differences between methods, I could not find a more elegant solution quickly.

R/data-importer-configuration.R
PavelBal: @msevestre OK then I will change the constructor to get a `ref` as argument instead of file path to load from, and create a function `loadDataImporterConfiguration`, agreed? This will be a breaking change but probably it was never used before anyways.

msevestre: yes agreed

R/data-importer-configuration.R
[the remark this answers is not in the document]
msevestre: ok let's create an issue for this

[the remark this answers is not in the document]
msevestre: ok let's create an issue for this

R/data-importer-configuration.R
PavelBal: @msevestre Implemented.

R/utilities-data-importer-configuration.R
PavelBal: Done.

R/utilities-data-importer-configuration.R
PavelBal: Moved to constructor.

msevestre:
@PavelBal I am fixing the build issues

msevestre:
I have undone the change in the dimension as well as removed usage of $ref as set.
Hopefully nothing broke. Let see with the tests

PavelBal:
@msevestre something went terribly wrong. I did a couple commits today that addressed e.g. the `ref` usage and other things, and at the end I got it all running. I have no idea why the commit history does not show all those commits:

image

The code I have now locally is totally different from yours, and I actually have no idea what the hell happened.

image

I don't know whether I am severely hallucinating but I am 99% sure I also pushed all these commits and saw AppVeyor checking and passing the build at the end...

msevestre:
oh man. That's why we have to do better with long lasting PR. We are in a mess now with all those files ending changes etc
Can you push your changes again?

PavelBal:
> Before I review this, can this feature branch please catch up with the `develop` branch?
> 
> I see some changes that are undoing changes in `develop`. For example, copy of `Printable` was removed but it's back in this PR.

The `Printable` is again in `develop`!! what the hell is going on??

PavelBal:
Something got severely messed up (and not only with this PR)... This is driving me crazy.

R/data-importer-configuration.R
```diff
@@ -35,11 +35,11 @@ DataImporterConfiguration <- R6::R6Class(
...
     },
 
-    #' @field timeUnitFromColumn If `TRUE`, units of the values in time column
+    #' @field isTimeUnitFromColumn If `TRUE`, units of the values in time column
     #' are defined in the column `timeUnit`. If `FALSE`, the unit is defined by
-    #' `timeUnit`.
-    timeUnitFromColumn = function(value) {
-      column <- private$.timeColumn
+    #' the value of `timeUnit`.
+    isTimeUnitFromColumn = function(value) {
```
PavelBal, line 41: @msevestre You probably missed it (or it wasn't pushed, or got eaten by GitHub, whatever). I am naming this field now with `isXXX` to make clear it returns a logical.

R/data-importer-configuration.R
```diff
@@ -239,28 +238,31 @@ DataImporterConfiguration <- R6::R6Class(
...
-    #' @param configurationFilePath Path to the XML file with stored configuration
-    #' (e.g. create in PK-Sim or MoBi).
+    #' @param ref Reference to .NET DataImporterConfiguration object
     #' If `NULL` (default), an empty configuration with columns "Time" and
     #' "Measurement" is created.
+    #' @description
+    #' Initialize a new instance of the class
     #' @return A new `DataImporterConfiguration` object.
-    initialize = function(configurationFilePath = NULL) {
+    initialize = function(ref = NULL) {
```
PavelBal, line 247: No "either or", only `ref`. For creating from file, there is a function.

R/data-importer-configuration.R
```diff
@@ -239,28 +238,31 @@ DataImporterConfiguration <- R6::R6Class(
...
 
-      private$.timeColumn <- rClr::clrCall(importerTask, "GetTime", ref)
-      private$.measurementColumn <- rClr::clrCall(importerTask, "GetMeasurement", ref)
-      private$.errorColumn <- rClr::clrCall(importerTask, "GetError", ref)
+      # set timeColumn dimension and unit to default ("Time" and "h") if it is
+      # not set properyl in the .NET object. This could happen when creating
+      # a configuration for a xls-sheet and the unit could not be recognized.
+      # Because the user cannot set the Dimension if time values, this must be
+      # done during initialization phase.
+      if (self$timeUnit == "?") {
```
PavelBal, line 260: Probably only relevant when "guessing" the configuration for a sheet, but better having it in the constructor.

PavelBal:
@msevestre Merged and pushed again... Build is running.

msevestre:
AWESOME! REviewing and merging
Then I am going to edit all files using a command to hopefully have all those white spaces gone

R/data-importer-configuration.R
msevestre: ok great

R/data-importer-configuration.R
msevestre: good!

PR #706, opened 2021-12-17
Fixes #698 Add artifacts without version number to enable usage as pe…

### LINKED ISSUE #698: Add artifacts without version number to enable usage as permalink
Create **additional** artifacts **ospsuite.zip** and **ospsuite_<LinuxDistribution>.tar.gz** (without the version number), so that they can e used in the permalinks to the latest built package version (s. the discussion in Open-Systems-Pharmacology/OSPSuite.ReportingEngine#592)

Yuri05:
…rmalink

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,7 +57,7 @@ test_script:
 after_test:
   - rake "create_linux_build[%APPVEYOR_BUILD_VERSION%, %APPVEYOR_BUILD_FOLDER%, ubuntu18]"
   - rake "create_linux_build[%APPVEYOR_BUILD_VERSION%, %APPVEYOR_BUILD_FOLDER%, centOS7]"
-  - ps: copy ospsuite_*.zip ospsuite_windows.zip
+  - ps: copy ospsuite_*.zip ospsuite.zip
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -57,7 +57,7 @@ test_script:
 after_test:
   - rake "create_linux_build[%APPVEYOR_BUILD_VERSION%, %APPVEYOR_BUILD_FOLDER%, ubuntu18]"
   - rake "create_linux_build[%APPVEYOR_BUILD_VERSION%, %APPVEYOR_BUILD_FOLDER%, centOS7]"
-  - ps: copy ospsuite_*.zip ospsuite_windows.zip
+  - ps: copy ospsuite_*.zip ospsuite.zip
```
Yuri05, line 60: @msevestre 
needed to rename the zip file, otherwise it cannot be installed under windows
(under linux no problem, if the .tar.gz name differs from the package name, but under windows apparently it is)

msevestre: I knew I  had tried this before
and ospsuite_windows...you can install it on windows? WHAT

PR #711, opened 2022-01-11
Fixes #710 make ref readonly

### LINKED ISSUE #710: Make DotNetWrapper$ref read-only and override for classes if necessary
The reference to the .NET object in `DotNetWrapper` class is exposed as a public field. I propose to make this field read-only and override this behavior for classes when it should be possible.

Why? Changing `$ref` can destroy the connection between objects. Consider the `Simulation`-class and it's constructor:

```
    #' @description
    #' Initialize a new instance of the class
    #' @param ref Reference to .NET simulation object
    #' @param sourceFile (Optional) File used to load the simulation
    #' @return A new `Simulation` object.
    initialize = function(ref, sourceFile = NULL) {
      super$initialize(ref)
      private$.sourceFile <- sourceFile
      private$.buildConfiguration <- rClr::clrGet(self$ref, "BuildConfiguration")
      private$.settings <- SimulationSettings$new(rClr::clrGet(private$.buildConfiguration, "SimulationSettings"))
    }
```

Changing `$ref` after initialization would make `private$.buildConfiguration` (and as the consequence `private$.settings`) invalid.

msevestre:


msevestre (commented):
That should do it with some tests

PavelBal:
Yepp looks good, but let me try something tomorrow... I want to try if we still can allow setting ref by overriding 

```
    ref = function(value) {
      private$readOnlyProperty("ref", value, private$.ref)
    }
```

msevestre:
I don't understand what you mean. You want to define a new function in an object to set the ref? If you are deriving from the class, you have access to the private member so you can do what you want (as opposed to typical OOP languages where the prop is not accessible). So it would be possible to overwrite the ref for sure but this needs to be done ON purpose and not by accident

msevestre:
oh  I see you want to allow for override in a class...lot well that works by construction as you have access to private members..at least I think
But I would say this is really not what should be done. I'd like to understand a use case where this would be needed. This ref should be seen as  a private readonly and should never be changed

PavelBal:
Kk, can merge.

PR #715, opened 2022-01-21
molecular weight is always numeric

IndrajeetPatil:
So choose the appropriate type of NA

### DIFF
--- a/R/simulation.R
+++ b/R/simulation.R
@@ -82,7 +82,7 @@ Simulation <- R6::R6Class(
     molWeightFor = function(quantityPath) {
       ospsuite.utils::validateIsString(quantityPath)
       mw <- rClr::clrCall(self$ref, "MolWeightFor", quantityPath)
-      mw %||% NA
+      mw %||% NA_real_
     },
     #' @description
     #' Returns the applications ordered by start time associated to the quantity with path `quantityPath` or an empty list if not found


R/simulation.R
```diff
@@ -82,7 +82,7 @@ Simulation <- R6::R6Class(
     molWeightFor = function(quantityPath) {
       ospsuite.utils::validateIsString(quantityPath)
       mw <- rClr::clrCall(self$ref, "MolWeightFor", quantityPath)
-      mw %||% NA
+      mw %||% NA_real_
```
msevestre, line 85: What is this ? Real NA?

IndrajeetPatil: `NA` of real type, yes. The default NA in R is logical, so when we know the expected data type, we can use the specific variant of it. R usually does this coercion implicitly, so you never need to worry about it.

``` r
is.logical(NA)
#> [1] TRUE
is.character(NA_character_)
#> [1] TRUE
is.double(NA_real_)
#> [1] TRUE
is.complex(NA_complex_)
#> [1] TRUE
is.integer(NA_integer_)
#> [1] TRUE
```

<sup>Created on 2022-01-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

msevestre: Oh my...I have used NA in a few places in the code ....maybe we should do a quick search for other instances?

Just for me:
Why do we care to have NA typed? Isn't NA a short cut for not available?

IndrajeetPatil: > Why do we care to have NA typed? Isn't NA a short cut for not available?

We actually needn't care. It's just that my OCD acts up every time I see that the output type is inconsistent with what the users expect.

For example, when I convert simResults object to a dataframe, the `molWeight` column for molecular weight can be `NA` if the weights can't be extracted from the object. But the `NA` is of `logical` type (`<lgl>`), when, as a user, I would expect a `double`.

``` r
library(ospsuite)

simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
sim <- loadSimulation(simPath)
results <- runSimulations(sim)
df <- simulationResultsToDataFrame(results)

head(dplyr::select(df, molWeight))
#> # A tibble: 6 x 1
#>   molWeight
#>   <lgl>    
#> 1 NA       
#> 2 NA       
#> 3 NA       
#> 4 NA       
#> 5 NA       
#> 6 NA
```

<sup>Created on 2022-01-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

PavelBal:
I don't even understand what `NA` should be :D  `NULL` tells me that the certain variable is not assigned to any memory space, that is clear... `NaN` is the result of an invalid mathematical (like division by 0)... What is `NA`??

msevestre:
Lol....NA vs NaN. 
Let's create an issue for this as well. I would like to discuss when to use NA and when to use NaN...and when. To use NULL.

But for now this PR is great

IndrajeetPatil:
`NA` is missing data.

For example, if I am doing a survey and ask a simple yes(1) or no(0) question to 5 people, and the fourth person refuses to respond, the data will be: `c(1, 0, 0, NA, 1)`

R/simulation.R
PavelBal: Ok I get it now.. I hate R always trying to cast data types without any strict rules..

IndrajeetPatil:
> Oh my...I have used NA in a few places in the code ....maybe we should do a quick search for other instances?

I was working on this! See #716 now.

IndrajeetPatil:
> I would like to discuss when to use NA and when to use NaN...and when. To use NULL.

I think 99% of the time, especially in dataframes, we need to use `NA`.

And here is a succinct summary of what these unsual datatypes represent:

``` r
0 / 0 
#> [1] NaN
1 / 0
#> [1] Inf
NA / 0
#> [1] NA
c()
#> NULL
```

<sup>Created on 2022-01-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

IndrajeetPatil:
Also, I should say I quite like how R deals with operations with `NA`s:
https://www.linkedin.com/posts/indrajeet-patil-397865174_rstats-python-rprogramming-activity-6885126017476898816-OPz9

R/simulation.R
IndrajeetPatil: Dynamic type checking 🤷‍♂️

IndrajeetPatil:
For future reference:

image

PR #716, opened 2022-01-21
choosing the correct NAs

IndrajeetPatil:
Follow up to #715

### DIFF
--- a/tests/testthat/test-utilities-data-set.R
+++ b/tests/testthat/test-utilities-data-set.R
@@ -15,10 +15,10 @@ test_that("It can convert an empty data set", {
   expect_equal(
     dataSetToDataFrame(dataSet),
     data.frame(
-      name = character(0), xValues = logical(0), yValues = logical(0), yErrorValues = logical(0),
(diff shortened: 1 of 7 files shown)


tests/testthat/test-utilities-data-set.R
```diff
@@ -15,10 +15,10 @@ test_that("It can convert an empty data set", {
   expect_equal(
     dataSetToDataFrame(dataSet),
     data.frame(
-      name = character(0), xValues = logical(0), yValues = logical(0), yErrorValues = logical(0),
+      name = character(0), xValues = numeric(0), yValues = numeric(0), yErrorValues = numeric(0),
```
IndrajeetPatil, line 18: @msevestre This is why care about having the right kind of `NA`! 

Note how `xValues` and `yValues` used to be vectors of length 0 and type `logical` before!

msevestre: the fact that they were logical before and that we tested this and found that ok is what is more of a mystery for me lol
Very cool

PR #717, opened 2022-01-22
remove unnecessary namespace

IndrajeetPatil:
We import the entire utilities package, so :: is not necessary.

Plus each use of :: comes with a penalty of 60 microseconds in performance


msevestre:
Very nice. Much easier to read

PR #700, opened 2021-12-14
adds `simulationResultsToDataFrame()`; closes #701

IndrajeetPatil:
- docs and examples
- tests
- NEWS entry

For more, see: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/701

PR #732, opened 2022-01-30
Fixes #731

### LINKED ISSUE #731: add option in getSimulationTree to return parameter paths only

abdullahhamadeh:


R/path-explorer.R
```diff
@@ -47,15 +48,21 @@ nextStep <- function(listSoFar, originalString, arrayToGo) {
 #'
 #' liver_volume_path <- tree$Organism$Liver$Volume$path
 #' @export
-getSimulationTree <- function(simulationOrFilePath) {
+getSimulationTree <- function(simulationOrFilePath, returnParameterTree = FALSE) {
```
Yuri05, line 51: Would rename the new argument to `parameterPathsOnly` or `parametersOnly`
Because when I see `returnParameterTree = FALSE` I think "return the tree except the parameters"

PavelBal:
I would propose another solution. Instead of defining whether you want parameters only or not, define an argument `quantityType` which takes the class `Quantity`, `Molecule`, or `Parameter` as input (have to check whether we are exporting the classes or should use the string names instead), with default `Quantity`, and then just call `getAllEntitiesMatching`:

getSimulationTree <- function(simulationOrFilePath, quantityType = Quantity){
...
allPaths <- getAllEntityPathsIn(simulation, quantityType)
...
}

msevestre:
Agreed with @PavelBal

abdullahhamadeh:
Is it correct to say that quantitiesPaths = parameterPaths + moleculePaths + observerPaths?  

Would it make sense to also define an 'Output' option where outputPaths = moleculePaths + observerPaths?  I think it would be good to have access to the set of all possible entities that can be simulated and to have these separated from the other 'Quantity' entities.

PavelBal:
> Is it correct to say that quantitiesPaths = parameterPaths + moleculePaths + observerPaths?

I guess yes..

> I think it would be good to have access to the set of all possible entities that can be simulated and to have these separated from the other 'Quantity' entities.

You can have an output of all quantities, including parameters, so I am not sure this setting makes sense..

msevestre:
Yes output can be anything. 
Maybe you can pass a list of type to return e.g c(Parameter, Molecule, Observer)
Default would be Quantity which is all

abdullahhamadeh:
> Yes output can be anything. Maybe you can pass a list of type to return e.g c(Parameter, Molecule, Observer) Default would be Quantity which is all

OK.  And for completeness, since I don't see a `getAllObserverPathsIn`, how about defining something like

```
getAllObserverPathsIn <- function(container){
setdiff(getAllQuantityPathsIn(container), c(getAllParameterPathsIn(container),getAllMoleculePathsIn(container)))
}
```

?

R/path-explorer.R
```diff
@@ -66,7 +66,7 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
   }
 
   # Build a vector, with no duplicated entries, of all paths corresponding to entities in `simulation` that fall under the types specified in quantityType
-  allPaths <- sapply(quantityType,function(type){quantityTypeList[[type]](simulation)}) %>% unname %>% unlist %>% unique
+  allPaths <- unique(unlist(unname(sapply(quantityType,function(type){quantityTypeList[[type]](simulation)}))))
```
msevestre, line 69: Why removing pipes? I learnt from @IndrajeetPatil that it might be more efficient since it does not duplicate the data in memory (maybe I am completely wrong)

abdullahhamadeh: Lol I wasn't sure what to take away from #700: is the consensus that we should use pipes or not?  I couldn't find a single instance of pipes in the whole package, so I thought we're leaving them out for now.  Personally I prefer pipes too in cases like this.

msevestre: Pipe all the way....there is a big PR with lots of pipe coming

msevestre (approved):
I think this works for now. Probably using an enum instead of string would be clearer but as long as the message is clear...it's ok. 
I believe filtering could be done in core by passing a type instead of doing some diff but we need to add support for observers . Fine with me for now

Can you please run the styler as some files are not styled properly?

PR #737, opened 2022-02-01
move `simulationResultsToDataFrame` to utilities file

IndrajeetPatil:
I had done this change in the PR but had to revert it to resolve merge conflict.


DESCRIPTION
```diff
@@ -86,7 +86,6 @@ Collate:
     'simulation-run-options.R'
     'simulation-settings.R'
     'simulation.R'
-    'simulationResultsToDataFrame.R'
```
PavelBal, line 89: Why is it not exported any more?

IndrajeetPatil: It is exported. The field you are looking at is `Collate`, which is just showing to you that the `.R` file was removed.

Look at `NAMESPACE` file instead, and you will see that this function is exported.

PavelBal: Ah ok then.

PavelBal (approved):

PR #738, opened 2022-02-01
#736 data set name

### LINKED ISSUE #736: `DataSet` constructor should require a name if not created from a repository.

PavelBal:
Fixes #736


msevestre (approved):

IndrajeetPatil (approved):
Looks good to me.

I think we also need to log this change in `NEWS.md`.

PR #750, opened 2022-02-03
Update packages

msevestre:


R/simulation-batch-run-values.R
```diff
@@ -1,46 +1,50 @@
...
+    #' @field parameterValues Vector of parameter values used in a batch run
+    parameterValues = function(value) {
+      private$wrapVectorProperty("ParameterValue", "ParameterValues", value, "Values")
+    },
+    #' @field initialValues Vector of initial values used in a batch run
+    initialValues = function(value) {
+      private$wrapVectorProperty("InitialValue", "InitialValues", value, "MoleculeValues")
+    },
+    #' @field id Internal id of the batch run value
+    id = function(value){
```
msevestre, line 46: So i need to allow the id to be returned

PR #752, opened 2022-02-03
Fixes #748 test utilties simulation

### LINKED ISSUE #748: Add .gitattribute specifying what we should use for line ending etc..
based off this
https://github.com/alexkaratarakis/gitattributes/blob/master/R.gitattributes

And ensure that we set the line ending to crlf
see https://stackoverflow.com/a/2825829

msevestre:


tests/testthat/test-utilities-simulation.R
```diff
@@ -1,378 +1,378 @@
-context("exportIndividualSimulations")
```
msevestre, line 1: @IndrajeetPatil This is happening again. I think RStudio is doing some weird stuff. Hopefully this won't happen again with this change. Merging since this is the same as in the other repo

IndrajeetPatil: Yeah, I noticed 😅 

But ignoring the whitespace change is quite the neat trick to avoid any headache that might arise from this while reviewing the PR, so I am not too concerned about this anymore.

I wish this was GitHub's default settings for showing code changes.

PavelBal:
@msevestre Delete the branch https://github.com/Open-Systems-Pharmacology/OSPSuite-R/tree/748-add-git-attribute ?

msevestre:
Yes it's merged

PR #759, opened 2022-02-07
move test to the correct file

IndrajeetPatil:
Follow up to #737

Also runs test on linux which was previously skipped


IndrajeetPatil:
@msevestre Do you know where these changes are coming from? I didn't make any changes to this folder.

image

msevestre:
You probably did not update the submodule after pulling from develop. git status would have shown normally that this was required. So you probably commited the old version of the submodule again. This needs to be reverted

IndrajeetPatil:
> You probably did not update the submodule after pulling from develop

How am I supposed to know when I should or should not update submodules? I definitely can't keep checking commit hashes each time I pull, so I am assuming there is some trick to this.

Do you think it's worth mentioning how to update submodules here?

https://github.com/Open-Systems-Pharmacology/Suite/blob/develop/GIT_WORKFLOW.md

msevestre:
>How am I supposed to know when I should or should not update submodules?

I am assuming that git for windows should display this. I am using Cmder and i see it right away
image

the yellow indicates some changes and the M comes from git shows that dimension is outdated

IndrajeetPatil:
How come this was merged? I didn't revert the change to the submodule. 👀 

Thanks for the tip; I will start using `Cmder`.

msevestre:
Oh crap..I thought you had done it. Well this needs to be reverted. Sorry

PR #769, opened 2022-02-14
Fixes #767

### LINKED ISSUE #767: DataSet constructor
`DataSet` can get either a ref or a name to create an empty DS... which is just a bad signature because 

```
ds1 <- ospsuite::DataSet$new("ds1")

Error in private$.dataRepository$baseGrid : 
$ operator is invalid for atomic vectors
3.
private$.initializeCache() at data-set.R#202
2.
initialize(...)
1.
ospsuite::DataSet$new("ds1")
```

This happens because the first argument should be `ref` and the second is `name`. As the user would only use `name` as an argument when creating a DataSet from scratch, I would propose to either 

1) Swap the arguments in the constructor, such that `name` is the first and `ref` the second, so `DataSet$new("ds1")`, would work as expected, or
2) Only allow `ref` in the constructor and create a function for an empty data set, or a function `createDataSet(xValues, yvalues, yError)`.

I vote for Nr.1 as for me it is easier to have an overview of what I can do with the object in the object, and not to have to look through all available functions in the package.

@msevestre please give your feedback.

PavelBal:
DataSet constructor first argument is `name`, second `dataRepository`

### DIFF
--- a/R/data-set.R
+++ b/R/data-set.R
@@ -194,7 +194,7 @@ DataSet <- R6::R6Class(
     #' @param name Name of the `DataSet` if created from scratch (no `dataRepository`)
     #' provided. Ignored if `dataRepository` is not `NULL`.
     #' @return A new `DataSet` object.
-    initialize = function(dataRepository = NULL, name = NULL) {
+    initialize = function(name = NULL, dataRepository = NULL) {
       if (is.null(dataRepository) && is.null(name)) {
         stop(messages$errorDataSetNameMissing)
       }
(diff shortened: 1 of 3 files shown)


msevestre:
build is failing

PavelBal:
We urgently have to fix the issue with not being able to run check locally (PK-Sim issue)... this is pure pain.

msevestre:
What issue is this?

PavelBal:
Building running again.

PR #770, opened 2022-02-14
Create .gitattributes

msevestre:


msevestre:
@PavelBal @IndrajeetPatil Here we go...

PR #771, opened 2022-02-14
closes #768; re-export enum and helpers

### LINKED ISSUE #768: re-export enum
I don't know why this is not re-imported, but I think we should to this. Only problems with `enum` in third-party code since it has been moved to utils.

IndrajeetPatil:


msevestre:
Why are we doing this? I don't understand (e.g. lack of knowledge) why this is good to reexport from another package

IndrajeetPatil:
See https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/768#issuecomment-1039359477

PR #772, opened 2022-02-14
fixes normalized line ending

msevestre:


msevestre:
FYI
I ran the following command

git add --renormalize .

msevestre:
@PavelBal @IndrajeetPatil This should hopefully be it... we will see.

PR #777, opened 2022-02-18
WIP: Consolidate with utils

### LINKED ISSUE #714: Use `messages` from OSPSuite.RUtils
`messages` have been transferred to `OSPSuite.RUtils` but are still present in `ospsuite`, which is a duplication.

### LINKED ISSUE #730: getOSPSuiteSetting: enum instead of string?
At the moment `getOSPSuiteSetting` function takes a string as argument and it's not clear what are possible values for the `settingName`
The online help does not provide this information.
Should we define an **enum** for the `settingName `instead?

### LINKED ISSUE #774: Remove `Printable`
In favor of utils

PavelBal:
Fixes #774
Fixes #730
Fixes #714


R/ospsuite-env.R
```diff
@@ -23,10 +21,6 @@ ospsuiteEnv$containerTask <- NULL
 # Separator defined in OSPSuite.Core.
 ospsuiteEnv$pathSeparator <- "|"
 
-# Default values for the formatNumerics() helper function
```
PavelBal, line 26: Not used any more

R/utilities-pksim.R
```diff
@@ -53,7 +53,7 @@ initPKSim <- function(pksimFolderPath = NULL) {
     stop("Only Windows platforms are supported")
   }
 
-  suite.name <- ospsuiteEnv$suiteName
+  suite.name <- ospsuite.utils::getOSPSuiteUtilsSetting("suiteName")
```
PavelBal, line 56: I don't know, maybe move it back to this package..

PavelBal:
> Can I try something by directly pushing into this PR?

sure

IndrajeetPatil:
Unfortunately, downloading the GitHub package doesn't work because we seem to be running into [this](https://github.com/r-lib/remotes/issues/130) issue.

But the updated version of the utility package is [already on CRAN](https://cran.r-project.org/web/packages/ospsuite.utils/index.html), and the binaries should be built in a day or two, and then the AppVeyor check can be run again.

R/cache.R
```diff
@@ -77,7 +77,7 @@ Cache <- R6::R6Class(
       if (missing(value)) {
         ls(private$cachedObjects)
       } else {
-        stop(messages$errorPropertyReadOnly("keys"), call. = FALSE)
+        stop(ospsuite.utils::messages$errorPropertyReadOnly("keys"), call. = FALSE)
```
IndrajeetPatil, line 80: We had decided to not use the namespace operator for the utilities package to reduce verbosity and to slightly increase readability. This is reverting changes made in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/717.

R/utilities-units.R
```diff
@@ -42,8 +42,26 @@ hasUnit <- function(unit, dimension) {
...
+#' Check if quantity can be represented in the unit
+#'
+#' @param quantity `Quantity` object
+#' @param unit Unit name to check for
+#'
+#' @return
+#' If validations are successful, `NULL` is returned. Otherwise, error is
+#' signaled.
+#' @export
+validateHasUnit <- function(quantity, unit) {
```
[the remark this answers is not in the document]
IndrajeetPatil: We were not exporting this function, but we do now. I think we should make an entry for this new export in `NEWS.md`.

R/utilities-units.R
```diff
@@ -77,9 +95,9 @@ getBaseUnit <- function(dimension) {
 #' valuesInBaseUnit <- toBaseUnit(par, c(1000, 2000, 3000), "ml")
 #' @export
 toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL, molWeightUnit = NULL) {
-  validateIsOfType(quantityOrDimension, c(Quantity, "character"))
-  validateIsNumeric(values, nullAllowed = TRUE)
-  validateIsNumeric(molWeight, nullAllowed = TRUE)
+  ospsuite.utils::validateIsOfType(quantityOrDimension, c(Quantity, "character"))
```
IndrajeetPatil, line 98: Should this instead be `c("Quantity", "character")`?

R/ospsuite-env.R
```diff
@@ -60,6 +54,11 @@ ospsuiteEnv$isPKSimLoaded <- FALSE
 # NetTask `DimensionTask` cached for performance benefits. Created the first time it is requested.
 ospsuiteEnv$dimensionTask <- NULL
 
+#' Names of the settings stored in ospsuiteEnv. Can be used with `getOSPSuiteSetting()`
+#' @include utilities.R
+#' @export
+ospsuiteSettingNames <- enum(names(ospsuiteEnv))
```
IndrajeetPatil, line 60: Worth logging this new export in `NEWS.md`.

IndrajeetPatil (changes requested):

R/utilities-units.R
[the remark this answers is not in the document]
PavelBal: Actually I removed as export. We didn't export it so far, and unless there is a need for it, let's not export too much.

R/utilities-units.R
PavelBal: Using string names for all `isOfType` and `validateIsOfType` now.

R/ospsuite-env.R
PavelBal: Would add it when we have the separation between user-defined "global" settings and "read-only".

PR #780, opened 2022-02-21
aliases and deprecations for dataframe methods

### LINKED ISSUE #779: Conventions for naming dataframe methods
Is there some hidden logic to how we are naming the dataframe methods? 

Why do some methods have `To`, while others have `As` in their names?

``` r
library(ospsuite)

grep(
  "dataframe",
  getNamespaceExports("ospsuite"),
  ignore.case = TRUE,
  value = TRUE
)
#> [1] "dataSetToDataFrame"           "pkAnalysesAsDataFrame"       
#> [3] "simulationResultsToDataFrame" "populationAsDataFrame"
```

<sup>Created on 2022-02-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

IndrajeetPatil:
closes #779


R/zzz_deprecated.R
```diff
@@ -0,0 +1,26 @@
...
+#' @rdname ospsuite_deprecated
+#' @export
+pkAnalysesAsDataFrame <- function(...) {
+  .Deprecated("pkAnalysesAsDataFrame")
+  pkAnalysesToDataFrame(...)
+}
+
+#' @rdname ospsuite_deprecated
+#' @export
+populationAsDataFrame <- function(...) {
```
msevestre, line 23: Oh that's awesome. It's using the details from the comments at the top of the file?

msevestre (approved):

R/zzz_deprecated.R
IndrajeetPatil: Yes, and the docs just redirect the user to the docs for the primary function.

So:
- if someone uses `populationAsDataFrame(...)`, it will work but will issue a deprecation warning
- if someone searches for `?populationAsDataFrame`, it will take them to `ospsuite_depcreated` docs that detail which function they should instead be using

PR #781, opened 2022-02-21
issue warning when dataset name not found

### LINKED ISSUE #778: Issue warning when `$setGroups` has a non-existing dataset name
Currently, this is silently ignored, but better to issue a warning. This is most likely to happen when a dataset name is misspelled.

``` r
library(ospsuite)

# load the simulation
simulationResults <- runSimulations(simulations = sim)

# create a new dataset object
dataSet <- DataSet$new(name = "DS")

myCombDat$addSimulationResults(simulationResults)

myCombDat$setGroups(list(
  "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)" = "group1",
  "x" = "group2" # this doesn't exist and is silently ignored
))

myCombDat$groupMap
#> # A tibble: 1 x 3
#>   group  name                                                           dataType
#>   <chr>  <chr>                                                          <chr>   
#> 1 group1 Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral V~ simulat~
```

<sup>Created on 2022-02-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

IndrajeetPatil:
- closes #778
- adds an additional check to make sure that all dataset to group mappings are unique

### DIFF
--- a/tests/testthat/test-data-combined.R
+++ b/tests/testthat/test-data-combined.R
@@ -128,15 +128,61 @@ test_that("setting groups fails when group specification is invalid", {
   myCombDat <- DataCombined$new()
   myCombDat$addDataSets(dataSet)
 
-  expect_error(myCombDat$setGroups(c(2, 4)))
-  expect_error(myCombDat$setGroups(list("x" = 2, "y" = 4)))
+  expect_error(
+    myCombDat$setGroups(c(2, 4)),
+    "You need to provide a named list with at least one valid grouping."
+  )
+
+  expect_error(
+    myCombDat$setGroups(list("x" = 2, "y" = 4)),
+    "Names for groups can only be of `character` type."
+  )
 })
 
 test_that("setting groups fails when group specification is in a nested list", {
   myCombDat <- DataCombined$new()
   myCombDat$addDataSets(dataSet)
 
-  expect_error(myCombDat$setGroups(list("m" = list("x" = "2", "y" = "4"))))
+  expect_error(
+    myCombDat$setGroups(list("m" = list("x" = "2", "y" = "4"))),
+    "A nested list is not a valid argument here."
+  )
+})
+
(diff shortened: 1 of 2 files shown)


tests/testthat/test-data-combined.R
```diff
@@ -128,15 +128,61 @@ test_that("setting groups fails when group specification is invalid", {
   myCombDat <- DataCombined$new()
   myCombDat$addDataSets(dataSet)
 
-  expect_error(myCombDat$setGroups(c(2, 4)))
-  expect_error(myCombDat$setGroups(list("x" = 2, "y" = 4)))
+  expect_error(
+    myCombDat$setGroups(c(2, 4)),
+    "You need to provide a named list with at least one valid grouping."
```
msevestre, line 133: This is where using a constant file for strings would be much easier and future proof. Right now, we have "text" spread out everywhere. Changing a typo or modifying the text should (arguably) not break a test.

msevestre (approved):
Looks good to me
I would like to discuss following the idea of collating all strings back into one file instead.

tests/testthat/test-data-combined.R
IndrajeetPatil: Let's track this in https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/issues/76

PavelBal:
> I would like to discuss following the idea of collating all strings back into one file instead.

From my perspective there is nothing to discuss - this was always the paradigm and we didn't agree to change it.

tests/testthat/test-data-combined.R
msevestre: I don't think package specific messages should be in utils. They should be in the lowest package possible in the hierarchy

PR #784, opened 2022-02-24
783 update dimension and dll

### LINKED ISSUE #783: Update dimensions submodule to the latest version
is outdated

msevestre:
Fixes #783


msevestre:
FOr some reasons, the dlls were back to being the old ones again so I updated them at the same time. Probably due to one of our merge conflict lately

msevestre:
@PavelBal @IndrajeetPatil Dll should never be added by "hand" in the ext/lib folder. There is this Threading.Task.Extensions.dll that keeps coming back. It's not in the package.config nor does it need to be here in the first place I believe. Any idea where this is coming from?

IndrajeetPatil:
I have no idea what this `dll` is about.

Is this something related to your system, because the [history for this file](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/commits/develop/inst/lib/System.Threading.Tasks.Extensions.dll) shows that it was introduced in your commit on Jan 4?

appveyor.yml
```diff
@@ -35,7 +35,7 @@ before_build:
   - nuget sources add -name core -source https://ci.appveyor.com/nuget/ospsuite-core
   - nuget sources add -name centos -source https://ci.appveyor.com/nuget/ospsuite-centos
   - nuget install packages.config -OutputDirectory packages # cannot use automatic artifact publish because it's not recognized in a C++/CLI project
-  - rake "prepare_for_build[%APPVEYOR_BUILD_VERSION%, ckd-population-feature]"
+  - rake "prepare_for_build[%APPVEYOR_BUILD_VERSION%]"
```
msevestre, line 38: This I have changed twice. This is the 3rd time.

msevestre:
I do not trust the history on this repo anymore. I would **never** add a library by hand like this. 
And also I have changed the appveyor.yml about a week ago and it was building and it is back to what it was showing NO Changes in the history
So as long as we have a build green soon...we should be good. But this is painful

msevestre:
and now a test is failing with the importer.... I am going to cry

msevestre:
@IndrajeetPatil 
I am getting those errors locally
```
-- Failure (test-data-combined.R:80:3): data frame output is as expected when only `DataSet` is provided --
names(df) not equal to c(...).
Lengths differ: 17 is not 20

-- Failure (test-data-combined.R:325:3): data transformations as expected when both DataSet and SimulationResults provided --
dim(df) not equal to c(830L, 21L).
1/2 mismatches
[2] 18 - 21 == -3

-- Failure (test-data-combined.R:328:3): data transformations as expected when both DataSet and SimulationResults provided --
names(df) not equal to c(...).
Lengths differ: 18 is not 21

-- Failure (test-data-combined.R:883:3): sequential update when objects have different values --
dim(df1) not equal to c(77L, 20L).
1/2 mismatches
[2] 17 - 20 == -3

-- Failure (test-data-combined.R:884:3): sequential update when objects have different values --
dim(df2) not equal to c(76L, 20L).
1/2 mismatches
[2] 17 - 20 == -3

-- Failure (test-data-combined.R:885:3): sequential update when objects have different values --
dim(df1Filter) not equal to c(25L, 20L).
1/2 mismatches
[2] 17 - 20 == -3

-- Failure (test-data-combined.R:886:3): sequential update when objects have different values --
dim(df2Filter) not equal to c(24L, 20L).
1/2 mismatches
[2] 17 - 20 == -3
```

Is there a package that I need to update?

msevestre:
some more code was missing
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/commit/eca8bc2243a8101bc74b29d16f50893027c71a54

@IndrajeetPatil @Yuri05 @PavelBal 
It would be good to understand how we messed up like this. I have lost 2 hours trying to catch up on missing code that I committed weeks ago. (and it is still not running)

PavelBal:
@msevestre Remember we had this case where all my commits were lost and were not shown in the history? Absolutely no idea how that happened, I even had a thought that GitHub had some kind of a hick-up..

msevestre:
@IndrajeetPatil Same error on appveyor. Probably a fix was also lost here (or update to latest dll is creating an issue).
Any idea?

msevestre:
ALternatively the code in ospsuite.core now creates a different set of data.... @georgeDaskalakis @abdelr Is this possible?

IndrajeetPatil:
What do you see locally if you run the following code?

(N.B. You will need to source the helpers for tests first and run this in the `/testthat` folder.)

``` r
library(ospsuite)

sim <- loadTestSimulation("MinimalModel")
  simulation = sim,
  filePaths = getTestDataFilePath("Stevens_2012_placebo_indiv_results.csv")
)

dataSet <- loadDataSetsFromExcel(
  xlsFilePath = getTestDataFilePath("CompiledDataSetStevens2012.xlsx"),
  importerConfiguration = loadDataImporterConfiguration(getTestDataFilePath("ImporterConfiguration.xml"))
)


myCombDat$addDataSets(dataSet[[1]])

# checking dataframe methods
names(myCombDat$toDataFrame())
```

I see:

```r
 [1] "name"         "group"        "dataType"     "xValues"      "xUnit"        "xDimension"   "yValues"     
 [8] "yUnit"        "yDimension"   "yErrorValues" "yErrorType"   "yErrorUnit"   "molWeight"    "lloq"        
[15] "Source"       "Sheet"        "Organ"        "Compartment"  "Molecule"     "Group Id"
```

Three of these columns seem to be missing.

msevestre:
> names(myCombDat$toDataFrame())
 [1] "name"         "group"        "dataType"     "xValues"      "xUnit"        "xDimension"  
 [7] "yValues"      "yUnit"        "yDimension"   "yErrorValues" "yErrorType"   "yErrorUnit"  
[13] "molWeight"    "lloq"         "Source"       "Sheet"        "Organ"        "Compartment" 
[19] "Molecule"     "Group Id"

msevestre:
20 columns as well

IndrajeetPatil:
All failing tests seem to trace to the dataframe output for `DataSet` class objects. 

Was there any change recently to methods related to this class that result in fewer columns being returned?

PavelBal:
> Was there any change recently to methods related to this class that result in fewer columns being returned?

Yes! I realize it now, this is related to https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1429

The new version does not return meta data that is empty (almost all columns would be empty in this example)

IndrajeetPatil:
Now I am so confused.

For example, one of the failing tests is-

```r
-- Failure (test-data-combined.R:78:3): data frame output is as expected when only `DataSet` is provided --
dim(df) not equal to c(12L, 20L).
1/2 mismatches
[2] 17 - 20 == -3
```

This is saying that this dataframe has 17 columns when 20 were expected. 

> 20 columns as well

So if you locally see 20 columns as well, why are the tests failing on your local machine? Or are they no longer failing?

It was kind of reassuring to know that the same tests were failing for you locally and on `AppVeyor` because then there is a systemic issue. If tests pass locally on our machines but not on `AppVeyor`, then it's going to be a headache to figure out why.

msevestre:
ok so is it safe to update the column to the expected # ? e.g. 17 instead of 20 etc?

IndrajeetPatil:
> Yes! I realize it now, this is related to https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1429

Phew! So just the tests need to be changed in _ospsuite_ to accommodate for this new change.

IndrajeetPatil:
> ok so is it safe to update the column to the expected # ? e.g. 17 instead of 20 etc?

Yes! 

Also, in tests like the following, some column names will also have to be removed:

```r
  expect_equal(
    names(df),
    c(
      "name", "group", "dataType", "xValues", "xUnit", "xDimension",
      "yValues", "yUnit", "yDimension", "yErrorValues", "yErrorType",
      "yErrorUnit", "molWeight", "lloq", "Source", "Sheet", "Organ",
      "Compartment", "Molecule", "Group Id"
    )
  )
```

msevestre:
@IndrajeetPatil This is where targeted tests would have helped. Exactly the use case where we need to go into the code to understand why the test is failing.

msevestre:
also why is 17 the right value since your test returns 20 columns?

IndrajeetPatil:
> targeted tests would have helped

Why the conditional perfect tense? 

I thought it did help, no? It traced the problem to dataframe output for `DataSet` objects, which was a good enough start, I thought. 

I am not sure how the test could have helped narrow this even further to removal of metadata columns because in the dataframe there is no way to distinguish between columns that came from data and the ones that came from metadata.

PavelBal:
This is why the tests should not contain any hard-coded stuff. Deduce the names of the columns etc. directly from `DataSet` - i.e. following from the logic *what* the function does (creates columns for all meta data`), read the meta data from data set and check for these columns in the test.

msevestre:
A specific test checking that the returned data set has the expected number of columns (<=that being the name of test) would have been clearer to me. 

I have fixed the tests now. Indeed column organ, compartments and species are now gone. We might want to add a test in the future with actual meta data to ensure that this use case is also covered

msevestre:
Once this is merged (and green hopefully lol), I 'll update the website. Luckily we have all those tests here and an amazing coverage.

IndrajeetPatil:
> This is why the tests should not contain any hard-coded stuff. Deduce the names of the columns etc. directly from DataSet - i.e. following from the logic what the function does (creates columns for all meta data`), read the meta data from data set and check for these columns in the test.

Hard disagree. Imagine a scenario where the PR unintentionally removes more than meta-data columns and there are no tests to catch this in the dependency. 

Your tests would never catch this either because they are blindly following the imported function's behaviour; they will _always_ pass. You will need to wait for the users to discover this change and then figure out what led to this change. 

If this were a change to a function in `v10`, for example, we could even have logged that these 3 columns will not be returned by dataframe when there is no meta-data. The tests need to catch these kinds of changes, and that's why hard-coding some things is recommended IMHO.

msevestre:
>dim(df) not equal to c(12L, 20L).
1/2 mismatches
[2] 17 - 20 == -3

True. This is a very good help already.  I am just always interesting in reading the test "context" and expectations in plain english I guess

msevestre:
> why hard-coding some things is recommended IMHO.

Yep. You need to test the behavior here and therefore need to make sure that what you get is what you are expecting. I think the only thing that might be missing now is a test with actual meta data (e.g. 20 columns again) because we have lost this aspect I believe.

PR #802, opened 2022-03-04
Fixes #797

### LINKED ISSUE #797: Explicitely set `cloneable` for all classes
As the `cloneable`-inheritance is somhow random, we should set the value of the property in every class. By defaul, cloneable should be set to `FALSE` for all classes descending from `DotNetWrapper`.

Following classes should implement `cloneable = TRUE`:

PavelBal:
Explicitly set the `cloneable` property.


PavelBal:
image

What the? The file is not in the `develop` branch, howcan it be changed/deleted through the PR??

msevestre:
you probably have this file locally.

msevestre:
please remove the dll. I have been deleting that file 3 times already. dll should never be copied by hand

R/data-importer-configuration.R
```diff
@@ -6,7 +6,7 @@
 DataImporterConfiguration <- R6::R6Class(
   "DataImporterConfiguration",
   inherit = DotNetWrapper,
-  cloneable = TRUE,
+  cloneable = FALSE,
```
msevestre, line 9: Why are we doing this explicitley since the base DotNetWrapper is FALSE? I am confused

PavelBal:
I had it until this PR and I removed it in the branch. But this is what the change-section is showing - that the PR will remove the file from `develop`?

R/data-importer-configuration.R
PavelBal: > I am confused

So was I. But if we not specify `cloneable` explicitely, NOT the property of the parent class but `TRUE` is used by default. See here: 

https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/pull/82#discussion_r819474906

msevestre: ok I See. Thanks @PavelBal

IndrajeetPatil:
@msevestre Long and short of the R6 class cloning behavior [here](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/797#issuecomment-1059081163).

PR #806, opened 2022-03-07
Add a method to extract paired dataframe for "predicted versus observed" plot

### LINKED ISSUE #804: Add function to interpolate predicted data
As mentioned in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/674#issuecomment-987833373, `DataCombined` object needs to provide data for creating predicted versus observed plot. 

image

For such a plot, the data needs to be _paired_. That is, at a given time (`xValues`), there needs to be a _single_ pair of `yValues` for observed and simulated data.

The `$toObsVsPredDataFrame()` public method for `DataCombined` will provide such a dataframe.

IndrajeetPatil:
Closes #804 

This PR adds only the function needed to extract the dataframe for "predicted versus observed" plot. 

The next step will be implementing the plotting function, but that will be part of a different PR. The plotting function will take in a `DataCombined` object and internally extract this dataframe for plotting.

**TODO**:

- [ ] Decide which columns need to be retained in the returned dataframe
- [ ] Add `return` section to roxygen docs describing the returned columns
- [ ] Add tests once the output is finalized

Here is an example of what a working example might look like to extract the dataframe and then using it with `{tlf}`:

```r
library(ospsuite)
library(dplyr)

# import observed data (will return a list of DataSet objects)
dataSet <- loadDataSetsFromExcel(
  xlsFilePath = "CompiledDataSet.xlsx",
  importerConfiguration = loadDataImporterConfiguration("ImporterConfiguration.xml")
)

# simulation results
sim <- importResultsFromCSV(
  simulation = loadSimulation("Stevens_2012_placebo_indiv_results.pkml"),
  filePaths = "Stevens_2012_placebo_indiv_results.csv"
)

# create a `DataCombined` object and add observed and simulated data
myDataComb <- DataCombined$new()
myDataComb$addDataSets(dataSet)
myDataComb$addSimulationResults(sim,
  quantitiesOrPaths = list(
    "Organism|Lumen|Stomach|Metformin|Gastric retention",
    "Organism|Lumen|Stomach|Metformin|Gastric retention distal",
    "Organism|Lumen|Stomach|Metformin|Gastric retention proximal"
  )
)

# set groupings
myDataComb$setGroups(
  list(
    "Stevens_2012_placebo.Placebo_distal"                         = "Stevens 2012 solid distal",
    "Organism|Lumen|Stomach|Metformin|Gastric retention distal"   = "Stevens 2012 solid distal",
    "Stevens_2012_placebo.Placebo_proximal"                       = "Stevens 2012 solid proximal",
    "Organism|Lumen|Stomach|Metformin|Gastric retention proximal" = "Stevens 2012 solid proximal",
    "Stevens_2012_placebo.Placebo_total"                          = "Stevens 2012 solid total",
    "Organism|Lumen|Stomach|Metformin|Gastric retention"          = "Stevens 2012 solid total"
  )
)

# a dataframe in wide format with paired observations
df <- myDataComb$toObsVsPredDataFrame(threshold = 1)

# example code that illustrates how this dataframe can be immediately plugged into a plotting function
library(tlf)
useTheme(loadThemeFromJson("theme.json"))
setDefaultLegendPosition(LegendPositions$insideTopLeft)

plotObsVsPred(
  data = as.data.frame(df),
  dataMapping = ObsVsPredDataMapping$new(
    x = "yValuesObserved",
    y = "yValuesPredicted",
    color = "group",
    shape = "group"
  ),
  plotConfiguration = ObsVsPredPlotConfiguration$new(
    title = "Stevens 2012",
    xlabel = "Observed Values",
    ylabel = "Simulated Values"
  )
)
```

image


R/data-combined.R
```diff
@@ -346,6 +349,144 @@ DataCombined <- R6::R6Class(
...
+      # make an internal copy for this function scope.
+      df <- private$.dataCombined %>%
+        # Remove unnecessary columns
+        private$.cleanDF(.) %>%
+        # Datasets which are not part of any grouping can't be paired
+        dplyr::filter(!is.na(group))
+
+      if (is.null(df)) {
+        stop(
+          "There are currently no datasets. You can add them with `$addDataSets()` and/or `$addSimulationResults()` methods.",
```
msevestre, line 396: Should not we move all those strings into a constant file with messages as we do with everything else?

R/data-combined.R
```diff
@@ -346,6 +349,144 @@ DataCombined <- R6::R6Class(
...
+          call. = FALSE
+        )
+      }
+
+      if (length(unique(df$group)) == 0L) {
+        stop(
+          "There are currently no grouping specified. You can set them with `$setGroups()` method.",
+          call. = FALSE
```
msevestre, line 404: why do you always add the call. = FALSE? What does this mean?

R/data-combined.R
```diff
@@ -346,6 +349,144 @@ DataCombined <- R6::R6Class(
...
+      # observed versus simulated datasets need to be paired for plotting.
+      #
+      # Additionally, the used plotting library (`{tlf}`) also expects a data in
+      # the wide format.
+      #
+      # This entails that if both of these wide dataframes need to be joined at
+      # some point, they should also have unique names.
+      obsData <- df %>%
+        dplyr::filter(dataType == "observed") %>%
+        dplyr::rename_with(~ paste0(.x, "Observed"), -dplyr::matches("^group$"))
```
msevestre, line 433: I don't understand what is it you are doing here? You are adding Observed and Predicted to the column name to avoid duplication? 
Also why not Simulated (vs Predicted since the dataType is Simulated)

Yuri05:
My expectation is: if observed time value is not in the list of simulated time values, then the **observed time is kept** and **simulated value for this observed time is interpolated**.

E.g. if my simulated data is
time|value
-|-
0|0
2|4
4|8

and my observed data is 
time|value
-|-
1| 1.9
3| 6.1

then I would expect the outcome
time|predicted|observed
-|-|-
1|2|1.9
3|6|6.1

R/data-combined.R
IndrajeetPatil: > You are adding Observed and Predicted to the column name to avoid duplication?

It's to highlight which type of data the values are associated with. For example, both for observed and simulated data, the time values are in `xValues` column, and we need to distinguish them in a wide format dataframe by having two columns `xValuesObserved` and `xValuesPredicted`.

Here is a simple example:

``` r
library(tidyverse, warn.conflicts = FALSE)

# tidy ---------

df1 <- tibble(name = "a", group = "x")
df2 <- tibble(name = "b", group = "y")

dplyr::full_join(df1, df2, by = "group")
#> # A tibble: 2 x 3
#>   name.x group name.y
#>   <chr>  <chr> <chr> 
#> 1 a      x     <NA>  
#> 2 <NA>   y     b

# wide ---------

df1 <- dplyr::rename_with(df1, ~ paste0(.x, "Observed"), -dplyr::matches("^group$"))
df2 <- dplyr::rename_with(df2, ~ paste0(.x, "Predicted"), -dplyr::matches("^group$"))

dplyr::full_join(df1, df2, by = "group")
#> # A tibble: 2 x 3
#>   nameObserved group namePredicted
#>   <chr>        <chr> <chr>        
#> 1 a            x     <NA>         
#> 2 <NA>         y     b
```

<sup>Created on 2022-03-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

We need data to be in the wide format because that's the format _tlf_ function expects.

> Also why not Simulated (vs Predicted since the dataType is Simulated)

No strong opinions either way. I can change it to `Simulated` if so desired.

I chose Predicted only because the plotting function is going to be called `plotPredictedVersusObserved()`, and not `plotSimulatedVersusObserved()`. So thought better to be consistent with naming.

But I am totally flexible here.

R/data-combined.R
IndrajeetPatil: Yes, definitely! 

Should I move them to _ospsuite.utils_?

R/data-combined.R
IndrajeetPatil: It decides whether the call should become part of the error.

I always set it to `FALSE` because the error messages are cleaner that way.

``` r
divide1 <- function(x, y) {
  if (y == 0) {
    stop("Can't divide by zero.", call. = TRUE)
  }

  x / y
}

divide1(1, 0)
#> Error in divide1(1, 0): Can't divide by zero.

divide2 <- function(x, y) {
  if (y == 0) {
    stop("Can't divide by zero.", call. = FALSE)
  }

  x / y
}

divide2(1, 0)
#> Error: Can't divide by zero.
```

<sup>Created on 2022-03-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

We can create a helper in _ospsuite.utils_ to avoid this duplication:

```r
stopWithoutCall <- function(...) stop(..., call. = FALSE)
```

What do you think?

PavelBal: I prefer to have the call in the message, as it makes debugging easier. It does not matter in your example as the error is thrown by the direct call, but in case of an error being thrown from a function or an object's method, `call. = FALSE` is just not informative enough.

R/data-combined.R
PavelBal: > We need data to be in the wide format because that's the format tlf function expects.

I was initially against having this method public and would prefer that such a DF is created internally by the `plotXXXVsObserved`, but you argumented that the user might want to have this information independently of calling `plotXXXVsObserved` because he might want to create his own plot... But now you are tailoring the output exactly to the use with TLF within the function to be created, and not for the general purpose. And now we end up in two completely different output formats (`$toDataFrame` and `$toObsVsPredDataFrame`) but containing actually the same information. I would be (and am) very confused.

R/data-combined.R
IndrajeetPatil: Because, typically, the errors bubble up from somewhere deep in the code, and we can't expect the user to be familiar with our _implementation_. Our error messages tend to be verbose and insightful enough for them to know what went wrong. And, if the user is savvy enough, they can just call `traceback()` and it will show them where the error is originating from.

It might seem helpful in this toy example to show the call, but imagine the more realistic scenario. 

Let's say I run a function called `computeEllipsisCircumference()` and it errors with `#> Error in divide1(1, 0): Can't divide by zero`. 

As a user, how is `divide1(1, 0)` call useful?

PavelBal: > As a user, how is divide1(1, 0) call useful?

To be honest, at this state I am more concernd about me as the dev, and `divide1(1, 0)` would definitely help me to find out where I have to debug.

R/data-combined.R
IndrajeetPatil: > But now you are tailoring the output exactly to the use with TLF within the function to be created, and not for the general purpose.

Nope. You _need_ the data to be in the wide format _irrespective of_ what package you choose, doesn't matter if it's _tlf_, _ggplot2_, or _base_. Paired data that needs to be plotted as a scatterplot will always have to be in wide format. It will be confusing for it _not_ to be.

I am mentioning _tlf_ here only because that's the one we will be using. The story is going to be the same no matter what you use.

```r
# ggplot2

ggplot(df, aes(xValuesObserved, xValuesPredicted)) + 
  geom_point()

# base

plot(df$xValuesObserved, df$xValuesPredicted)
```

IndrajeetPatil:
> the observed time is kept and simulated value for this observed time is interpolated

This seems simple and straightforward with your toy data, but is going to be a nightmare with multiple groupings containing multiple datasets and with varying threshold producing unpredictable gaps in the selected time points. 

I will have a look day after tomorrow.

R/data-combined.R
IndrajeetPatil: As a developer, can't you just run `traceback()`?

PavelBal: I would like to have the function call in the error message.

PavelBal:
Let's create an issue where we can discuss the specification of this feature. It's intenden use, desired output, and details like how to treat missing x-values points. The implementation then shoul follow the specification. Otherwise we will have too many discussion and changing the code back and forth.

R/data-combined.R
IndrajeetPatil: Since Juri and you both seem to find the function call in error message useful, I have reverted this change.

The error messages will now show the function call.

# before

``` r
library(ospsuite)

myCombDat$toObsVsPredDataFrame()
#> Error: There are currently no datasets. You can add them with `$addDataSets()` and/or `$addSimulationResults()` methods.
```

<sup>Created on 2022-03-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

# after

``` r
library(ospsuite)

myCombDat$toObsVsPredDataFrame()
#> Error in myCombDat$toObsVsPredDataFrame(): There are currently no datasets. You can add them with `$addDataSets()` and/or `$addSimulationResults()` methods.
```

<sup>Created on 2022-03-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

Yuri05:
>Let's create an issue where we can discuss the specification of this feature. 

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/804 ? 😉

PavelBal:
> https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/804 ?

Yes, for example. Specification first, implementation then.

R/data-combined.R
msevestre: cool thanks

R/data-combined.R
msevestre: I don;t think so. Those messages should leave in the package are they are very specific to what ospsuite-r is doing

msevestre:
interpolation seems to be the obvious choice here as well. I do not think we need the threshold parameter at all but I am probably not seeing the big picture here

msevestre:
I am wondering if this method should be outside of the class as a helper. So publically available but as a service doing something with the data. Do we expect more of these transformations or not? It reminds me a bit of serialziation. It's not the concern of the class to be able to fit all possible formats. Let's have a philosophical discussion about this :)

R/data-combined.R
IndrajeetPatil: https://adv-r.hadley.nz/conditions.html#signalling-conditions

Screenshot 2022-03-24 at 23 33 22

Yuri05: I am just thinking about the situation where user reports a bug.
With `call=FALSE` we will not immediately get the call stack and will need to ask the user to rerun the workflow, call after the  `traceback()` and send us the error call stack.

PavelBal: I think we already agreed on having `.call = TRUE`.

IndrajeetPatil:
> Specification first, implementation then.

For future reference, this implementation was based on the following [specification](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/674#issuecomment-987833373):

Screenshot 2022-04-08 at 23 30 34

PR #808, opened 2022-03-08
Fixes #807 fails if unit is provided and path is wrong

### LINKED ISSUE #807: `setParameterValuesByPath` fails if the path cannot be found if a unit is provided, even if `stopIfNotFound = FALSE`
Test: 

```
  sim <- loadTestSimulation("S1", loadFromCache = TRUE)
  parameterPath <- "foo"
  setParameterValuesByPath(parameterPath, 100, sim, stopIfNotFound = FALSE, units = "l")

Error in rClr::clrCall(task, "DimensionNameByPath", simulation$ref, path) : 
Type: OSPSuite.Utility.Exceptions.OSPSuiteException
Message: Could not find quantity with path 'foo'.
Method: OSPSuite.Core.Domain.IQuantity singleQuantityByPath(OSPSuite.Core.Domain.IModelCoreSimulation, System.String)
Stack trace:
at OSPSuite.R.Services.ContainerTask.singleQuantityByPath(IModelCoreSimulation simulation, String path)
at OSPSuite.R.Services.ContainerTask.DimensionNameByPath(IModelCoreSimulation simulation, String path)
3.
rClr::clrCall(task, "DimensionNameByPath", simulation$ref, path) at utilities-quantity.R#158
2.
setQuantityValuesByPath(quantityPaths = parameterPaths, values = values, 
simulation = simulation, units = units, stopIfNotFound = stopIfNotFound) at utilities-parameter.R#128
1.
setParameterValuesByPath(parameterPath, 100, sim, stopIfNotFound = FALSE, 
units = "l")
```

msevestre:


tests/testthat/test-utilities-quantity.R
```diff
@@ -56,7 +56,7 @@ test_that("It returns null if the quantity by path does not exist and stopIfNotF
 })
 
 test_that("It throws an error when trying to retrieve a quantity by path that would result in multiple quantities", {
-  expect_that(getQuantity(toPathString(c("Organism", "Liver", "*")), sim), throws_error())
+  expect_error(getQuantity(toPathString(c("Organism", "Liver", "*")), sim))
```
IndrajeetPatil, line 59: I think we should also be checking the exact error thrown here using the `regexp` [argument](https://testthat.r-lib.org/reference/expect_error.html).

This way, we know that the expected error occurred.

tests/testthat/test-utilities-quantity.R
```diff
@@ -102,17 +102,21 @@ test_that("It can set multiple quantity values with units", {
 
 test_that("It throws an exception when setting values for a quantity that does not exist", {
   parameterPath <- "Organism|Liver|NOPE|Volume"
-  expect_that(setQuantityValuesByPath(parameterPath, 100, sim), throws_error())
+  expect_error(setQuantityValuesByPath(parameterPath, 100, sim))
+})
+
+test_that("It does not throw an exception when setting values for a quantity that does not exist with unit", {
+  parameterPath <- "Organism|Liver|NOPE|Volume"
+  expect_error(setQuantityValuesByPath(quantityPaths = parameterPath, values = 100, simulation = sim, units = "ml", stopIfNotFound = FALSE), NA)
```
IndrajeetPatil, line 110: See above. 

The same for the two other instances below of `expect_error()`.

R/utilities-simulation-results.R
```diff
@@ -27,8 +27,6 @@
 #' @param population population used to calculate the `simulationResults`
 #'   (optional). This is used only to add the population covariates to the
 #'   resulting data table.
-#' @param stopIfNotFound Boolean. If `TRUE` and no result exist for the given
```
IndrajeetPatil, line 30: DRY docs. Nice!

IndrajeetPatil (changes requested):

R/utilities-quantity.R
```diff
@@ -155,8 +155,11 @@ setQuantityValuesByPath <- function(quantityPaths, values, simulation, units = N
     path <- enc2utf8(quantityPaths[[i]])
     value <- values[[i]]
     if (!is.null(units)) {
-      dimension <- rClr::clrCall(task, "DimensionNameByPath", simulation$ref, path)
-      value <- toBaseUnit(quantityOrDimension = dimension, values = value, unit = units[[i]])
+      dimension <- rClr::clrCall(task, "DimensionNameByPath", simulation$ref, path, stopIfNotFound)
+      # Dimension ca be be empty if the path was not found and stopIfNotFound is false
+      if (dimension != "" || stopIfNotFound) {
```
PavelBal, line 160: I am still not quite sure I understand this :D

I think you should exit early, if dimension is `""`, than the path is definitely not found and no need to continue:

```
    if (!is.null(units)) {
      dimension <- rClr::clrCall(task, "DimensionNameByPath", simulation$ref, path, stopIfNotFound)
      # Dimension ca be be empty if the path was not found and stopIfNotFound is false
      if (dimension == "") {
        next
      }
      value <- toBaseUnit(quantityOrDimension = dimension, values = value, unit = units[[i]])
    }
```

tests/testthat/test-utilities-quantity.R
PavelBal: Yes, I agree, we should try to follow this practice. Sometimes the error is thrown and the test passes, but the error is actually somewhere else and we do not really test the case.

tests/testthat/test-utilities-quantity.R
PavelBal: Here we are actually testing that it does NOT throw an error :D

IndrajeetPatil: Ah, I think, then this should be `expect_output()`.

tests/testthat/test-utilities-quantity.R
PavelBal: I added the message

tests/testthat/test-utilities-quantity.R
PavelBal: Changed to expect_output()

tests/testthat/test-utilities-quantity.R
```diff
@@ -102,17 +103,41 @@ test_that("It can set multiple quantity values with units", {
...
 
 test_that("It throws an error when the number of quantity paths differs from the number units", {
   quantityPath1 <- "Organism|Liver|Intracellular|Volume"
   quantityPath2 <- "Organism|VenousBlood|Plasma|CYP3A4"
-  expect_that(setQuantityValuesByPath(c(quantityPath1, quantityPath2), c(40, 50), sim, units = "ml"), throws_error())
+  expect_error(setQuantityValuesByPath(quantityPaths = c(quantityPath1, quantityPath2), values = c(40, 50),simulation =  sim, units = "ml"),
+               regexp = messages$errorDifferentLength("quantityPaths, units"))
+})
+
+test_that("It throws an error when setting values for a set of quantities with units and one path does not exist", {
```
PavelBal, line 126: I added two tests for multiple paths.

PavelBal:
@msevestre Thanks for fixing it so quickly, amazing. I just have one suggestion to exit early when the dimension could not be retrieved. However, this does not throw a warning "path could not be found", this should be extended in .NET maybe? (showing a warning when entity cannot be found).

IndrajeetPatil (approved):

tests/testthat/test-utilities-quantity.R
msevestre: `expect_output` !! I was wondering how to test that there is NO Error lol. THANK YOU @IndrajeetPatil . A quick search in SO showed that I need to put NA at the end. But this looked very weird to me and I was about to write an expect_no_error method lol (which would be easier to understand than expect_output IMO)

tests/testthat/test-utilities-quantity.R
msevestre: Interesting. I see pros and cons. Let's discuss

R/utilities-quantity.R
msevestre: done

msevestre:
ok warning also when dimension or unit is not  found.
Once the build is running, I'll nerge

PR #817, opened 2022-03-10
Data combined comments cleanup

IndrajeetPatil:
Before I work on the plotting functionality, I first wanted to clean up some remaining mess in `DataCombined`:

- Some comments reiterated the code, which is a red flag. I had initially done this because I wasn't sure how familiar the team was with the pipe-based workflow. But I have revised these comments to be more abstract and general.
- Removed `call. = FALSE` in `stop()` statement to print error call.
- Adopts consistent parameters names across public and private methods to avoid any possible confusion. 
- Passive voice as much as possible.

### DIFF
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -158,7 +158,7 @@ DataCombined <- R6::R6Class(
           quantitiesOrPaths = quantitiesOrPaths,
           population        = population,
           individualIds     = individualIds,
-          names             = newNames
+          newNames          = newNames
         )
       )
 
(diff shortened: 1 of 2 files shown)


Yuri05 (approved):

R/data-combined.R
```diff
@@ -3,14 +3,13 @@
 #' @description
 #'
 #' A class for storing simulated and/or observed in a single dataframe, which
-#' can be further used for visualization methods.
+#' can be further used data wrangling or data visualization pipelines.
```
PavelBal, line 6: `for` or `in` missing?

PavelBal (approved):

R/data-combined.R
IndrajeetPatil: Good catch! Fixed.

PR #821, opened 2022-03-14
use the recent R major version

IndrajeetPatil:


appveyor.yml
```diff
@@ -20,7 +20,7 @@ environment:
   NOT_CRAN: true
   KEEP_VIGNETTES: true
   R_ARCH: x64
-  R_VERSION: "4.0.5"
+  R_VERSION: "4.1.0"
```
msevestre, line 23: @Yuri05 is 4.1 supported by the R validated environment?

msevestre (commented):
We can only move to 4.1 if this is validated

appveyor.yml
Yuri05: oh wait, it's appveyor, not the minimal package version.
When compiled for 4.0, the package still runs under 3.6
But if we compile under 4.1 - it must be tested if it runs under 3.6

IndrajeetPatil: LOL

We might as well roll back to [S language](https://en.wikipedia.org/wiki/S_(programming_language)). 🤣

IndrajeetPatil:
I am confused:
How is bumping R version for **AppVeyor** related to whether the package works for the `3.6` user or not?

IndrajeetPatil:
Or is this about AppVeyor building binaries after each commit, which might not be usable by `3.6` user?

If this is the reason, then I am not sure. But, if the compilation survived the transition in _major_ version (`3.x` to `4.x`), I doubt if a change in _minor_ version (`4.x` to `4.y`) would break it.

appveyor.yml
IndrajeetPatil: LMAO

PavelBal:
> Or is this about AppVeyor building binaries after each commit, which might not be usable by 3.6 user?

Yes

> If this is the reason, then I am not sure. But, if the compilation survived the transition in major version (3.x to 4.x), I doubt if a change in minor version (4.x to 4.y) would break it.

Better check

PR #834, opened 2022-03-20
Fixes #825 load pkml error

### LINKED ISSUE #825: `loadDataSetFromPKML` sometimes mixes yValues and error columns
It is possible to load a data repository from PKML and create a `DataSet` out of it:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/dec536de121f9a7e1c4ad519c97ee5239ad5935c/R/utilities-data-set.R#L39-L40

The colums of the `DataRepository` are then assigned to the columns of the `DataSet`. So far we were assuming that the y values are always in the first column of the `DataRepository` that is not a base grid:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/dec536de121f9a7e1c4ad519c97ee5239ad5935c/R/data-set.R#L370-L380

[data.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/files/8252806/data.zip)

This is the case for the file `obs_data.pkml`, but not for `ObsDataAciclovir_2.pkml`

@msevestre we need a more robust way to get the y-column from the `DataRepository`.

msevestre:


R/data-set.R
```diff
@@ -341,13 +341,9 @@ DataSet <- R6::R6Class(
       validateEnumValue(errorType, DataErrorType)
       column <- private$.yErrorColumn
       values <- private$.getColumnValues(column)
-
       dataInfo <- rClr::clrGet(column$ref, "DataInfo")
-      rClr::clrSet(
```
msevestre, line 346: I prefer defining a local variable to make the code more readable

R/data-set.R
```diff
@@ -368,12 +364,13 @@ DataSet <- R6::R6Class(
...
     .initializeCache = function() {
+      dataRepositoryTask <- getNetTask("DataRepositoryTask")
       private$.xColumn <- private$.dataRepository$baseGrid
-      # We assume for now that the first column not base grid in the data column
-      private$.yColumn <- private$.dataRepository$allButBaseGrid[[1]]
 
-      dataRepositoryTask <- getNetTask("DataRepositoryTask")
-      netYErrorColumn <- rClr::clrCall(dataRepositoryTask, "GetErrorColumn", private$.yColumn$ref)
+      netYColumn <- rClr::clrCall(dataRepositoryTask, "GetMeasurementColumn", private$.dataRepository$ref)
+      private$.yColumn <- DataColumn$new(netYColumn)
```
msevestre, line 371: The one issue here. The instance of DataColumn won't be in the cache in the DataRepository. I do not think that this is an issue as the ref are the same...

PavelBal: No, don't think it will be a problem.

PavelBal (approved):

PR #835, opened 2022-03-21
Adds a utility function for unit conversion

### LINKED ISSUE #822: How to make sure observed and simulated data have the same units?
In order to plot them together for a comparison, these datasets need to be using the same units. While experimenting with datasets to create plots, here is what I have seen.

Pseudocode to generate `DataCombined` object and a dataframe:

```r
myCombDat$addSimulationResults(simResults)
myCombDat$addDataSets(dataSet)

df <- myCombDat$toDataFrame()

obsData <- dplyr::filter(df, dataType == "observed")
simData <- dplyr::filter(df, dataType == "simulated")
```

# Possibility-1

```r
> unique(obsData$yErrorUnit)
[1] "%"
> unique(simData$yErrorUnit)
[1] NA
```

# Possibility-2

```r
> unique(obsData$yErrorUnit)
[1] "%"
> unique(simData$yErrorUnit)
[1] ""
```

So here are my questions:

- Should `$toDataFrame()` be returning a dataframes where the observed and simulated data are already in the same units? This way, every single plotting function wouldn't have to do these checks and carry out unit conversions.
- Are there other possibilities than the ones I have seen?
- What to do when unit column is `NA`, like in Possibility-1?
- Are there any helper functions in the package to take care of unit conversions instead of doing them manually?

IndrajeetPatil:
I am creating a new file for it, only temporarily. 

Once the entire plotting functionality has been implemented, this function (along with its cousins) will be moved to `utilities-plotting.R`.

Closes #822


PavelBal:
I do not think this function should be exported. I ok having it to avoid code repetition in plotting functions, but not exporting. We already export a bunch of generic functions for unit conversion, but this is a very specific case that only works with a certain data frame structure.

Without the export, also some of the validation steps can be skipped.

DESCRIPTION
```diff
@@ -91,6 +91,7 @@ Collate:
     'utilities-path.R'
     'standard-path.R'
     'table-formula.R'
+    'unitConverter.R'
```
IndrajeetPatil, line 94: I am creating a new file for it, only temporarily. 

Once the entire plotting functionality has been implemented, this function (along with its cousins) will be moved to `utilities-plotting.R`.

R/unitConverter.R
```diff
@@ -0,0 +1,103 @@
...
+#'   yValues = c(0.25, 45, 78), yUnit = c("", "%", "%"), yDimension = c("Fraction", "Fraction", "Fraction"),
+#'   molWeight = c(12.5, 10, 5)
+#' ))
+#'
+#' ospsuite:::.unitConverter(df)
+#' ospsuite:::.unitConverter(df, xUnit = ospUnits$Time$h)
+#' ospsuite:::.unitConverter(df, yUnit = ospUnits$Fraction$`%`)
+#' ospsuite:::.unitConverter(df, xUnit = ospUnits$Time$h, yUnit = ospUnits$Fraction$`%`)
+#'
+#' @keywords internal
```
IndrajeetPatil, line 38: This is an internal function, but never hurts to have it documented. No `.Rd` files will be created for it.

PavelBal: Documentation is always better than no documentation! :D

R/unitConverter.R
```diff
@@ -0,0 +1,103 @@
...
+#' unit was specified by the user.
+#'
+#' @keywords internal
+#' @noRd
+.selectDefaultTargetUnit <- function(sourceUnit) {
+  argName <- deparse(substitute(sourceUnit))
+  sourceUnitLevels <- unique(sourceUnit)
+
+  # `toUnit` can't carry out conversion if source unit is `NA`
+  if (any(is.na(sourceUnitLevels))) {
```
PavelBal, line 97: This should NEVER happen, this implies that something went terrbly  wrong. As the whole purpose of this function is to check for `NA`, I would rather use `sourceUnitLevels[[1]]` in the parent function and not define a new function for it.

R/unitConverter.R
```diff
@@ -0,0 +1,103 @@
...
+  targetYUnit <- yUnit %||% .selectDefaultTargetUnit(data$yUnit)
+
+  # *WARNING*: Do not change the order of two `mutate()` statements.
+  #
+  # Since the old `xUnit` and `yUnit` columns are need for unit conversion,
+  # those columns must be updated only *after* the conversions have taken place.
+  data %>%
+    dplyr::rowwise() %>% # transform values to common units
+    dplyr::mutate(
+      xValues = toUnit(
```
PavelBal, line 59: I am wondering.... Will it call `toUnit` for every single value? In that case it appears to me as super inefficient. Image a population simulan with thousands of values - will it make thousands of calls to `toUnit`? It would be much better to convert a whole array with only one call.

PavelBal (changes requested):

R/unitConverter.R
IndrajeetPatil: > In that case it appears to me as super inefficient

This is the case, but the problem is that `toUnit` has only single parameter vectorized, but not the others, and this doesn't play well with data frame workflows.

I can get rid of this efficiency, but it's comes at the cost of increased code complexity.

msevestre: What other parameters should be vectorized?

IndrajeetPatil:
I am marking this as a draft again because the changes that made the function more efficient are ugly AF and very difficult to reason about.  🤢

I need to think if there is a way to avoid this complexity by changing `toUnit()`. Will create a separate issue.

PavelBal:
You should pass the whole vectors of values to convert at once. Calling .NET method for each single value will produce a huge overhead.

Imagine a population simulation with 1000 individualls, and simulated time of 1000 minutes. In theory you could convert all points with only one call, but with this implementation it will result in massive 1000x1000 calls.

PavelBal:
I just counted how often `toUtils` has been called in the simple example:

```
# small data frame to illustrate the conversion
> df <- dplyr::tibble(
  xValues = c(15, 30, 60), xUnit = "min", xDimension = "Time",
  yValues = c(0.25, 45, 78), yUnit = c("", "%", "%"), yDimension = c("Fraction", "Fraction", "Fraction"),
  molWeight = c(12.5, 10, 5)
)

# default conversion -------------------

> dfConvert <- .unitConverter(df)
[1] "ping for values: 15"
[1] "ping for values: 30"
[1] "ping for values: 60"
[1] "ping for values: 0.25"
[1] "ping for values: 45"
[1] "ping for values: 78"
```

IndrajeetPatil:
Which commit are you working with? The latest one?

PavelBal:
It does not take the correct molecular weight for conversion:

```
##### Wrong mw!
yVals <- c(1, 2, 3)
mw <- c(10, 10, 10)
df <- dplyr::tibble(
  xValues = c(15, 30, 60), xUnit = "min", xDimension = "Time",
  yValues = yVals, yUnit = c("mol", "mol", "mol"), yDimension = c(ospDimensions$Amount, ospDimensions$Amount, ospDimensions$Amount),
  molWeight = mw
)

# default conversion -------------------

dfConvert <- .unitConverter(df, yUnit = "g")

expectedYVals <- toUnit(ospDimensions$Amount, values = yVals, targetUnit = "g", sourceUnit = "mol", molWeight = 10, molWeightUnit = ospUnits$`Molecular weight`$`g/mol`)
```

Results oof unitConverter:

```
> dfConvert$yValues
[1] 2 4 6
```

expected results:

```
> expectedYVals
[1] 10 20 30
```

IndrajeetPatil:
> It does not take the correct molecular weight for conversion

Thanks, should be fixed now. I have also added tests for this.

PavelBal:
It again converts every single value:

```
[1] 78
> df <- dplyr::tibble(
+     xValues = c(15, 30, 60), xUnit = "min", xDimension = "Time",
+     yValues = c(0.25, 45, 78), yUnit = c("", "%", "%"), yDimension = c("Fraction", "Fraction", "Fraction"),
+     molWeight = c(12.5, 10, 10)
+ )
> .unitConverter(df)
[1] "ping:"
[1] 15
[1] "ping:"
[1] 0.25
[1] "ping:"
[1] 30
[1] "ping:"
[1] 45
[1] "ping:"
[1] 60
[1] "ping:"
[1] 78
```

Why don't you just convert data set by data set?

IndrajeetPatil:
I have reverted the changes that led to regression. Now they should be as many calls to toUnit as there are unique source units.

``` r
df <- dplyr::tibble(
  xValues = c(15, 30, 60), xUnit = "min", xDimension = "Time",
  yValues = c(0.25, 45, 78), yUnit = c("", "%", "%"), yDimension = c("Fraction", "Fraction", "Fraction"),
  molWeight = c(10, 10, 10)
)

ospsuite:::.unitConverter(df)
#> [1] "ping for min"
#> [1] "ping for "
#> [1] "ping for %"
#> # A tibble: 3 x 7
#>   xUnit xDimension xValues yUnit yDimension molWeight yValues
#>   <chr> <chr>        <dbl> <chr> <chr>          <dbl>   <dbl>
#> 1 min   Time            15 ""    Fraction          10    0.25
#> 2 min   Time            30 ""    Fraction          10    0.45
#> 3 min   Time            60 ""    Fraction          10    0.78
```

<sup>Created on 2022-03-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

> Why don't you just convert data set by data set?

Because if there are (e.g.) 10 datasets and 2 source units, there will still be 10 calls to `toUnit` when we only need 2 calls.

My current approach achieves that.

PavelBal:
Seems to work now ;)

PavelBal (approved):

PR #846, opened 2022-03-25
Fixes #843 bug pk parameters calculations

### LINKED ISSUE #843: Bug in calculating PK parameters in populations
ReportingEngine seems to expect IndividualID from 0 to N (steps of 1). Using a population with for example IndividualID 3,7,8,20,78, etc. leads to calculations for IDs that are notr part of the population, i,e, 0,1,2,4,5,6 and so on with NaN values...

msevestre:


R/quantity-pk-parameter.R
```diff
@@ -12,7 +12,7 @@ QuantityPKParameter <- R6::R6Class("QuantityPKParameter",
   active = list(
     #' @field values All values for `quantityPath` and `name`
     values = function(value) {
-      private$wrapReadOnlyProperty("Values", value)
+      private$wrapReadOnlyProperty("ValuesAsArray", value)
```
msevestre, line 15: This is the breaking change introduced in Core

tests/data/pop_5_spared_id.csv
```diff
@@ -0,0 +1,8 @@
...
...|Large intestinal transit time [min],Organism|LargeIntestine|Volume [l],Organism|LargeIntestine|Specific blood flow rate [l/min/kg organ],Organism|Liver|Volume [l],Organism|Liver|Specific blood flow rate [l/min/kg organ],Organism|Lung|Volume [l],Organism|Lung|Fraction vascular,Organism|Muscle|Volume [l],Organism|Muscle|Vf (neutral lipid)-PT,Organism|Muscle|Vf (lipid),Organism|Muscle|Vf (phospholipid)-PT,Organism|Muscle|Vf (protein),Organism|Muscle|Vf (water),Organism|Muscle|Vf (water)-PT,Organism|Muscle|Fraction interstitial,Organism|Muscle|Vf (neutral lipid)-WS,Organism|Muscle|Vf (neutral lipid)-RR,"Organism|Muscle|Vf (neutral phospholipid, plasma)-WS",Organism|Muscle|Vf (neutral phospholipid)-RR,Organism|Muscle|Vf (extracellular water)-RR,Organism|Muscle|Vf (protein)-WS,Organism|Muscle|Vf (intracellular water)-RR,Organism|Muscle|Vf (water)-WS,Organism|Muscle|Specific blood flow rate [l/min/kg organ],Organism|Pancreas|Volume [l],Organism|Pancreas|Specific blood flow rate [l/min/kg organ],Organism|PortalVein|Volume [l],Organism|Skin|Volume [l],Organism|Skin|Specific blood flow rate [l/min/kg organ],Organism|Spleen|Volume [l],Organism|Spleen|Specific blood flow rate [l/min/kg organ]
```
msevestre, line 3: I have here a new pop with 5 entries  and various (sparse) ids (0,5,10,15,20)

tests/testthat/test-utilities-pk-analysis.R
```diff
@@ -23,6 +23,17 @@ test_that("It should return null when retrieving a pk parameter that does not ex
...
 
+test_that("it can return pk analysis for sparse population and only return ids defined in the population file", {
+  clearOutputs(sim)
+  addOutputs(outputs, sim)
+  population <- loadPopulation(getTestDataFilePath("pop_5_spared_id.csv"))
+  results <- runSimulation(sim, population = population)
+  pkAnalyses <- calculatePKAnalyses(results = results)
+  df <- pkAnalysesToDataFrame(pkAnalyses)
+  d <- dplyr::distinct(df, IndividualId)
+  expect_equal(nrow(d), 5)
```
msevestre, line 34: Checking that there are only 5 different values in the actual df. Before it would have been 21

msevestre (commented):
This should solve the issue

msevestre:
/cc @abdullahhamadeh @pchelle 
this should fix the issue we talked about this morning

IndrajeetPatil (approved):
LGTM!

tests/testthat/test-utilities-pk-analysis.R
IndrajeetPatil: I don't know the context, but do you think it makes sense for `pkAnalysesToDataFrame()` to return only distinct rows?

So instead of the user calling `dplyr::distinct()`, like here, the function would handle this internally.

PavelBal (approved):

Yuri05 (approved):

tests/testthat/test-utilities-pk-analysis.R
msevestre: No no. This is not what I am trying to do. This is a test and I want to verify that I only have 5 unique values in one column of the returned table. (This is the IndividualId) column. But the values can definitely repeat themselves, one for each pk parameter

PR #855, opened 2022-03-29
adopt new API for setting and removing groups

### LINKED ISSUE #845: Is it possible to hide those prints in the test env?
image

### LINKED ISSUE #853: Implement new API for methods to set and remove groups for `DataCombined` class
Action points from https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/827

The new API for `$setGroups()` will be:

To set groupings-

```r
$setGroups(names = c(...),    groups = c(...))
$setGroups(names = list(...), groups = list(...)) # also acceptable
``` 

To remove groupings-

```r
$setGroups(names = c(name1, name2),    groups = c(NA, NA))
$setGroups(names = list(name1, name2), groups = list(NA, NA)) # also acceptable
$setGroups(names = list(name1, name2), groups = list(NULL, NULL)) # also acceptable
``` 

and, alternatively,

```r
$removeGroupAssignment(names = c(...))
$removeGroupAssignment(names = list(...))
```

IndrajeetPatil:
- [x] Change API for `$setGroups()`
- [x] Add `$removeGroupAssignment()`
- [x] Change tests to adapt to the new API
- [x] Add tests for the new functionality
- [x] Get rid of printed messages in tests

Closes #853
Closes #845

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,7 +20,7 @@ Depends:
     rClr (>= 0.9.1)
 Imports: 
     dplyr (>= 1.0.0),
-    ospsuite.utils (>= 1.2.0),
+    ospsuite.utils (>= 1.2.0.9000),
     purrr,
     R6,
     readr,
--- a/R/utilities-data-combined.R
+++ b/R/utilities-data-combined.R
@@ -1,68 +1,39 @@
-#' Count number of objects
-#'
-#' @details
-#'
-#' If the argument is not a vector, unlike `length()`, this function will not
-#' count the number of named bindings in an environment, but only the number of
-#' instances of a class.
-#'
-#' For example, `length(mtcars)` will return 11, but `objCount(mtcars)` will
-#' return 1.
-#'
-#' @param x An object (an atomic vector, a list, or instance(s) of a class).
-#'
-#' @examples
-#'
-#' objCount(c(1, 2, 3)) # 3
-#' objCount(list("a", "b")) # 2
-#' objCount(mtcars) # 1
-#' @return Integer representing the count of objects.
-#'
-#' @keywords internal
-#' @noRd
-objCount <- function(x) {
-  # `is.vector()` can handle both atomic vectors and lists, i.e.
-  # both `is.vector(c(1, 2))` and `is.vector(list(1, 2))` will be `TRUE`
-  #
-  # For anything other than a vector, the object count should be 1
-  if (!is.vector(x)) {
-    return(1L)
-  }
-
-  return(length(x))
-}
-
 #'  Validate arguments provided as vectors
 #'
 #' @details
 #'
-#' Validation of arguments provided as a vector involves:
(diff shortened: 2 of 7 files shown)


DESCRIPTION
```diff
@@ -20,7 +20,7 @@ Depends:
     rClr (>= 0.9.1)
 Imports: 
     dplyr (>= 1.0.0),
-    ospsuite.utils (>= 1.2.0),
+    ospsuite.utils (>= 1.2.0.9000),
```
Yuri05, line 23: Why do we need to rely on the .9000 version of the package?

IndrajeetPatil: Because some functions (e.g. `objectCount()`) I am using are not present in 1.2.0 version.

Yuri05: Then I would suggest that we increment the **ospsuite.utils** minor (or major?) version and require 1.3 (or 2.0) as the minimal version.
What do you think?

IndrajeetPatil: Sure.

I am also happy to make a new CRAN release `1.3.0` today or tomorrow, and we can use that for this PR. 

`1.2.0.9000` is deliberate. 

[R package development idiom](https://r-pkgs.org/release.html#post-release) is `x.x.x` mark CRAN release and then `.xxxx` denote the development version. 

image

So by relying on that version is making it explicit that I am using a development version of the package and the user will also need to update to that development version if they wish to use _ospsuite_.

Yuri05: >I am also happy to make a new CRAN release 1.3.0 today or tomorrow

If the current ospsuite.utils version is stable and will not change soon - why not? :)
Otherwise, we could increment the version to 1.3 (and in this PR) on GitHub first and release on Cran later.
Just let me know soon, because the current Reporting Engine Builds do not run right now, so depending on if you make the Cran release right now or not, RE can either wait for it or must switch from the CRAN version to the latest GitHub-version of the ospsuite.utils

IndrajeetPatil: Once https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/pull/98 is merged, I will make a new CRAN release.

R/utilities-data-combined.R
```diff
@@ -1,69 +1,40 @@
...
+#' An atomic vector of desired data type.
 #'
 #' @examples
 #'
 #' cleanVectorArgs(list(1, 2, NA, NULL), 4L, "numeric")
 #' cleanVectorArgs(c(1, 2, NA, NA_complex), 4L, "numeric")
+#'
 #' @keywords internal
 #' @noRd
 cleanVectorArgs <- function(arg = NULL, expectedLength = NULL, type) {
```
PavelBal, line 29: Internal functions should start with a '.'.

PavelBal (changes requested):

PavelBal (approved):

PR #859, opened 2022-04-05
add `*ToTibble()` variants

### LINKED ISSUE #857: Revert Always return tibble for ToDataFrame functions #841
This is what I was talking about all the time - a tiblle is NOT a data frame! esqlabsR is now broken exactly because of this - some code there expected a data frame but the subsetting does not work any more because it is different between df and tibble.

This is a breaking change that is not justified.

IndrajeetPatil:
- [x] `simulationResultsToTibble()`
- [x] `dataSetToTibble()`
- [x] `pkAnalysesToTibble()`
- [x] `populationToTibble()`
- [x] tests for the same

Closes #857


R/utilities-data-set.R
```diff
@@ -77,27 +77,24 @@ dataSetToDataFrame <- function(dataSets) {
   dataSets <- c(dataSets)
   validateIsOfType(dataSets, "DataSet")
 
-  # styler: off
```
IndrajeetPatil, line 80: Like we had decided, no tags to turn off styler for any portion of the code.

R/utilities-data-set.R
```diff
@@ -109,9 +106,19 @@ dataSetToDataFrame <- function(dataSets) {
     df[[name]] <- col
   }
 
-  return(dplyr::as_tibble(df))
+  # consistently return a (classical) data frame
+  return(as.data.frame(df))
```
PavelBal, line 110: Why do you use `as.data.frame` here? Isn't `df` already created as a classical df?

R/utilities-pk-analysis.R
```diff
@@ -95,5 +95,16 @@ pkAnalysesToDataFrame <- function(pkAnalyses) {
     }
   )
 
-  return(dplyr::as_tibble(dataFrame))
+  # consistently return a (classical) data frame
+  return(as.data.frame(df))
```
PavelBal, line 99: Same question as for  `dataSetToDataFrame`.

R/utilities-data-set.R
IndrajeetPatil: Reverted.

R/utilities-pk-analysis.R
IndrajeetPatil: Reverted

R/utilities-pk-analysis.R
```diff
@@ -95,5 +95,16 @@ pkAnalysesToDataFrame <- function(pkAnalyses) {
     }
   )
 
-  return(dplyr::as_tibble(dataFrame))
+  # consistently return a (classical) data frame
+  return(df)
```
PavelBal, line 99: `df` is not present, here it is called `dataFrame`

IndrajeetPatil: Strange that tests didn't fail.

Renamed `dataFrame` to `df` to be consistent across functions.

PavelBal (approved):

Issue #723, opened 2022-01-26
Define a print method for `SimulationBatch` object?


IndrajeetPatil:
Currently, it just dumps everything in its environment to the console (which is `R6`'s print method):

```r
> simBatch
<SimulationBatch>
  Inherits from: <DotNetWrapper>
  Public:
    addRunValues: function (parameterValues = NULL, initialValues = NULL) 
    clone: function (deep = FALSE) 
    initialize: function (ref, simulation) 
    ref: cobjRef
    runValuesIds: active binding
    simulation: active binding
  Private:
    .simulation: Simulation, ObjectBase, DotNetWrapper, Printable, R6
    finalize: function () 
    printClass: function () 
    printLine: function (entry, value = NULL, addTab = TRUE) 
    readOnlyProperty: function (propertyName, value, returnValue) 
    throwPropertyIsReadonly: function (propertyName) 
    wrapExtensionMethod: function (typename, methodName, propertyName, value) 
    wrapExtensionMethodCached: function (typename, methodName, propertyName, cachedValue, value) 
    wrapIndexProperty: function (propertyName, value) 
    wrapIntegerProperty: function (propertyName, value) 
    wrapNullableIntegerProperty: function (propertyName, value) 
    wrapProperty: function (propertyName, value, shouldSetNull = TRUE) 
    wrapReadOnlyProperty: function (propertyName, value) 
    wrapVectorProperty: function (propertyNameSingular, propertyNamePlural, value, returnPropertyName)
```

PavelBal:
Related: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/645

Issue #645, opened 2021-10-13
Option to print parametersOrPaths of a simulationBatch 


abdullahhamadeh:
Helps with debugging.

eg if we have

```
simulationBatch <- createSimulationBatch(simulation = simulation, parametersOrPaths = c("first|path","second|path"))
```

then 'simulationBatch$paths' would output `c("first|path","second|path")`

PR #864, opened 2022-04-07
Fixes issue #723 (and #645)

hannaei:
Defined a print method for SimulationBatch object


R/simulation-batch.R
```diff
@@ -71,6 +71,25 @@ SimulationBatch <- R6::R6Class(
       batchRunValues <- SimulationBatchRunValues$new(parameterValues, initialValues)
       rClr::clrCall(self$ref, "AddSimulationBatchRunValues", batchRunValues$ref)
       return(batchRunValues$id)
+    },
+
+    #' @description
+    #' Print the object to the console
+    #' @param ... Rest arguments.
```
IndrajeetPatil, line 78:
```suggestion
    #' @param ... Rest of the arguments.
```

msevestre: isn't it called in fact Rest argument or (Rest parameter)? Or is it just in Javascript that it is called like this? Just curious

IndrajeetPatil: Not in R, no. 

Typically, it is documented as 

- for `S3` methods (e.g., see `?base::mean`, `?tidyr::extract`, etc.)

```r
... further arguments passed to or from other methods
```

- for regular functions

```r
... Additional arguments
```

No hard and fast rules here, of course. 

So, if `Rest arguments` is the convention you have adopted, ignore my suggestion, but it's definitely not idiomatic of R package docs.

R/simulation-batch.R
```diff
@@ -71,6 +71,25 @@ SimulationBatch <- R6::R6Class(
...
       return(batchRunValues$id)
+    },
+
+    #' @description
+    #' Print the object to the console
+    #' @param ... Additional arguments.
+    print = function(...) {
+      simulationBatchOptions <- rClr::clrGet(self$ref, "SimulationBatchOptions")
+      private$printClass()
+      writeLines(paste0("   ", capture.output(self$simulation)))
```
msevestre, line 82: What does `capture.output` do?

hannaei: It sends output to a character string.

msevestre: What does that mean ? The result of print is concatenated to a string? Wouldn't printing the name of the simulation be enough?

hannaei: Yes, it looks like this:
```
SimulationBatch: 
   Simulation: 
      Name: Simple 
      Source file: D:/github/OSPSuite-R/tests/data/simple.pkml 
   runValuesIds: 
   Parameters: Organism|Liver|Volume R1|k1 
   Molecules: Organism|Liver|A 
```
But we could only have the name printed, sure.

msevestre: yeah. Let's remove the file name.  We do not need this I would think. Plus if we ever add more info in the simulation print., this is going to pollute this print

msevestre:
#645 is probably not fixed automatically because the message just says and #645 instead of and fixes #645

PR #868, opened 2022-04-10
use expect_error instead of throws

IndrajeetPatil:
A single line of code does the trick!

```r
xfun::gsub_dir(
  "(expect_that\\()(.*)(, throws_error\\(\\))",
  "\\expect_error(\\2",
  ext = "R",
  dir = "tests/"
)
```

### DIFF
--- a/tests/testthat/test-container.R
+++ b/tests/testthat/test-container.R
@@ -30,19 +30,19 @@ test_that("It can retrieve the fullPath of a container", {
 
 test_that("It throws an error when trying to set the path of a container", {
   container <- getContainer(liverPath, sim)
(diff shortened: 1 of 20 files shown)


msevestre:
Very satisfying 😍... even if you spend 30min making it work

IndrajeetPatil:
> even if you spend 30min making it work

I **wish** I had spent only 30 min! 😁

> _Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems._
>   Jamie Zawinski

But deeply satisfying when it worked!

msevestre:
First I wrote even if you spent "3hours"

PR #869, opened 2022-04-10
exclude zzz from code coverage computation

IndrajeetPatil:
Using these tags would work with all CI, and not just with AppVeyor, and it's future-proof. 
Plus, the tags are next to the actual code which can't be tested, and this is much more informative for someone wondering where are tests for this code.

Alternative to this:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/c31ecc2bc2b7668de0dc4d00d6beb0f579699d06/appveyor.yml#L54-L55

I am not sure why we are excluding print functions from code coverage. They can be easily tested using snapshots.
cf. https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/blob/develop/tests/testthat/test-Printable.R

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -1,6 +1,8 @@
+# nocov start
 .onLoad <- function(...) {
 
-  # Only for x64 bits packages. This is required to avoid error when package is being checked on CI for x86
+  # Only for x64 bits packages.
+  # This is required to avoid error when package is being checked on CI for x86
   is64 <- (.Machine$sizeof.pointer == 8)
   if (!is64) {
     return()
@@ -8,3 +10,4 @@
 
   initPackage()
 }
+# nocov end
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -52,7 +52,7 @@ before_test:
   - pandoc -v
 
 on_success:
-  - Rscript -e "covr::codecov(function_exclusions = c('print.*', '\\.onLoad'))"
+  - Rscript -e "covr::codecov(function_exclusions = c('print.*'))"
 
 test_script:
   - travis-tool.sh run_tests


msevestre:
>They can be easily tested using snapshots.

You are not testing anything here. You are just ensuring that it does not crash and that the version from previous is the same as now.  **You are not validating that the output is as expected**. 

Snapshot testing is very controversial in the unit testing world and the trend is now, at least in languages where Snapshot testing was created in the first place, to NOT use snapshot tests anymore because they most often than not,  lead to test being very britle
Let say that we want add a property to the print of the simulation, the test would actually fail! Why, because the snapshot is outdated. 

I am at the moment not convince that those tests are good. Please convince me otherwise :)

msevestre:
>- Rscript -e "covr::codecov(function_exclusions = c('print.*', '\\.onLoad'))" 

we should remove \\.onload from the appveyor file probably?

IndrajeetPatil:
> Let say that we want to add a property to the print of the simulation, the test would actually fail! Why, because the snapshot is outdated.

Yes! That's the point. You changed the behaviour of your function, so the expected side effects should catch up to that. 
The failing test is a crucial reminder for you to update the snapshot **if** the change is **deliberate**. 

This is also the same strategy used for checking visual outputs from plotting functions.

If not using snapshots, how _does_ one test function side effects? I don't see how a valid alternative to _fragile_ testing is _no_ testing at all.

msevestre:
Testing is about behavior for me. The behavior is correct before and after. Getting a failing test for a correct behavior is IMO noise

Print testing can be easily done by just calling the print method as we do in tests. The gain is terms of testing is exactly the same, but we do not have the noise coming from false negative

IndrajeetPatil:
> Getting a failing test for a correct behavior 

Yes, but if your behaviour changed, you also need to update the snapshot and only then you have established that this is the new correct behaviour. Until then, the test **must** fail because the testing framework doesn't _know_ that you changed your mind about what constitutes correct behaviour. 

> Print testing can be easily done by just calling the print method as we do in tests. 

How are you going to do that for printing plots? 
If you are just going to print them to a graphics device, how do you make sure that the outputs are **correct**? 
Human eye is terrible at this. Anything over a dozen points, and you won't be able to tell the difference, and try doing this for plot after plot after plot.

IndrajeetPatil:
> Snapshot testing is very controversial in the unit testing world and the trend is now, at least in languages where Snapshot testing was created in the first place, to NOT use snapshot tests anymore because they most often than not, lead to tests being very britle

Pragmatically speaking, this is irrelevant and uninteresting to me. My experience with them has been in the form of managing two massive visualization libraries in the R ecosystem, and I wouldn't have it any other way. I don't care if React community no longer uses them; it is the de facto standard for testing visualization libraries in R. Hell, [even ggplot2](https://github.com/tidyverse/ggplot2/tree/main/tests/testthat/_snaps) uses snapshot testing to make sure plots are correct. So the onus is on you to demonstrate why the R community is in the wrong to have adopted a standard like this.

Also, I am happy to stop suggesting using snapshot testing, but then whatever alternative you suggest should guarantee that accidental regressions like [this](https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/220) can never get past testing. Because, with snapshot testing, I can _guarantee_ that this kind of regression will never take place accidentally.

msevestre:
This is not a matter of right or wrong. As far as I know, ospsuite-r is not a visualization library. What I am trying to bring to the table here is that Snapshot testing has been overused (also by me) and people are now moving away from this, because more often, there are better ways of testing it.

In our case, when dealing with visualization, such as for tlf or the new integration with tlf, **I do not see any other choices** and I think they are valuable tools. 

Snapshot testing should, in my opinion, only be used at the very last option in our testing tool belt. 
Note also that I am much more interested in testing algorithms and calculations etc... This is why I mentioned that Snapshot testing for methods like print, which are only and purely side-effects, without any impact whatsoever on the goodness of what the package is doing, is probably not required unless the goal is to have 100% test coverage which is also a goal that does not make sense.  

Happy to discuss it further live.

IndrajeetPatil:
> As far as I know, ospsuite-r is not a visualization library.

Well, it's kind of becoming one now. We already have our first [plotting function](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/R/plotGrid.R), and it already uses [snapshot tests.](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/tree/develop/tests/testthat/_snaps/plot-grid) 🙃

> What I am trying to bring to the table here is that Snapshot testing has been overused (also by me) and people are now moving away from this, because more often, there are better ways of testing it.

Duly noted! 

If you have any reading material on alternatives to snapshot testing, I'll be happy to give it a read! I am curious to see what alternative are out there. I definitely haven't found anything in the R ecosystem for testing visualization, at least not that is developer-friendly and maintainable.

PR #876, opened 2022-04-13
unit converter can deal with missing values

### LINKED ISSUE #875: Make sure `.unitConverter()` doesn't fail in presence of missing values
This happens in combined data frame where observed data has `NA`s in the `molWeight` column. Unit conversion fails in such cases because `split()` creates empty data frames and passes them for conversion.

IndrajeetPatil:
closes #875

### DIFF
--- a/R/unitConverter.R
+++ b/R/unitConverter.R
@@ -51,18 +51,49 @@
 
   # xUnit --------------------------
 
+  # Calls to `toUnit` should equal the number of unique source units.
+  #
+  # So create a separate data frame for each source unit, make sure that the
+  # data frame is not empty, and then carry out unit conversion.
   xDataList <- split(data, data$xUnit)
+
+  # Re-combine data frames for each converted source unit.
   data <- dplyr::bind_rows(lapply(xDataList, .xUnitConverter, xTargetUnit))
 
   # yUnit ----------------
 
-  yDataList <- split(data, list(data$yUnit, data$molWeight))
+  # Same logic as `xUnit`. The only additional wrinkle is that, in addition to
+  # source units, molecular weight can also be different.
+  #
+  # So create a separate data frame for each source unit and molecular weight
+  # combination present, make sure that the data frame is not empty, and then
+  # carry out unit conversion.
+  #
+  # If molecular weights are missing, don't use it for splitting to prevent
+  # creating empty data frames in the list.
+  if (any(is.na(data$molWeight))) {
+    yDataList <- split(data, list(data$yUnit))
+  } else {
+    yDataList <- split(data, list(data$yUnit, data$molWeight))
+  }
+
   data <- dplyr::bind_rows(lapply(yDataList, .yUnitConverter, yTargetUnit))
 
   # yUnit error ----------------
 
+  # Follows the same logic as `yUnit`.
   if ("yErrorValues" %in% names(data)) {
-    yErrorDataList <- split(data, list(data$yErrorUnit, data$molWeight))
+    # Because `split()` will drop missing values from splitting variable, this
+    # can lead to loss of data. In such case, a splitting variable which doesn't
+    # have missing values needs to be used, which in this case is `yUnit.`
+    if (any(is.na(data$molWeight)) && any(is.na(data$yErrorUnit))) {
+      yErrorDataList <- split(data, list(data$yUnit))
+    } else if (any(is.na(data$molWeight)) && !any(is.na(data$yErrorUnit))) {
+      yErrorDataList <- split(data, list(data$yErrorUnit))
+    } else {
+      yErrorDataList <- split(data, list(data$yErrorUnit, data$molWeight))
+    }
+
     data <- dplyr::bind_rows(lapply(yErrorDataList, .yErrorUnitConverter, yTargetUnit))
   }
 
@@ -72,6 +103,11 @@
 #' @keywords internal
 #' @noRd
 .xUnitConverter <- function(xData, xTargetUnit) {
+  # no unit conversion possible if source unit is missing
+  if (is.na(xData$xUnit[[1]])) {
+    return(xData)
+  }
+
   xData$xValues <- ospsuite::toUnit(
     quantityOrDimension = xData$xDimension[[1]],
     values = xData$xValues,
@@ -87,6 +123,11 @@
 #' @keywords internal
 #' @noRd
 .yUnitConverter <- function(yData, yTargetUnit) {
+  # no unit conversion possible if source unit is missing
+  if (is.na(yData$yUnit[[1]])) {
+    return(yData)
+  }
+
   yData$yValues <- toUnit(
     quantityOrDimension = yData$yDimension[[1]],
     values = yData$yValues,
@@ -104,6 +145,11 @@
 #' @keywords internal
 #' @noRd
 .yErrorUnitConverter <- function(yData, yTargetUnit) {
+  # no unit conversion possible if source unit is missing
+  if (is.na(yData$yErrorUnit[[1]])) {
+    return(yData)
+  }
+
   yData$yErrorValues <- toUnit(
     quantityOrDimension = yData$yDimension[[1]],
     values = yData$yErrorValues,
--- a/tests/testthat/test-unitConverter.R
+++ b/tests/testthat/test-unitConverter.R
@@ -160,3 +160,80 @@ test_that("Correct conversion for yValues having the same unit but different MW"
 
   expect_equal(dfConvert$yValues, c(10, 10, 20))
 })
+
+# missing values handled properly -------------------
+
+dfNA <- dplyr::tibble(
+  dataType = c(rep("simulated", 3), rep("observed", 3)),
+  xValues = c(0, 14.482, 28.965, 0, 1, 2),
+  xUnit = "min", xDimension = "Time",
+  yValues = c(25.579, 32.446, 32.103, 0, 0.995, 0.991),
+  yUnit = c("%", "%", "%", "", "", ""),
+  yDimension = "Fraction",
+  yErrorValues = c(2.747, 2.918, 2.746, NA, NA, NA),
+  yErrorUnit = c("%", "%", "%", NA, NA, NA),
+  molWeight = c(NA, NA, NA, 129.1636, 129.1636, 129.1636)
+)
+
+dfNAConvert <- .unitConverter(dfNA)
+dfNAXYConvert <- .unitConverter(dfNA, xUnit = "h", yUnit = "")
+
+test_that("simulated data is as expected in presence of missing values - default units", {
+  # because of the missing values, no conversion should have taken place and so these two
+  # data frames should be identical
+  expect_equal(
+    dplyr::filter(dfNAConvert, dataType == "simulated"),
+    dplyr::filter(dfNA, dataType == "simulated")
+  )
+})
+
+test_that("observed data is as expected in presence of missing values - default units", {
+  # only `yValues` should change; everything else should remain the same
+  expect_equal(
+    dplyr::filter(dfNAConvert, dataType == "observed"),
+    dplyr::filter(dfNA, dataType == "observed") %>%
+      dplyr::mutate(yValues = 100 * yValues, yUnit = "%")
+  )
+})
+
+test_that("data have expected units in presence of missing values - custom units", {
+  expect_equal(unique(dfNAXYConvert$xUnit), "h")
+  expect_equal(unique(dfNAXYConvert$yUnit), "")
+})
+
+test_that("data have expected dimensions in presence of missing values - custom units", {
+  expect_equal(unique(dfNAXYConvert$xDimension), "Time")
+  expect_equal(unique(dfNAXYConvert$yDimension), "Fraction")
+})
+
+test_that("simulated xValues are converted as expected in presence of missing values - custom units", {
+  expect_equal(
+    dfNAXYConvert[dfNAXYConvert$dataType == "simulated", ]$xValues,
+    c(0.0000000, 0.2413667, 0.4827500),
+    tolerance = 0.001
+  )
+})
+
+test_that("observed xValues are converted as expected in presence of missing values - custom units", {
+  expect_equal(
+    dfNAXYConvert[dfNAXYConvert$dataType == "observed", ]$xValues,
+    c(0, 0.0166667, 0.0333333),
+    tolerance = 0.001
+  )
+})
+
+test_that("simulated yValues are converted as expected in presence of missing values - custom units", {
+  expect_equal(
+    dfNAXYConvert[dfNAXYConvert$dataType == "simulated", ]$yValues,
+    c(0.25579, 0.32446, 0.32103),
+    tolerance = 0.001
+  )
+})
+
+test_that("observed yValues are converted as expected in presence of missing values - custom units", {
+  expect_equal(
+    dfNAXYConvert[dfNAXYConvert$dataType == "observed", ]$yValues,
+    c(0, 0.995, 0.991),
+    tolerance = 0.001
+  )
+})


PavelBal:
> I am wondering if instead of having in it its own file, this function should instead be moved to [`utilities-units.R`](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/R/utilities-units.R) file.

Agree.

IndrajeetPatil:
All set here.

IndrajeetPatil:
> Unit conversion is broken again.

When was it broken before?

IndrajeetPatil:
> the second value should not even be transformed (toUnit would throw an error because no MW is provided)

What should happen when MW is missing? Should `.unitConverter` just produce an error and quit? If yes, that will simplify things a lot.

PavelBal:
Yes, actually you can just pass it to 'toUnit' and it will produce an error.

Indrajeet Patil ***@***.***> schrieb am Mi., 20. Apr. 2022,
09:14:

IndrajeetPatil:
But, if I am not mistaken, this means that plotting functions will _always_ fail if both simulated and observed datasets are present in the `DataCombined`, right?

All the examples I have seen in the package seem to have `NA`s in `molWeight` column for the observed data. For example,

```r
> setwd("~/GitHub/OSPSuite-R/tests/testthat")
>   file <- getTestDataFilePath("obs_data.pkml")
>   dataSet <- loadDataSetFromPKML(file)
> dataSetToTibble(dataSet)
# A tibble: 7 x 17
  name         xValues yValues yErrorValues xDimension xUnit yDimension yUnit yErrorType yErrorUnit molWeight  lloq Source Sheet
  <chr>          <dbl>   <dbl>        <dbl> <chr>      <chr> <chr>      <chr> <chr>      <chr>          <dbl> <dbl> <chr>  <chr>
1 ObservedData    1.80 1.01e-3         1.15 Time       min   Concentra~ mg/l  Arithmeti~ mg/l              NA    NA "C:\\~ Tabe~
2 ObservedData    4.87 8.30e-4         1.09 Time       min   Concentra~ mg/l  Arithmeti~ mg/l              NA    NA "C:\\~ Tabe~
3 ObservedData   10.2  7.30e-4         1.13 Time       min   Concentra~ mg/l  Arithmeti~ mg/l              NA    NA "C:\\~ Tabe~
4 ObservedData   30    2.80e-4         1.22 Time       min   Concentra~ mg/l  Arithmeti~ mg/l              NA    NA "C:\\~ Tabe~
5 ObservedData   60    1.20e-4         1.41 Time       min   Concentra~ mg/l  Arithmeti~ mg/l              NA    NA "C:\\~ Tabe~
6 ObservedData  120    3.05e-5         1.18 Time       min   Concentra~ mg/l  Arithmeti~ mg/l              NA    NA "C:\\~ Tabe~
7 ObservedData  240    5.93e-6         1.48 Time       min   Concentra~ mg/l  Arithmeti~ mg/l              NA    NA "C:\\~ Tabe~
# ... with 3 more variables: DoubleValue <dbl>, StringValues <chr>, IntegerValue <int>
```

This means that `DataCombined$toDataFrame()` will also produce a data frame with `NA`s in `molWeight` if observed dataset is present. This data frame will be passed to all plotting functions.

If `.unitConverter()` produces an error for _any_ missing MW, then that means plotting functions also fail because all plotting functions are going to have unit conversion as the first step to get the data in required/common units. 

This won't be desirable default behaviour for the plotting functions, IMO. I also don't know how to deal with missing MW here. The current approach was to return the rows where MW is missing untouched, but I guess that is also misleading.

How to proceed here?

PavelBal:
MW is only required for the conversion between certain dimensions (`Amount` and `Mass` and `Concentration (molar)` and `Concentration (mass)`). Conversion between any other dimension (or between units within one dimension) does not rely on MW. The `toUnit` method takes care of all the checks, so if you pass `NA` as MW it will fail with the appropriate message if MW is required.

IndrajeetPatil:
Thanks! That's handy to know.

Now the function will fail if MW is missing and dimensions is one of the ones you mentioned:

``` r
library(ospsuite)

dfNA <- dplyr::tibble(
  dataType = c(rep("simulated", 3), rep("observed", 3)),
  xValues = c(0, 14.482, 28.965, 0, 1, 2),
  xUnit = "min", xDimension = "Time",
  yValues = c(1, 1, 1, 1, 1, 1),
  yUnit = c("mol", "mol", "mol", "g", "g", "g"),
  yDimension = c("Amount", "Amount", "Amount", "Mass", "Mass", "Mass"),
  yErrorValues = c(2.747, 2.918, 2.746, NA, NA, NA),
  yErrorUnit = c("mol", "mol", "mol", "g", "g", "g"),
  molWeight = c(10, 10, 20, 20, NA, 10)
)

ospsuite:::.unitConverter(dfNA)
#> Error in rClr::clrCall(dimensionTask, "ConvertToUnit", dimension, targetUnit, : Type:    OSPSuite.Core.Domain.UnitSystem.UnableToResolveParametersException
#> Message: Unable to convert to mol.
#> Molecular Weight not available.
#> Method:  Double BaseUnitValueToUnitValue(OSPSuite.Core.Domain.UnitSystem.Unit, Double)
#> Stack trace:
#>    at OSPSuite.Core.Domain.UnitSystem.MergedDimensionFor`1.BaseUnitValueToUnitValue(Unit unit, Double valueInBaseUnit)
#>    at OSPSuite.Core.Domain.UnitSystem.DimensionExtensions.<>c__DisplayClass4_0.<BaseUnitValuesToUnitValues>b__0(Double value)
#>    at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
#>    at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
#>    at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
#>    at OSPSuite.Core.Domain.UnitSystem.DimensionExtensions.BaseUnitValuesToUnitValues(IDimension dimension, Unit unit, IEnumerable`1 values)
#>    at OSPSuite.R.Services.DimensionTask.convertToUnit(IDimension dimension, String targetUnit, Nullable`1 molWeight, Double[] valuesInBaseUnit)
#>    at OSPSuite.R.Services.DimensionTa
```

<sup>Created on 2022-04-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

R/utilities-units.R
```diff
@@ -336,3 +336,200 @@ initializeDimensionAndUnitLists <- function() {
...
+
+  return(data)
+}
+
+#' Remove empty data frames sometimes produced due to the non-existent
+#' combination of source unit and molecular unit.
+#'
+#' @keywords internal
+#' @noRd
+.removeEmptyDataFrame <- function(x) {
```
msevestre, line 483: I don't like this method because it seems to have side effects and also returns something
I cannot tell if we are modifying x in place or not. 
Also what is x?

R/utilities-units.R
```diff
@@ -336,3 +336,200 @@ initializeDimensionAndUnitLists <- function() {
...
+  )
+
+  yData$yUnit <- yTargetUnit
+
+  return(yData)
+}
+
+#' @keywords internal
+#' @noRd
+.yErrorUnitConverter <- function(yData, yTargetUnit) {
```
msevestre, line 522: In general, method with side effects should not return anything, unless this is used for method chaining. Is this the case here?

msevestre: I cannot tell from the way it's used 
`data <- dplyr::bind_rows(lapply(yDataList, .yUnitConverter, yTargetUnit))`

What does `lapply(yDataList, .yUnitConverter, yTargetUnit)` means
Apply yUnitConverter to all element in yData and pass yTargetUnit as second parameter?

Does yUnitConverter  needs to return something?

msevestre (changes requested):

R/utilities-units.R
```diff
@@ -336,3 +336,200 @@ initializeDimensionAndUnitLists <- function() {
...
+#'
+#' @seealso toUnit
+#'
+#' @examples
+#'
+#' # small dataframe to illustrate the conversion
+#' (df <- dplyr::tibble(
+#'   dataType = c(rep("simulated", 3), rep("observed", 3)),
+#'   xValues = c(0, 14.482, 28.965, 0, 1, 2),
+#'   xUnit = "min", xDimension = "Time",
```
msevestre, line 377: put xDimension on anothe rline

R/utilities-units.R
```diff
@@ -336,3 +336,200 @@ initializeDimensionAndUnitLists <- function() {
...
+  # No validation of inputs for this non-exported function.
+  # All validation will take place in the `DataCombined` class itself.
+
+  # The observed and simulated data should have the same units for
+  # visual/graphical comparison.
+  #
+  # Therefore, if target units are not specified by the user, we need to choose
+  # one ourselves. For no special reason, the first element from a vector of
+  # unique units will be selected: one for X-axis, and one for Y-axis, i.e.
+  xTargetUnit <- xUnit %||% unique(data$xUnit)[[1]]
```
msevestre, line 407: What about you take the one with the most occurrences?

R/utilities-units.R
```diff
@@ -336,3 +336,200 @@ initializeDimensionAndUnitLists <- function() {
...
+  # new columns with `split()` makes sure that parts of a data frame where
+  # source units are missing won't be dropped. Note that the original columns
+  # containing source units remain untouched.
+  #
+  # These newly created columns are removed before the converted data frame is
+  # returned to the user.
+  data <- data %>%
+    dplyr::mutate(
+      dplyr::across(
+        .cols = dplyr::matches("Unit$|Weight$"),
```
msevestre, line 433: Do you know for sure that those columns will be here? 
Can you add a comment explaining what columns you are targeting with this ?

R/utilities-units.R
```diff
@@ -336,3 +336,200 @@ initializeDimensionAndUnitLists <- function() {
...
+        .fns = as.character,
+        .names = "{.col}Split"
```
msevestre, line 435: Is this going to add Split at the end of the name for the one matching?

I feel that almost every line of code needs a comment.

R/utilities-units.R
IndrajeetPatil: >  it seems to have side effects

What's the side effect here?

> I cannot tell if we are modifying x in place or not.

**Only** environments exhibit reference semantics in R, everything else has a value semantics. 

So, no, we are not modifying `x` in place. We are creating a new copy and binding it to the same name.

R/utilities-units.R
```diff
@@ -336,3 +336,200 @@ initializeDimensionAndUnitLists <- function() {
...
+      )
+    ) %>%
+    dplyr::mutate(
+      dplyr::across(
+        .cols = dplyr::matches("Split$"),
+        .fns = ~ tidyr::replace_na(.x, "missing")
+      )
+    ) %>%
+    # `split()` will change the row order of the data frame depending on the
```
msevestre, line 444: you talk about split() everywhere but I don't see it being used. Am i blind?

R/utilities-units.R
IndrajeetPatil: And, `x` is a list.

msevestre (commented):
This is me being too novice with dplyr. I have mentioned for the past few months that we need some kind of a workshop to understand this powerful tool. 
I cannot tell what the code does because every single line looks like magic to me. 
Sometimes data is reassigned, sometimes data is changed using %>%. 

This code looks very good. Just out of my league for me to comment

R/utilities-units.R
msevestre: >Only environments exhibit reference semantics in R

R6 class also. They are maybe environment, maybe not. I don't know

R/utilities-units.R
IndrajeetPatil: > Apply yUnitConverter to all element in yData and pass yTargetUnit as second parameter?

Exactly. 

I'd like to use `purrr::map(yDataList, ~ .yUnitConverter(.x, yTargetUnit))` which makes this clearer, but I've been told to use as much base-R as possible.

> Does .yUnitConverter needs to return something?

Yes, for each data frame in the list, ` .yUnitConverter()` returns a data frame with converted units. These data frames are then bound by rows by  `dplyr::bind_rows()`.

R/utilities-units.R
msevestre: well it looks like you are modifying x
x <- purrr::keep(x, ~ nrow(.x) > 0L)
  x
if not, just return `purrr::keep(x, ~ nrow(.x) > 0L)` and name that parameter correctly unless it is common in R to name parameter x for non inline method (e.g. lambda expression)

IndrajeetPatil: Parameter `x` is idiomatic for atomic vector or list arguments.

E.g.

``` r
formals(list2DF)
#> $x
#> list()
#> 
#> $nrow
#> NULL
```

<sup>Created on 2022-04-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

But happy to rename it to something else.

R/utilities-units.R
IndrajeetPatil: Screenshot 2022-04-22 at 17 45 02

msevestre: Ok so you talk about split() on line 424 but the usage is online 458. (more than 30 lines below). 
Why don't you comment close to where something is happening? Mentally, it is very hard to read all those comments if they are not directly related to the code. (at least for me).

IndrajeetPatil: > Ok so you talk about split() on line 424 but the usage is online 458

Because I _need to_. Otherwise, you would be wondering, 

> "why the heck are you doing all this?! Why are you adding _Split_ suffixes? Why are you replacing missing values with 'missing'? 

I need to set the stage for what's coming. If I mention all this when `split()` is used for the first time, it's already too late. Reader is already frustrated by not having answers to all the questions above.

R/utilities-units.R
IndrajeetPatil: > Do you know for sure that those columns will be here?

Yes. `DataCombined$toDataFrame()` guarantees that. Also, the pattern matching makes sure that nothing will happen even if they were not to be found. But, in that case, I am not sure why would be using this internal function any way.

> Can you add a comment explaining what columns you are targeting with this ?

Done.

R/utilities-units.R
IndrajeetPatil: Done

R/utilities-units.R
IndrajeetPatil: > Is this going to add Split at the end of the name for the one matching?

Yes.

> I feel that almost every line of code needs a comment.

Done. They are the worst kind of comments: the ones that just reiterate the code. But, I guess it is essential in this case, since not every contributor can understand that code easily.

R/utilities-units.R
IndrajeetPatil: You mean the one with the most number of corresponding rows?
No strong opinion about that. Seems as good as the current strategy to select the first one.

msevestre: If I have min h h h h h h, I would expect hours to be used and not min. 
But not a big deal

R/utilities-units.R
msevestre: I Think `~ .yUnitConverter(.x, yTargetUnit)` is not clear at all if you are not familiar with the package
I do not know what `.x` means or what `~` means. This is probably very specific to purr. (or maybe not)

I think either code is fine. We are now using 3rd party packages anwyays so we might as well us it as efficient as possible. What we are missing, is a short summary of the main functions of dplyr and purr. I cannot take the time to navigate to their doc and try to understand what this means when there is so much to such for Instead we need a dpply and purr for dummy wiki where we explain the methods we are using with concrete examples

R/utilities-units.R
msevestre: >They are the worst kind of comments: the ones that just reiterate the code

This not what I am asking here. Do not comment the obvious


```
# Return data
return(data) 
```
is indeed not good.

Comment that is going to explain some 3rd party code, that no one in the team is familiar with, is in my opinion not only good, it's absolutely required so that we can maintain this in the future. Until you have managed to bring the team to your level, this is a requirement.

R/utilities-units.R
```diff
@@ -336,3 +336,212 @@ initializeDimensionAndUnitLists <- function() {
...
+}
+
+#' Remove empty data frames from a list
+#'
+#' @description
+#'
+#' Remove empty data frames sometimes produced due to the non-existent
+#' combination of source unit and molecular weight.
+#'
+#' @param x A list.
```
msevestre, line 493: so this is a list of dataframe?

R/utilities-units.R
```diff
@@ -336,3 +336,212 @@ initializeDimensionAndUnitLists <- function() {
...
+#'
+#' @keywords internal
+#' @noRd
+.removeEmptyDataFrame <- function(x) purrr::keep(x, ~ nrow(.x) > 0L)
+
+
+#' @keywords internal
+#' @noRd
+.xUnitConverter <- function(xData, xTargetUnit) {
+  xData$xValues <- ospsuite::toUnit(
```
msevestre, line 503: why do you use  ospsuite::?

IndrajeetPatil: Left it in by mistake. Removed it.

R/utilities-units.R
IndrajeetPatil: Yepp. 

Added that clarification, added an example, and created `.Rd` file for the same.

R/utilities-units.R
IndrajeetPatil: That's a good idea, but it doesn't fall in the scope of this PR, which mostly about making sure that this function can deal with missing values.

Changing the default logic will also involve changing tests, etc. Better to handle it in a separate PR.

R/utilities-units.R
IndrajeetPatil: Okay, fair enough. I had added the comments, but, for the sake of completeness, I've also made a PR to update coding standards to reflect this: https://github.com/Open-Systems-Pharmacology/Suite/pull/108

Also, I want to mention that `DataCombined` code also contains such clarifying comments, which were indeed included to make it easier for the team to understand the code.

IndrajeetPatil:
Hmm, builds are failing due to failure to download pandoc:

```r
Failures
 - pandoc (exited 1) - pandoc not installed. An error occurred during installation:
 The remote server returned an error: (429) Too Many Requests. Too Many Requests
Command exited with code 1
```

IndrajeetPatil:
@msevestre This should be set for a merge. I have completely removed _tidyverse_ use, except `dplyr::bind_rows()`.

I can get rid of it when I work on #891.

msevestre:
The goal is not to change the code . I don't think I was clear. We just need to have a place where we can refer to what those idiomatic R methods are doing so that we can review on confidence.

IndrajeetPatil:
Gotcha. 

I've opened https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/892, which undoes the changes I have done here to not rely on _tidyverse_. After discussion in #890, we can decide how to proceed and which one to merge.

PR #880, opened 2022-04-20
minor improvements for NET task docs and code

IndrajeetPatil:


R/get-net-task.R
```diff
@@ -1,19 +1,28 @@
...
-#' @return An instance of the Task
+#' @return
+#'
+#' - `getNetTask()` returns an instance of of the specified `.NET` task.
+#' - `getContainerTask()` returns a reference to the container task for
+#' optimization purposes only.
+#'
+#' @examples
+#'
+#' if (FALSE) {
```
msevestre, line 13: @IndrajeetPatil Why FALSE? This should run without any problem no?

R/get-net-task.R
```diff
@@ -1,19 +1,28 @@
-#' Returns an instance of the specified .NET Task
+#' Get an Instance of the Specified `.NET` Task
```
msevestre, line 1: Returns an instance is good. Why Get and Instance with capital I?

R/get-net-task.R
```diff
@@ -1,19 +1,28 @@
...
 }
 
+#' @rdname getNetTask
+#' @keywords internal
 getContainerTask <- function() {
-  if (is.null(ospsuiteEnv$containerTask)) {
-    ospsuiteEnv$containerTask <- getNetTask("ContainerTask")
-  }
-  ospsuiteEnv$containerTask
+    ospsuiteEnv$containerTask %||% getNetTask("ContainerTask")
```
msevestre, line 27: This is not equivalent. You are removign setting in the environment.!

msevestre (changes requested):
This introduces a bug as containerTask is not cached. 
As to Get vs Returns... I am not sure I understand

R/get-net-task.R
IndrajeetPatil: I see:

``` r
library(ospsuite)
simulationLoader <- ospsuite:::getNetTask("SimulationLoader")
#> Error in rClr::clrCallStatic("OSPSuite.R.Api", paste0("Get", taskName)): Type:    System.MissingMethodException
#> Message: Could not find a suitable static method GetSimulationLoader on type OSPSuite.R.Api without method parameters
#> Method:  Void ThrowMissingMethod(System.Type, System.String, System.String, System.Type[])
#> Stack trace:
#>    at Rclr.ReflectionHelper.ThrowMissingMethod(Type classType, String methodName, String modifier, Type[] types)
#>    at Rclr.ClrFacade.InternalCallStaticMethod(Type classType, String methodName, Boolean tryUseConverter, Object[] arguments)
#>    at Rclr.ClrFacade.CallStaticMethod(String typename, String methodName, Object[] arguments)
```

<sup>Created on 2022-04-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

msevestre: Let's use a class that exists then.
SimulationRunner for instance is one of them
so 
`simulationRunner <- getNetTask("SimulationRunner")`

And what does the ::: does again?

IndrajeetPatil: `:::` is a way to access non-exported (internal-only, i.e.) package objects.

That works, thanks! I'll change the example.

``` r
library(ospsuite)
simulationLoader <- ospsuite:::getNetTask("SimulationRunner")
```

<sup>Created on 2022-04-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

R/get-net-task.R
IndrajeetPatil: Yikes! How come nothing fails?!

msevestre: This is an optimization. All the code will still work. Just a bit slower

PavelBal: The idea is to call `getNetTaks("ContainerTask") only once, the instance is then cashed and returned from the ospsuiteEnv instead of creating a new one.

msevestre: You can add a test and make sure that this is set after it's being called. But the success of the test might depend on the order of how the tests are run etc... Not as easy as it seems.

R/get-net-task.R
IndrajeetPatil: Added this new example, which is no longer skipped.

R/get-net-task.R
IndrajeetPatil: Reverted the change. I will see if I can think of a way to test this. But, yeah, non-trivial to test something like this.

One possibility is to use `{touchstone}`: https://lorenzwalthert.github.io/touchstone/

I'll explore that option if I find time.

msevestre: I not even sure you'll be able to measure a net difference with the few amounts of call we are doing.

PR #883, opened 2022-04-21
more clean up for `.NET` things

IndrajeetPatil:
Trying to understand some .NET implementations, and while doing so, came across a few things that could be improved:

- prepend internal functions with `.`
- add examples, if possible
- styling docs to follow 80-char rule


NAMESPACE
```diff
@@ -137,7 +137,6 @@ export(validateDimension)
 export(validateUnit)
 import(ospsuite.utils)
 import(rClr)
-import(stringr)
```
IndrajeetPatil, line 140: We are using only a single function. No need to import the entire package. Namespacing `::` is enough.

R/dot-net-wrapper.R
```diff
@@ -1,12 +1,16 @@
-#' @title DotNetRwapper
+#' @title Wrapper class for `.NET` objects
 #' @docType class
-#' @description  Wrapper class for .net Object
 #'
-#' @section Methods:
-#' \describe{
-#'   \item{wrapProperty}{Simple way to wrap a get;set; .NET property}
-#'   \item{wrapReadOnlyProperty}{Simple way to wrap a get; .NET Read-Only property}
-#'   }
```
IndrajeetPatil, line 9: These are private methods, so no need to document them at the top level. 

Moved them to comments for respective method definitions.

msevestre (approved):

PR #892, opened 2022-04-23
Unit converter deals with missing values - with tidyverse

### LINKED ISSUE #875: Make sure `.unitConverter()` doesn't fail in presence of missing values

IndrajeetPatil:
Closes #875 

Same as #876 but uses tidyverse


msevestre:
This looks good to me.

msevestre (approved):

IndrajeetPatil:
> #Iterate through each data set
> for (name in unique(data$name)){

This is inefficient. If there are 20 datasets and only 1 source (x or y) unit, you are calling `toUnit` 60 times!
While my implementation will call it only 3 times: once for x, once y, and once y error (if present).

>  xTargetDim <- data$xDimension[[1]]
>  yTargetDim <- data$yDimension[[1]]

What happens if multiple datasets added to `DataCombined` have different dimensions? Your solution will give inaccurate results, while my implementation can handle it.

PavelBal:
> This is inefficient. If there are 20 datasets and only 1 source (x or y) unit, you are calling toUnit 60 times!
While my implementation will call it only 3 times: once for x, once y, and once y error (if present).

Ah not a big deal, my initial concern was about calling `toUnit` for every single value, which would results in thousands of calls.

> What happens if multiple datasets added to DataCombined have different dimensions? Your solution will give inaccurate results, while my implementation can handle it.

Actually both solutions are wrong, it should be 

```
  xTargetUnit <- xUnit %||% unique(data$xUnit)[[1]]
  yTargetUnit <- yUnit %||% unique(data$yUnit)[[1]]
  xTargetDim <- getDimensionForUnit(xTargetUnit)
  yTargetDim <- getDimensionForUnit(yTargetUnit)
```

Otherwise we might get a mismatch between the dimension and the unit in the output df.

IndrajeetPatil:
> Otherwise we might get a mismatch between the dimension and the unit in the output df.

Good catch! I've included the suggested change.

msevestre:
Let's discuss this on Wedneday. I think this is a good time to talk this out verbally as opposed to text.
@PavelBal In your example is is possible to define a variable and store it instead of repeating ` data[data$name == name, ]` all the time? 

and is it the same as `select(data, "name")`  @IndrajeetPatil 
I need to loose a day playing around with all those options i think

IndrajeetPatil:
For future reference:

Systematic comparison of the merged tidyverse solution and the proposed base-R solution here shows that, even with a small data frame and only a few unique source units, the tidyverse solution is twice as fast as the base-R solution. 

The performance gain will climb rapidly as the number of source units in the data frame and the size of the data frame increases.

``` r
library(ospsuite)

df <- data.frame(
  name = paste0("dataset", 1:6),
  dataType = c(rep("simulated", 3), rep("observed", 3)),
  xValues = c(0, 14.482, 28.965, 0, 1, 2),
  xUnit = "min", 
  xDimension = "Time",
  yValues = c(25.579, 32.446, 32.103, 0, 0.995, 0.991),
  yUnit = c("%", "%", "%", "", "", ""),
  yDimension = "Fraction",
  yErrorValues = c(2.747, 2.918, 2.746, NA, NA, NA),
  yErrorUnit = c("%", "%", "%", NA, NA, NA),
  molWeight = c(NA, NA, NA, 129.1636, 129.1636, 129.1636),
  stringsAsFactors = FALSE
)

baseRUnitConverter <- function(data, xUnit = NULL, yUnit = NULL) {
  xTargetUnit <- xUnit %||% unique(data$xUnit)[[1]]
  yTargetUnit <- yUnit %||% unique(data$yUnit)[[1]]
  xTargetDim <- data$xDimension[[1]]
  yTargetDim <- data$yDimension[[1]]
  xTargetDim <- getDimensionForUnit(xTargetUnit)
  yTargetDim <- getDimensionForUnit(yTargetUnit)

  # Iterate through each data set
  for (name in unique(data$name)) {
    # Get source dimensions and units. Also get source values and MW
    xSourceUnit <- data[data$name == name, ]$xUnit[[1]]
    ySourceUnit <- data[data$name == name, ]$yUnit[[1]]
    yErrorSourceUnit <- data[data$name == name, ]$yErrorUnit[[1]]
    xVals <- data[data$name == name, ]$xValues
    yVals <- data[data$name == name, ]$yValues
    yErrorVals <- data[data$name == name, ]$yErrorValues
    mw <- data[data$name == name, ]$molWeight[[1]]

    # Convert xValues
    data[data$name == name, ]$xValues <- toUnit(
      quantityOrDimension = xTargetDim,
      values = xVals,
      targetUnit = xTargetUnit,
      sourceUnit = xSourceUnit
    )
    # Convert yValues
    data[data$name == name, ]$yValues <- toUnit(
      quantityOrDimension = yTargetDim,
      values = yVals,
      targetUnit = yTargetUnit,
      sourceUnit = ySourceUnit,
      molWeight = mw,
      molWeightUnit = ospUnits$`Molecular weight`$`g/mol`
    )
    # Convert yErrorValues
    data[data$name == name, ]$yErrorValues <- toUnit(
      quantityOrDimension = yTargetDim,
      values = yErrorVals,
      targetUnit = yTargetUnit,
      sourceUnit = yErrorSourceUnit,
      molWeight = mw,
      molWeightUnit = ospUnits$`Molecular weight`$`g/mol`
    )

    # Overwrite xUnit and yUnit and yDimension
    data[data$name == name, ]$xUnit <- xTargetUnit
    data[data$name == name, ]$yDimension <- yTargetDim
    data[data$name == name, ]$yUnit <- yTargetUnit
    data[data$name == name, ]$yErrorUnit <- yTargetUnit
  }

  return(data)
}

bench::mark(
  ospsuite:::.unitConverter(df),
  baseRUnitConverter(df),
  check = TRUE
)
#> # A tibble: 2 x 6
#>   expression                         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 ospsuite:::.unitConverter(df)   32.1ms     36ms      26.5    3.06MB     30.9
#> 2 baseRUnitConverter(df)          67.7ms   73.1ms      13.7  436.28KB     34.2
```

<sup>Created on 2022-04-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

PR #905, opened 2022-04-27
Run simulations should return list

### LINKED ISSUE #863: Sensitivity Analysis:  Return the simulated values
@IndrajeetPatil @PavelBal @Yuri05 

This is definitely possible with the current implementation
At the moment, we create a huge "Variation Table" that is being passed to the PopulationRunner
Each Column is a parameter, each row is a simulation. (or Individual so to speak). Each Row represents the variation of only one parameter

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Services/SensitivityAnalyses/SensitivityAnalysisEngine.cs#L68

The [runResults] https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Services/SensitivityAnalyses/SensitivityAnalysisEngine.cs#L69

are of type `PopulationRunResults`. which has a property internally that is well known in R [Results ](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/PopulationRunResults.cs#L62) of type `SimulationResults`

We are already using this object from ospsuite-R. There is for sure some mapping to be performed to understand what row corresponds to which variation but everything is already here somehow

We could also add this info directly to the `SensitivityAnalysisRunResult` which is also used directly in R

So all in all, I think this could be exposed easily to R

### LINKED ISSUE #895: `runSimulations` should always return a named list, even if only one simulation is executed
Currently, `runSimulation` returns a named list of `SimulationResults` (with IDs of simulations as keys) when mutliple simulations are passed, or only one `SimulationResults` object. I do not think that the type of the output should be different as it leads to some ugly checks in the code when using this function (I have to check if the output is a list or a single object to decide how to access the results).

I propose to always return a named list. `runSimulation` (without an `s`) should continue to return only one result (aditionally, we should introduce a validation that only one simulation is passed).

@msevestre if you agree I will make a PR.

PavelBal:
Fixes #895


R/utilities-simulation.R
```diff
@@ -84,8 +84,6 @@ saveSimulation <- function(simulation, filePath) {
 
 #' @title
 #'
-#' DEPRECATED
```
PavelBal, line 87: Lets not deprecate it any more, I think it makes sense to have the two versions as they now return different types.

R/utilities-simulation.R
```diff
@@ -116,28 +114,41 @@ saveSimulation <- function(simulation, filePath) {
 #' results <- runSimulation(sim, population, simulationRunOptions = simRunOptions)
 #' @export
 runSimulation <- function(simulation, population = NULL, agingData = NULL, simulationRunOptions = NULL) {
-  runSimulations(simulations = simulation, population = population, agingData = agingData, simulationRunOptions = simulationRunOptions)
+  #Check that only one simulation is passed
+  simulation <- c(simulation)
+  validateIsOfLength(simulation, 1)
```
PavelBal, line 119: Checking that only one simulation is passed.

tests/testthat/test-utilities-simulation.R
```diff
@@ -155,6 +155,22 @@ test_that("It runs one individual simulation with simulationRunOptions", {
   expect_true(isOfType(results, "SimulationResults"))
 })
 
+test_that("It throws an erro when trying to run multiple simulations", {
```
msevestre, line 158: error not erro

msevestre: should we also maybe have a more explicit message. e.g. runSimulation is  a convenience method for better user experience and can only be called with a single simulation. To simulate multiple simulations in parallel, use runSimulations

msevestre: @PavelBal Going to fix the typo and merge

Issue #906, opened 2022-04-27
Run tests with R3.6


PavelBal:

PavelBal:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/762

IndrajeetPatil:
cf. https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/887

msevestre:
This build will run every Sunday

msevestre:
maybe the code in simulationResultsToDataFrame only works for 4? 

@Yuri05 When is migration to 4 planned? Then we should stop supporting 3.6

IndrajeetPatil:
I _think_ #908 should fix this.

Yuri05:
>When is migration to 4 planned? Then we should stop supporting 3.6

We will migrate to **4.0.4** this year. Cannot say more precisely, surely not before V11 release

Issue #762, opened 2022-02-10
30 unit tests (testthat) fail locally


Yuri05:
Just installed ospsuite-R 11.0.47 and ospsuite.utils 1.1.0 (Windows, R 3.6.3, R Studio 1.2.5019), manually copied the **tests** folder into the package installation folder from GitHub and executed the unit tests.
Some tests did not work because I have an OSP Suite 10 installed, not OSP Suite 11 - which is OK.

However I've got many failed tests which have nothing to do with the installation of PK-Sim

```
> setwd(system.file('tests', package = 'ospsuite'))
> source('testthat.R')
... 
-- Failure (test-data-combined.R:60:3): dataCombined - either dataSet or SimulationResults provided ---------------------------
myCombDat$dataTransformations not equal to structure(...).
Component “name”: 'current' is not a factor
-- Failure (test-data-combined.R:91:3): dataCombined - either dataSet or SimulationResults provided ---------------------------
unique(df$name) not equal to names(dataSet)[[1]].
'current' is not a factor
-- Failure (test-data-combined.R:394:3): dataCombined - same data order with or without `names` argument ---------------------------
unique(df1$name) not equal to c(...).
'current' is not a factor
-- Failure (test-data-combined.R:781:3): DataCombined works with data grouping ---------------------------
myCombDat3$groupMap$name[[1]] not equal to "Stevens_2012_placebo.Placebo_total".
'current' is not a factor
-- Error (test-data-combined.R:959:3): DataCombined works with population ---------------------------
Error: Could not find an installation of PK-Sim on the machine. Please install the OSPSuite or use 'initPKSim()' to specify the installation path
Backtrace:
    x
 1. \-ospsuite::createPopulationCharacteristics(...) test-data-combined.R:959:2
 2.   \-ospsuite::initPKSim()
-- Failure (test-simulationResultsToDataFrame.R:23:3): simulationResultsToDataFrame works as expected - minimal pkml ---------------------------
unique(df1$unit) not equal to "µmol".
'current' is not a factor
-- Failure (test-simulationResultsToDataFrame.R:24:3): simulationResultsToDataFrame works as expected - minimal pkml ---------------------------
unique(df1$dimension) not equal to "Amount".
'current' is not a factor
-- Failure (test-simulationResultsToDataFrame.R:25:3): simulationResultsToDataFrame works as expected - minimal pkml ---------------------------
unique(df1$TimeUnit) not equal to "min".
[...]

PR #908, opened 2022-04-27
fixes tests for R3?

IndrajeetPatil:
Related to-
#762
#906

For reasons mentioned in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/762#issuecomment-1105076491

### DIFF
--- a/R/utilities-data-set.R
+++ b/R/utilities-data-set.R
@@ -91,8 +91,19 @@ dataSetToDataFrame <- function(dataSets) {
   lloq <- .makeDataFrameColumn(dataSets, "LLOQ")
 
   df <- data.frame(
-    name, xValues, yValues, yErrorValues, xDimension, xUnit, yDimension,
-    yUnit, yErrorType, yErrorUnit, molWeight, lloq
+    name,
+    xValues,
+    yValues,
+    yErrorValues,
+    xDimension,
+    xUnit,
+    yDimension,
+    yUnit,
+    yErrorType,
+    yErrorUnit,
+    molWeight,
+    lloq,
+    stringsAsFactors = FALSE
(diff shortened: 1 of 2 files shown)


IndrajeetPatil:
@msevestre, @Yuri05 How should we do this?

We've two options:
- Merge this to `develop` and then trigger a new build on AppVeyor-3.6
- Yuri runs tests locally with this branch and checks if the tests pass on R 3.6, and then merge to `develop`

msevestre:
Once the changes are ok on appveyor on develop, I'll merge. 
What is the implication of your fix. Is this going to change how the returned DF is behaving. This is effectively a break in the shape of the df correct?

IndrajeetPatil:
This doesn't change _anything_. 

The tests failed due to different **default** behaviour of `data.frame()` function for R versions `>= 4.0` and `< 4.0` due to the following [change](https://cran.r-project.org/bin/windows/base/old/4.0.0/NEWS.R-4.0.0.html) introduced in `R 4.0`:

> R now uses a stringsAsFactors = FALSE default, and hence by default no longer converts strings to factors in calls to data.frame() and read.table().

So the tests failed on `R < 4.0`, but not on `R > 4.0`. 
Explicitly setting `stringsAsFactors = FALSE` makes the default behavior consistent across these two R versions.

IndrajeetPatil:
`stringsAsFactors` parameter default value was notorious before it was changed in R4 and had been a source of long-standing headache for all R users (and, as we are experiencing, continue to be). 

This is a good read that explains what the fuss is all about:
https://simplystatistics.org/posts/2015-07-24-stringsasfactors-an-unauthorized-biography/

PR #910, opened 2022-04-27
fix more tests for R3

IndrajeetPatil:
Follow up on #908; covers remaining instances of `data.frame()` without explicit `stringAsFactors = FALSE` set.

Related to-
#762
#906

For reasons mentioned in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/762#issuecomment-1105076491

### DIFF
--- a/tests/testthat/test-utilities-data-set.R
+++ b/tests/testthat/test-utilities-data-set.R
@@ -24,10 +24,19 @@ test_that("It can convert an empty data set", {
   expect_equal(
     dataSetToDataFrame(dataSet),
     data.frame(
-      name = character(0), xValues = numeric(0), yValues = numeric(0), yErrorValues = numeric(0),
-      xDimension = character(0), xUnit = character(0), yDimension = character(0),
-      yUnit = character(0), yErrorType = numeric(0), yErrorUnit = numeric(0), molWeight = numeric(0),
-      lloq = numeric(0)
+      name = character(0),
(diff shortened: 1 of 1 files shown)


tests/testthat/test-utilities-data-set.R
```diff
@@ -53,12 +71,19 @@ test_that("It can convert a data set with only non-empty fields, except for meta
...
     dataSetToDataFrame(dataSet),
     data.frame(
-      name = rep(dataSet$name, 5), xValues = dataSet$xValues,
-      yValues = dataSet$yValues, yErrorValues = dataSet$yErrorValues,
-      xDimension = rep(dataSet$xDimension, 5), xUnit = rep(dataSet$xUnit, 5),
-      yDimension = rep(dataSet$yDimension, 5), yUnit = rep(dataSet$yUnit, 5),
-      yErrorType = rep(dataSet$yErrorType, 5), yErrorUnit = rep(dataSet$yErrorUnit, 5),
-      molWeight = rep(dataSet$molWeight, 5), lloq = rep(dataSet$LLOQ, 5)
+      name = rep(dataSet$name, 5),
+      xValues = dataSet$xValues,
```
msevestre, line 75: I had comnmented at the time and I think this is another example here. Those tests are too brittle. We need to find a way for instance to not have to repeat 5 a million time. Anyways unrelated to this change

PR #911, opened 2022-04-27
Fixes #909 speed up build

### LINKED ISSUE #909: Build time needs drastic improvement
Almost 20 min on each PR, or update to PR. This takes way too much of our resources. We need faster feedback from CI

What can we remove from the default appveyor build and run nightly

Obviously Coverage. What about vignettes? Should we generate this only on nighty?
Anything else

This should be done consistently across all our R packages but ospsuite-r and reporting engine are the most important one ATM

msevestre:


appveyor-nightly.yml
```diff
@@ -0,0 +1,93 @@
+image: Visual Studio 2019
+
+version: '{build}'
```
msevestre, line 3: This is a copy of our old appveyor.yml. Only one creating artefact. It will run every night

appveyor.yml
```diff
@@ -18,11 +18,9 @@ install:
 environment:
   USE_RTOOLS: true
   NOT_CRAN: true
-  KEEP_VIGNETTES: true
+  KEEP_VIGNETTES: false
```
msevestre, line 21: This is the one that will run on each commit

appveyor.yml
```diff
@@ -79,15 +61,6 @@ artifacts:
   - path: '*.Rcheck\**\*.Rout'
     name: Logs
 
-  - path: "ospsuite_*ubuntu18.tar.gz"
```
msevestre, line 82: no artefact here

appveyor.yml
```diff
@@ -39,29 +37,13 @@ build_script:
...
-before_test:
-  - cinst pandoc
-  - ps: $env:Path += ";C:\Program Files (x86)\Pandoc\"
-  - pandoc -v
-
-on_success:
-  - Rscript -e "covr::codecov(function_exclusions = c('print.*'))"
-
 test_script:
   - travis-tool.sh run_tests
```
msevestre, line 45: I would kike to exclude pksim related tests here and focus on core functions

msevestre (commented):
Going to merge this to trigger builds and see if this works

appveyor-3.6.yml
```diff
@@ -49,36 +46,11 @@ before_test:
   - ps: $env:Path += ";C:\Program Files (x86)\Pandoc\"
   - pandoc -v
 
-on_success:
-  - Rscript -e "covr::codecov(function_exclusions = c('print.*'))"
-
 test_script:
   - travis-tool.sh run_tests
 
-after_test:
```
msevestre, line 58: @IndrajeetPatil This needs to be removed as well.

Yuri05:
if we don't create artifacts in every build - will our permalinks to the artifact in the latest build still work?
(e.g. https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/ospsuite-r/artifacts/ospsuite.zip )

msevestre:
This will point to the nightly now. So you 'll get the latest one a day or we need to trigger by hand

Yuri05:
ok. not sure how AppVeyor knows what is nightly, but as long as it works - I don't care :)

Issue #890, opened 2022-04-22
Create wiki explaining some of the 3rd party methods that we are using


msevestre:
We have introduced some performant code using dplyr, purr, tidyr packages.
But usages of said packages add complexity to the code, specifically for team member like me, who do not know how to use those packages.

we need to document (and it's ok to refer to the documentation if **it is clear**) the following methods and idiomatic R syntax (To be extended)

## Methods

* dplyr::mutate
* dplyr::across
* dplyr::matches
* dplyr::bind_rows

* purrrr::map (used in purrr::map(yDataList, ~ .yUnitConverter(.x, yTargetUnit)))
* 
## Syntax

* what is `~` and` .x` in ` .fns = ~ tidyr::replace_na(.x, "missing") ` or `purrr::map(yDataList, ~ .yUnitConverter(.x, yTargetUnit))`

Why sometime use %>% and sometime assignment
e.g.
```
data <- dplyr::arrange(data, .rowidInternal)

 data <- dplyr::select(data, -dplyr::matches("Split$|.rowidInternal"))
```

vs (might not be syntactically correct)

```
data  
%>% 
  dplyr::arrange(.rowidInternal)
%>% 
dplyr::select(-dplyr::matches("Split$|.rowidInternal"))
```

IndrajeetPatil:
> Why sometime use %>% and sometime assignment

There is no reason to prefer one way or another. They are identical, and what is preferred is just a matter of personal taste. That said, long piped chains can be difficult to debug.

IndrajeetPatil:
> what is ~ and .x in .fns = ~ tidyr::replace_na(.x, "missing") or purrr::map(yDataList, ~ .yUnitConverter(.x, yTargetUnit))

- `~` is lambda syntax to create anonymous functions. 
- `.x` is a pronoun that refers to data.

So, in the following:

```r
purrr::keep(list_of_data_frames, ~ nrow(.x) > 0L)
```

`.x` pronoun refers to `list_of_data_frames`, while `~ nrow(.x) > 0L` creates an anonymous function.

If you didn't want to use an anonymous function, you will have to create a named function:

```r
checkEmptyDataFrame <- function(data) nrow(data) > 0L
purrr::keep(list_of_data_frames, checkEmptyDataFrame)
```

IndrajeetPatil:
> Methods

I don't see the point in documenting them when they are already comprehensively documented on their respective package websites:
[...]

PR #912, opened 2022-04-27
Increase tidyverse code readability for base-R users

IndrajeetPatil:
Related to #890

- Avoids usage of lambda syntax in favor of anonymous functions
- Minimize use of pipe operator

### DIFF
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -484,14 +484,20 @@ DataCombined <- R6::R6Class(
         return(data)
       }
 
-      # Note that `name` column is based on lexical order of names.
-      #
-      # This always works because the order of data frame (and thus the `name`
+      # Create a nested data frame where all columns except `name` are collapsed
+      # into a single column named `data`, which will contain these columns in a
+      # list data frame.
+      data <- tidyr::nest(data, data = -name)
+
+      # Note that `name` column is based on lexical order of `names`. This
+      # always works because the order of data frame (and thus the `name`
       # column) returned by `*ToTibble()` functions never changes.
-      data %>%
-        tidyr::nest(data = -name) %>%
-        dplyr::mutate(name = names) %>%
-        tidyr::unnest(cols = c(data))
+      data <- dplyr::mutate(data, name = names)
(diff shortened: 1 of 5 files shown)


msevestre:
I know I should not commit on draft PR but I am doing it anyways
It is fine to use lambda
it is fine to use the pipe when it makes the code clearer to understand.  Overusing it can be hard to understand. We did not say DO NOT USE pipe this morning. 
Adding the comment on each line however as you did with the non piped version is good. You can do it with the pipe version as well.

IndrajeetPatil:
> It is fine to use lambda

Sure, but it's just one less thing to learn about for base-R users. It's not doing anything that anonymous functions don't do. So might as well not use them.

> it is fine to use the pipe when it makes the code clearer to understand. Overusing it can be hard to understand. We did not say DO NOT USE pipe this morning.

Yepp, that wasn't my takeaway either. There are still some pipes. 
But I agree with Pavel that debugging can be difficult with long chains. So I am using it only in places where it is essential, e.g. for a grouped workflow.

> Adding the comment on each line however as you did with the non piped version is good. You can do it with the pipe version as well.

At this point, I think every single LOC I've written has a comment, piped or not 😅

R/data-combined.R
```diff
@@ -471,14 +484,20 @@ DataCombined <- R6::R6Class(
...
-      # This always works because the order of data frame (and thus the `name`
-      # column) returned by `*ToDataFrame()` functions is never changed.
-      data %>%
-        tidyr::nest(data = -name) %>%
-        dplyr::mutate(name = names) %>%
-        tidyr::unnest(cols = c(data))
+      # Create a nested data frame where all columns except `name` are collapsed
+      # into a single column named `data`, which will contain these columns in a
+      # list data frame.
+      data <- tidyr::nest(data, data = -name)
```
msevestre, line 490: @IndrajeetPatil Did not you mentioned that using pipe was also more efficient for memory usage? or am I imaginating it?

msevestre (approved):

PR #936, opened 2022-05-12
Don't update linetype when it was blank before

### LINKED ISSUE #933: Legend update shouldn't introduce new lines where linetype was blank
image

IndrajeetPatil:
Closes #933


R/utilities-plotting.R
```diff
@@ -440,27 +443,39 @@
...
 
-  # Extract as many properties as there are datasets from the specified config
-  # object fields.
-  if (length(pointsColor) > 1L) {
-    pointsColor <- pointsColor[1:nrow(legendCaptionData)]
+  # As many unique colors and shapes will be needed as there are groupings
+  numberOfColorsNeeded <- numberOfShapesNeeded <- nrow(legendCaptionData)
+
+  # Number of needed line types would be equal to only non-blank line types.
+  numberOfLinetypesNeeded <- nrow(legendCaptionData[legendCaptionData$linetype != "blank", ])
```
msevestre, line 450: don't we have an enum value for this?

IndrajeetPatil: Done.

R/utilities-plotting.R
```diff
@@ -440,27 +443,39 @@
...
     color = pointsColor,
-    shape = pointsShape,
-    linetype = lineTypes
+    shape = pointsShape
   )
 
+  # New version of legend mappings for linetype: only non-blank line rows are replaced.
+  # Relevant only if there were any blank lines in the original legend data frame.
+  if (numberOfLinetypesNeeded > 0L) {
+    newLegendCaptionData[newLegendCaptionData$linetype != "blank", ]$linetype <- lineTypes
```
msevestre, line 477: Same here too.

IndrajeetPatil:
It has nothing to do with the PR.

The plot looks different every time the test is run. This is why I am not running that test on AppVeyr. I am assuming this has something to do with how population simulations are run but not sure.

PavelBal (approved):

PR #942, opened 2022-05-13
groups can be added via `$add*()` methods

### LINKED ISSUE #937: Setting a grouping in `DataCombined` when adding a data set
I cannot remember - why (and where) did we decide to **not** allow setting the group property in the `add` methods?

So instead of 

```
  currentDataCombined$addSimulationResults(simulationResults = simulatedScenarios[[scenarioName]]$results,
                                           quantitiesOrPaths = OutputPaths$bOH_production_fraction_liv,
                                           names = "4ßOH liver production fraction simulated")
  currentDataCombined$setGroups(names = "4ßOH liver production fraction simulated", groups = "liver")
```

it would be 

```
  currentDataCombined$addSimulationResults(simulationResults = simulatedScenarios[[scenarioName]]$results,
                                           quantitiesOrPaths = OutputPaths$bOH_production_fraction_liv,
                                           names = "4ßOH liver production fraction simulated",
                                           groups = "liver"
)
```

I find it not convenient to do it in two steps, as I have to keep track of the names and repeat them in the code.

IndrajeetPatil:
Closes #937

### DIFF
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -14,6 +14,14 @@
 #' @import tidyr
 #' @import ospsuite.utils
 #'
+#' @param groups A string or a list of strings assigning the data set to a
+#'   group. If an entry within the list is `NULL`, the corresponding data set is
+#'   not assigned to any group (and the corresponding entry in the `group`
+#'   column will be an `NA`). If provided, `groups` must have the same length as
+#'   `dataSets` and/or `simulationResults$quantityPath`. If no grouping is
+#'   specified for any of the dataset, the column `group` in the data frame
+#'   output will be all `NA`.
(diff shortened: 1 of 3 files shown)


R/data-combined.R
```diff
@@ -14,6 +14,14 @@
 #' @import tidyr
 #' @import ospsuite.utils
 #'
+#' @param groups A string or a list of strings assigning the data set to a
```
IndrajeetPatil, line 17: This is a common parameter across both `$add*()`. So, to avoid duplication of documentation, I am documenting it at the class documentation level.

PavelBal (approved):

PR #941, opened 2022-05-13
Don't store axes labels in plot config

IndrajeetPatil:
Closes #938
Closes #939

PR #943, opened 2022-05-13
Follow up on #941

IndrajeetPatil:
- Remove Concentration (molar) per time from "conc" list
- Add comments and examples


R/plot-individual-time-profile.R
```diff
@@ -56,6 +56,9 @@ plotIndividualTimeProfile <- function(dataCombined,
 
   # axes labels -----------------------------
 
+  # The type of plot can be guessed from the specific `PlotConfiguration` object
+  # used, since each plot has a unique corresponding class. The labels can then
+  # be prepared accordingly.
   plotType <- class(timeProfilePlotConfiguration)[[1]]
```
msevestre, line 62: I don;'t think this is right. You should pass the config as parameter and either
1 - do this weird class stuff in the .createAxesLabels (otherwise you will duplicate this code everywhere)
2 - OR create some methods in the configuration that returns what you need so that you do not have to do switch in your code (Tell don't ask principle => don't ask the class what type it is. Let the class tell you what properties you need)

Also the code as a copy paste error I think

```
xLabel <- switch(plotType,
    "TimeProfilePlotConfiguration" = xUnitString,
    "ResVsPredPlotConfiguration" = xUnitString,
    "ObsVsPredPlotConfiguration" = paste0("Observed values (", yUnitString, ")")
  )

```
As you can see ou are using yUnitString instead of xUnitString. If this is intended, a comment is required

msevestre (changes requested):

PR #944, opened 2022-05-13
Update roxygen docs with the 7.2.0 version

IndrajeetPatil:
Released today:
https://cran.r-project.org/web/packages/roxygen2/index.html


msevestre:
In general, it's good to wait a bit before updating a component, My experience is that you can lose a lot of time when being too eager. Seems to work in this case

PR #951, opened 2022-05-14
Avoid generic name `df` for data frames

IndrajeetPatil:


msevestre:
Yep. Easier in the long run. Note that this is ok in my opinion when the code is structured in smaller chunks

PR #953, opened 2022-05-16
Fixes #946 try catch follow up

### LINKED ISSUE #946: Encoding issue: trycatch aroung getUnitEnums so that loading works also on chinese/japanese OS

msevestre:


R/utilities-units.R
```diff
@@ -288,6 +288,7 @@ getUnitsForDimension <- function(dimension) {
   validateIsString(dimension)
   dimensionTask <- getDimensionTask()
   rClr::clrCall(dimensionTask, "AllAvailableUnitNamesFor", enc2utf8(dimension))
+  stop("errror")
```
msevestre, line 291: @IndrajeetPatil THis is on purpose just to ask you why the other code does not work

R/utilities-units.R
```diff
@@ -347,14 +348,18 @@ getUnitsEnum <- function() {
...
-        warning(messages$errorLoadingUnitsForDimension(dimension, cond))
+        errors <- c(errors, dimension)
         # making sure that in this case, the user sees that something went wrong
         return(c("Unavailable"))
       }
     )
     return(enum(replace(x, x == "", "Unitless")))
   })
 
+  if (length(errors) > 0L) {
```
msevestre, line 359: @IndrajeetPatil The errors vector is always empty. I am a bit confused here as I thought R would work with closure. What  am I doing wrong here. In this logic, I am collecting all errors at the function level and showing only one message. But it does not work.

R/utilities-units.R
```diff
@@ -347,14 +348,18 @@ getUnitsEnum <- function() {
...
+    message(messages$errorLoadingUnitsForDimensionoadingUnitsForDimension(errors))
```
IndrajeetPatil, line 360:
```suggestion
    message(messages$errorLoadingUnitsForDimension(errors))
```

R/utilities-units.R
IndrajeetPatil: Hmm, at face value, this seems perfectly fine. I will see if I can figure out the issue locally.

R/utilities-units.R
msevestre: sure this is copy paste BS. But why isn't the code working?

R/utilities-units.R
```diff
@@ -347,14 +348,18 @@ getUnitsEnum <- function() {
         getUnitsForDimension(dimension = dimension)
       },
       error = function(cond) {
-        warning(messages$errorLoadingUnitsForDimension(dimension, cond))
+        errors <- c(errors, dimension)
```
IndrajeetPatil, line 351: You need to do super-assignment here:

```suggestion
        errors <<- c(errors, dimension)
```

msevestre:
What is super assignment?

IndrajeetPatil:
> What is super assignment?

Ref: https://adv-r.hadley.nz/environments.html

image

msevestre:
Ok cool

PR #956, opened 2022-05-17
Remove function to update legend caption data frame

IndrajeetPatil:
Made unnecessary after change in https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/271

Can only be considered for review after [this PR](https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/271) is merged in the _tlf_.


IndrajeetPatil:
The update in plot is expected. Previously, the transparency parameters in plot config were being ignored. Now they are not.

Therefore, the colors get lighter.

PR #959, opened 2022-05-17
Default to frequent unit in `.unitConverter()`

### LINKED ISSUE #891: Use the most frequent unit for default conversion in `.unitConverter()`
Instead of the current approach of using the _first_ encountered unit in the column.

As suggested by @msevestre here: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/876#discussion_r856348349

IndrajeetPatil:
Closes #891

``` r
(df <- dplyr::tibble(
  xValues = c(15, 30, 60),
  xUnit = "min",
  xDimension = "Time",
  yValues = c(0.25, 45, 78),
  yUnit = c("", "%", "%"),
  yDimension = "Fraction",
  molWeight = 10
))
#> # A tibble: 3 x 7
#>   xValues xUnit xDimension yValues yUnit yDimension molWeight
#>     <dbl> <chr> <chr>        <dbl> <chr> <chr>          <dbl>
#> 1      15 min   Time          0.25 ""    Fraction          10
#> 2      30 min   Time         45    "%"   Fraction          10
#> 3      60 min   Time         78    "%"   Fraction          10

ospsuite:::.extractMostFrequentUnit(df, xUnit)
#> [1] "min"

ospsuite:::.extractMostFrequentUnit(df, yUnit)
#> [1] "%"

ospsuite:::.unitConverter(df)
#> # A tibble: 3 x 7
#>   xValues xUnit xDimension yValues yUnit yDimension molWeight
#>     <dbl> <chr> <chr>        <dbl> <chr> <chr>          <dbl>
#> 1      15 min   Time            25 %     Fraction          10
#> 2      30 min   Time            45 %     Fraction          10
#> 3      60 min   Time            78 %     Fraction          10
```

<sup>Created on 2022-05-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>


R/utilities-units.R
```diff
@@ -456,19 +456,20 @@ initializeDimensionAndUnitLists <- function() {
...
+
   # The observed and simulated data should have the same units for
   # visual/graphical comparison.
   #
   # Therefore, if target units are not specified by the user, we need to choose
-  # one ourselves. For no special reason, the first element from a vector of
-  # unique units will be selected: one for X-axis, and one for Y-axis, i.e.
-  xTargetUnit <- xUnit %||% unique(data$xUnit)[[1]]
-  yTargetUnit <- yUnit %||% unique(data$yUnit)[[1]]
+  # one ourselves. For no special reason, the most frequent units will be
```
PavelBal, line 468: Remove the "For no special reason".

R/utilities-units.R
```diff
@@ -495,7 +496,8 @@ initializeDimensionAndUnitLists <- function() {
 
   # `yErrorUnit` column won't be present when only simulated datasets are
   # entered, but it can be assumed to be the same as `yUnit`.
-  if (!"yErrorUnit" %in% names(data)) {
+  if (("yErrorValues" %in% names(data)) &&
```
PavelBal, line 499: Could you explain this change?

IndrajeetPatil: If there is no `yErrorValues` column in the entered data frame, it doesn't make sense for this function to introduce a new column called `yErrorUnit`.

R/utilities-units.R
```diff
@@ -456,19 +456,20 @@ initializeDimensionAndUnitLists <- function() {
...
   #
   # Therefore, if target units are not specified by the user, we need to choose
-  # one ourselves. For no special reason, the first element from a vector of
-  # unique units will be selected: one for X-axis, and one for Y-axis, i.e.
-  xTargetUnit <- xUnit %||% unique(data$xUnit)[[1]]
-  yTargetUnit <- yUnit %||% unique(data$yUnit)[[1]]
+  # one ourselves. For no special reason, the most frequent units will be
+  # selected: one for X-axis, and one for Y-axis. If multiple units are tied in
+  # terms of their frequency, the first will be selected.
+  xTargetUnit <- xUnit %||% .extractMostFrequentUnit(data, xUnit)
```
PavelBal, line 471: `.extractMostFrequentUnit(data, xUnit)` will only be executed, when `xUnit` is `NULL`, so why are you passing it to the function?

R/utilities-units.R
```diff
@@ -641,3 +643,49 @@ initializeDimensionAndUnitLists <- function() {
...
+#'   xDimension = "Time",
+#'   yValues = c(0.25, 45, 78),
+#'   yUnit = c("", "%", "%"),
+#'   yErrorUnit = c("", "%", "%"),
+#'   yDimension = "Fraction",
+#'   molWeight = 10
+#' )
+#'
+#' ospsuite:::.extractMostFrequentUnit(df, xUnit)
+#' ospsuite:::.extractMostFrequentUnit(df, yUnit)
```
PavelBal, line 667: `xUnit` and `yUnit` are variables not strings... I don't understand  this.

R/utilities-units.R
```diff
@@ -641,3 +643,49 @@ initializeDimensionAndUnitLists <- function() {
 
   return(yData)
 }
+
+
+#' Find the most common units
+#'
+#' @inheritParams .unitConverter
+#' @param ... The name of the column containing units (e.g. `xUnit`).
```
PavelBal, line 651: Why `...` and not `byColumn` or `unitColumn` or something? I hate `...`

PavelBal (changes requested):

R/utilities-units.R
```diff
@@ -495,7 +496,8 @@ initializeDimensionAndUnitLists <- function() {
 
   # `yErrorUnit` column won't be present when only simulated datasets are
   # entered, but it can be assumed to be the same as `yUnit`.
-  if (!"yErrorUnit" %in% names(data)) {
+  if (("yErrorValues" %in% names(data)) &&
+    (!"yErrorUnit" %in% names(data))) {
```
PavelBal, line 500: Please place the `!` before the brackets. This way it reads like "if NOT "yErrorValues" is in names".. weird

R/utilities-units.R
IndrajeetPatil: I was trying to avoid using embrace operator. 

At any rate, I've changed this now.

R/utilities-units.R
IndrajeetPatil: This is R's non-standard evaluation.

The `xUnit` on the left is argument and evaluated in the function environment.
The `xUnit` in `.extractMostFrequentUnit()` is the column name and is evaluated in the `data` mask.

I am naming the arguments now, so shouldn't be confusing.

R/utilities-units.R
```diff
@@ -456,19 +456,20 @@ initializeDimensionAndUnitLists <- function() {
...
   #
   # Therefore, if target units are not specified by the user, we need to choose
-  # one ourselves. For no special reason, the first element from a vector of
-  # unique units will be selected: one for X-axis, and one for Y-axis, i.e.
-  xTargetUnit <- xUnit %||% unique(data$xUnit)[[1]]
-  yTargetUnit <- yUnit %||% unique(data$yUnit)[[1]]
+  # one ourselves. The most frequent units will be selected: one for X-axis, and
+  # one for Y-axis. If multiple units are tied in terms of their frequency, the
+  # first will be selected.
+  xTargetUnit <- xUnit %||% .extractMostFrequentUnit(data, unitColumn = xUnit)
```
PavelBal, line 471: Oh wow, I see. Please let's use strings `""` when these are really strings, otherwise it is super confusing. I always think we are passing the value with a variable. So `unitColumn = "xUnit")

PavelBal (approved):

PR #979, opened 2022-05-24
Fixes nightly build issue?

IndrajeetPatil:


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: ospsuite
 Title: R package to manipulate OSPSuite Models
-Version: 12.0.0.9000
+Version: 12.0
```
Yuri05, line 4: we should keep the version AS IS, the weird problem on AppVeyor has gone

IndrajeetPatil: Cool; good to know that the issue is not the package version.

msevestre:
Probably it was the update of the version not synchronized with the update of the global version on appveyor. I updated the version yesterday to 12.

PR #985, opened 2022-05-24
984 down version 11.1

msevestre:


msevestre:
This will set the version of the package to 11.1 (compatible with v11) and not to 12 (which is not compatible)

PR #987, opened 2022-05-25
Fixes #986 Nightly build failing

### LINKED ISSUE #986: Nightly build failing
image

No idea what's going on., It seems that the build cannot find / does not produce the tar package?

Yuri05:
Nightly fails now because the Linux patching tries to update **ospsuite_11.1.N.tar.gz** - but the package created is **ospsuite_11.1.N.9000.tar.gz**

So I removed this .9000 from the DESCRIPTION and instead set `mode: devel` in **_pkgdown.yml**
With this, the website will be generated under **docs/dev** until the mode is changed to `mode: release` (which must be done once before release).

Probably a better solution anyway and we should do this for all other packages as well?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: ospsuite
 Title: R package to manipulate OSPSuite Models
-Version: 11.1.0.9000
+Version: 11.1.0
 Authors@R: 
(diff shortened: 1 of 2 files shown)


IndrajeetPatil (approved):
Sounds like a good compromise to me.

PR #1004, opened 2022-06-17
Add function for getting the paths of state variable parameters

### LINKED ISSUE #1003: `getAllStateVariablesParameterPaths`
In addition to the `getAllStateVariablesPaths` function that returns the paths of all molecules plus state variable parameters, a function that returns the paths of state variable parameters only is required for steady-state simulations with `SimulationBatch`.

PavelBal:
Fixes #1003


msevestre (changes requested):
@PavelBal Can you add one test calling this method? This will ensure that if the signature ever changes in core, we'll catch it

PavelBal:
@msevestre just did.

PR #945, opened 2022-05-13
Implement `plotObservedVsSimulated()`

IndrajeetPatil:
TODO

- [x] Implement interpolation function
- [x] Implement `plotObservedVsSimulated()`
- [x] tests for `plotObservedVsSimulated()` 
- [x] check comments
- [x] handle empty `DataCombined` objects
- [x] Address Pavel's code review comments (https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/945#issuecomment-1173770611)

Features that need to be implemented in _tlf_, and thus should not block this PR from being merged:
- https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/332
- https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/333

Related to #674 

Discussed in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/discussions/809

Closes #804 
Closes #924
Closes [#1010](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1010)
Closes [#1011](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1011)
Closes #1012

PR #1006, opened 2022-06-23
Is explicit formula by path

### LINKED ISSUE #997: Function `IsExplicitFormulaByPath`
Various scenarios require checking if a value of a quantity is given by an explicit formula or a constant. As we found out, getting this property from path using the `ContainerTask` and `IsExplicitFormulaByPath` method is much more efficient than actually creating the quantity objects.

So far, `IsExplicitFormulaByPath` has only been used experimentally in `esqlabsR` package for calculating the steady-state of a simulation. Now, the parameter identification package also requires this function. I have been using the `:::` reference to get the container task and call the .NET method:

```
task <- ospsuite:::getContainerTask()
isFormulaExplicit <- rClr::clrCall(task, "IsExplicitFormulaByPath", simulation$ref, enc2utf8(path))
```

However, I do not want to use this approuch in the "official" package, so I would vote for creating and exporting a `isExplicitFormulaByPath`-function.

@msevestre

PavelBal:
Fixes #997


msevestre:
buid failed

IndrajeetPatil:
The build failure is unrelated to the current PR. 

A test is failing because _tlf_ has been updated, but the snapshots for profile plots haven't been updated for this change. 

I am on vacation so I can't update them. Either someone else updates snapshot, or you can comment out the failing test for now.

IndrajeetPatil:
If you go to the file containing the failing test and run `devtools::test_active_file()`, the failure message will tell you how to update the snapshot. 

Basically, it will be running something like `snapshot_accept(...)`.

PavelBal:
Not sure where I have to run it... Do I have to be in a specific directory? Otherwise, just update it when you are back ;)

```
> testthat::snapshot_review('plot-population-time-profile/')
No snapshots to update
```

IndrajeetPatil:
In that case, you can just comment it out for now. 

Unless someone changes the `plot-*.R` files, it is unlikely that anything will be broken by mistake for the plotting functions.

msevestre:
I get it that tlf was updated and snapshot didn't match. But this action should have led to a build failing previously since it's not part of this commit. So why is the build failing *now*?

PavelBal:
> I get it that tlf was updated and snapshot didn't match. But this action should have led to a build failing previously since it's not part of this commit. So why is the build failing now?

Build is triggered on PR commit, or as nightly... Well this is the first PR (or actually not, https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/945 fails too now, as it has a commit after update of TLF), and the nightly build of develop branch is also failing.

We definitely need an instruction somewhere how to update the snapshots.

msevestre:
The update to TLF was the problem. The commit with this change should not have been merged
Also snapshot tests should not be part of nightly. This is just the PKSim stuff normally

msevestre:
>Well this is the first PR (or actually not, https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/945 fails too now, as it has a commit after update of TLF),

This commit does not update TLF. So I am very confused as to why the commit before worked and not this one.

msevestre:
>In ospsuite's AppVeyor workflow, we fetch binaries for tlf for the most recent commit

Oh I see. Ok. I am not sure this makes sense when we are working on a stable version.  (ok for me on a v12 but here, this is still v11)

Issue #1008, opened 2022-06-27
`plotIndividualTimeProfile` - legend contains "group" entry


PavelBal:
![image](https://user-images.githubusercontent.com/25903040/176049879-b5bfdacf-8a28-46df-b15a-28e3ed35fb51.png)

This is new, I guess. It should not be in the legend.

PR #1009, opened 2022-06-28
Fixes #718 and fixes build

### LINKED ISSUE #718: Irritating error message
After installation of ospsuite-R on a new machine I got the error message below:
grafik

Took a while till I noticed that R-Studio has taken the **32 bit version of R** as per default (both 32 and 64 bit versions of R are installed).
After changing the settings of R Studio (so that the 64 bit version of R is taken) everything was fine.

Let's improve the error message if ospsuite-R is erroneously started on 32 bit R

msevestre:


Yuri05:
The 64-bit change looks good.
Not sure about those new SVGs though. They contain this unnesessary "group" keyword in the caption (s. https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1008) and must be reverted back again as soon as this bug is fixed.
So maybe we should just temporarily disable the corresponding tests instead?

msevestre:
Well, I want to merge all the PR. So we can revert after this. @IndrajeetPatil said that the code was ok so I am just going with this now.

PR #1015, opened 2022-07-05
- Added function getQuantityValuesByPath

### LINKED ISSUE #1013: Update to latest core
To support GetValueByPath

PavelBal:
Fixes #1013

@msevestre You can review already. As soon PK-Sim gets the updated dlls, I will run styler and do a commit to trigger a build.


R/utilities-quantity.R
```diff
@@ -160,7 +159,13 @@ setQuantityValuesByPath <- function(quantityPaths, values, simulation, units = N
       if (dimension == "") {
         next
       }
-      value <- toBaseUnit(quantityOrDimension = dimension, values = value, unit = units[[i]])
+      #If the unit is NULL, the value is assumend to be in base unit and no conversion
+      #in necessary
+      if (!is.null(units[[i]])){
+        mw <- simulation$molWeightFor(path)
```
PavelBal, line 165: Good thing the method requires a path, so we can safely use it with the performance-optimized function.

PavelBal:
Huh, how is the build passing? :D

msevestre:
PKSim stuff is only done on nightly builds

msevestre:
we can merge this. The buildis working

PR #1020, opened 2022-07-18
Add Molecular weight to dapagliflozin data sheet

PavelBal:
Just added some data to example data sheets.


IndrajeetPatil (approved):
Changes to binary files, so not much I can review.

image

PR #1032, opened 2022-07-19
runSimulationBatches returns a named list

### LINKED ISSUE #1031: `runSimulationBatches` does not return a named list
The documentation states

> #' @return Nested list of `SimulationResults` objects. The first level of the list are the IDs of the simulations of SimulationBatches, containing a list of `SimulationResults` for each set of parameter/initial values. If a simulation with a parameter/initial values set fails, the result for this run is `NULL`

However, this should be the IDs of the batches and not the simulations (as multiple batches can be created from the same simulation). Variable naming is confusing and the documentation.

PavelBal:
Fixes #1031


R/utilities-simulation.R
```diff
@@ -374,35 +377,38 @@ runSimulationBatches <- function(simulationBatches, simulationRunOptions = NULL,
...
 
   simulationBatches <- c(simulationBatches)
-  # Result Id <-> simulation batch pointer id map to get the correct simulation for the results.
-  # Using the Id of the pointer instead of the Id of the simulation as multiple
+  # Result Id <-> simulation batch id map to get the correct simulation for the results.
+  # Using the Id of the batch instead of the Id of the simulation as multiple
   # SimulationBatches can be created with the same simulation
   # Each SimulationBatchRunValues has its own id, which will be the id of the result
-  resultsIdSimulationIdMap <- list()
+  resultsIdSimulationBatchIdMap <- list()
```
PavelBal, line 384: Renamed variables to reflect the actual code.

R/simulation-batch.R
```diff
@@ -110,6 +111,10 @@ SimulationBatch <- R6::R6Class(
       } else {
         private$throwPropertyIsReadonly("runValuesIds")
       }
+    },
+    #' @field id The id of the .NET wrapped object. (read-only)
```
IndrajeetPatil, line 115:
```suggestion
    #' @field id A unique identifier for the `.NET` wrapped object. (read-only)
```

IndrajeetPatil (changes requested):
You can delete the new image in `tests/testthat/_snaps/plot-population-time-profile`.

What's up with the two new image files (`figure.png` and `figure_.png`) that have been introduced in `tests/testthat` folder?

R/simulation-batch.R
msevestre: it sounds like it's not the one from .net where in fact, it IS

R/utilities-quantity.R
```diff
@@ -211,20 +215,24 @@ getQuantityValuesByPath <- function(quantityPaths, simulation, units = NULL, sto
...
-      #in necessary
-      if (!is.null(units[[i]])){
+      # If the unit is NULL, the value is assumend to be in base unit and no conversion
+      # in necessary
+      if (!is.null(units[[i]])) {
         mw <- simulation$molWeightFor(path)
-        value <- toUnit(quantityOrDimension = dimension, values = value, targetUnit = units[[i]],
-                        molWeight = mw)
+        value <- toUnit(
+          quantityOrDimension = dimension, values = value, targetUnit = units[[i]],
```
msevestre, line 233: same here

R/utilities-quantity.R
```diff
@@ -159,12 +159,14 @@ setQuantityValuesByPath <- function(quantityPaths, values, simulation, units = N
...
-      #in necessary
-      if (!is.null(units[[i]])){
+      # If the unit is NULL, the value is assumend to be in base unit and no conversion
+      # in necessary
+      if (!is.null(units[[i]])) {
         mw <- simulation$molWeightFor(path)
-        value <- toBaseUnit(quantityOrDimension = dimension, values = value, unit = units[[i]],
-                            molWeight = mw)
+        value <- toBaseUnit(
+          quantityOrDimension = dimension, values = value, unit = units[[i]],
```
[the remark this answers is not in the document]
PavelBal: Styler, I didn't change anything manually.

R/utilities-simulation.R
```diff
@@ -374,35 +377,38 @@ runSimulationBatches <- function(simulationBatches, simulationRunOptions = NULL,
...
 
   # Run the batch with the ConcurrentSimulationRunner
   results <- rClr::clrCall(simulationRunner, "RunConcurrently")
-  simulationResults <- .getConcurrentSimulationRunnerResults(results = results, resultsIdSimulationIdMap = resultsIdSimulationIdMap, simulationIdSimulationMap = simulationIdSimulationMap, silentMode = silentMode)
-
-  # output: list of lists of SimulationResults, one list per SimulationBatch
-  output <- lapply(names(simulationIdSimulationMap), function(simId) {
-    simulationResults[which(resultsIdSimulationIdMap == simId)]
-  })
+  simulationResults <- .getConcurrentSimulationRunnerResults(results = results, resultsIdSimulationIdMap = resultsIdSimulationBatchIdMap, simulationIdSimulationMap = simulationBatchIdSimulationMap, silentMode = silentMode)
```
msevestre, line 403: one parameter by line here as well. I need to scroll to see the code

R/simulation-batch.R
PavelBal: I would leave it as "ID" as this is how the field in .NET is called..

R/utilities-simulation.R
```diff
@@ -374,35 +377,38 @@ runSimulationBatches <- function(simulationBatches, simulationRunOptions = NULL,
...
+
+  # Returned is a named list of results with names being the IDs of the batches
+  output <- vector("list", length(simulationBatches))
+  names(output) <- names(simulationBatchIdSimulationMap)
+  # Iterate through simulation batch ids
+  for (simBatchId in names(simulationBatchIdSimulationMap)) {
```
msevestre, line 409: Can't you ued the output from before and set the name after the fact? creating the vector and a for seems so not R like :)

msevestre (changes requested):
minor changes

R/utilities-quantity.R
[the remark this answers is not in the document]
PavelBal: Re-styled

R/utilities-simulation.R
PavelBal: Reverted and just added assigning names :)

IndrajeetPatil (approved):

PR #1038, opened 2022-07-23
Add caps to error bars

IndrajeetPatil:
Catching up with https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/344


IndrajeetPatil:
The build is failing for the default branch, so it will be unsurprising if it also fails here.

IndrajeetPatil:
Hmm, there is one unexpected test failure. Probably, just the snapshot needs to be updated.

IndrajeetPatil:
Alright, build is succeeding now.

Yuri05 (approved):

PR #1040, opened 2022-07-24
Fixes #1035 enable pksim tests

### LINKED ISSUE #1035: Enable PK-Sim tests
After PK-Sim 11.1 with updated DLLs has been built.

msevestre:


msevestre:
The build is failing but this has nothing to do with the code. Some snapshot are outdated

PR #1085, opened 2022-08-16
Export `calculateResiduals()` and `convertUnits()`

### LINKED ISSUE #1081: Turn on tests for `plotObservedVsSimulated()` on CI
Once this issue is resolved:
https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/367

### LINKED ISSUE #1084: Export some plotting utility functions
Functions which have a broader utility.

- [ ] `ospsuite:::.calculateResiduals()`
- [ ] `ospsuite:::.unitConverter()`

If we decide to export the unit conversion function, I think it should be called `convertUnits()` (or something similar), since the verb should come first in the function name.

IndrajeetPatil:
TODO:

- [x] Export `calculateResiduals()`
- [x] Export `convertUnits()`
- [x] Add docs for both of them
- [x] Update NEWS
- [x] Closes #1081 
- [x] Closes #1084


R/utilities-data-combined.R
```diff
@@ -1,4 +1,344 @@
...
+  pairedData <- dplyr::tibble(
+    "xValues"      = observedData[, "xValues"],
+    "xUnit"        = unique(data$xUnit),
+    "xDimension"   = unique(data$xDimension),
+    "yValues"      = observedData[, "yValues"],
+    "yErrorValues" = yErrorValues,
+    "yUnit"        = unique(data$yUnit),
+    "yDimension"   = unique(data$yDimension),
+    "predValue"    = simulatedData[timeMatchedData, "yValues"]
+  )
```
Yuri05, line 228: I wonder if it wouldn't be better for the understanding to rename the columns:
* X-Columns to Time-columns (e.g. `xValues`=>`timeValues `etc.) 
* Y-Columns to Observed-Columns (e.g. `yValues`=>`observedValues`)
* `predValue `=> `predictedValues `(in any case it  should be plural)

R/utilities-data-combined.R
```diff
@@ -1,4 +1,344 @@
...
+
+  # The linear scaling is represented either of the following:
+  #
+  # - `"lin"` (in `DefaultPlotConfiguration`)
+  # - `"identity"` (in `tlf::PlotConfiguration`, because of `{ggplot2}`)
+  if (scaling %in% c("lin", "identity")) {
```
Yuri05, line 234: why using strings instead of `tlf::Scaling$lin` etc.?

IndrajeetPatil: Because `tlf::Scaling` doesn't have `"identity"` entry.

R/utilities-data-combined.R
IndrajeetPatil: I am just trying to be consistent with the naming schema present in the original data frame from which this paired data frame is extracted:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/af89b3a242649124e077458ce29f37dd87516b72/R/data-combined.R#L378-L381

Yeah, I agree. Following names, although verbose, will be better:

- predValue -> predictedValues
- resValue -> residualValues

R/utilities-data-combined.R
```diff
@@ -1,4 +1,344 @@
...
+#'   yValues = c(1.9, 6.1, 7, 8.2, 1, 0, 4, 8),
+#'   yErrorValues = rnorm(8),
+#'   yUnit = ospUnits$`Concentration [mass]`$`mg/l`,
+#'   yDimension = ospDimensions$`Concentration (mass)`
+#' )
+#'
+#' ospsuite:::.extractResidualsToTibble(df)
+#'
+#' @keywords internal
+.extractResidualsToTibble <- function(data, scaling = tlf::Scaling$lin) {
```
Yuri05, line 197: I think the `scaling `argument should be **mandatory** here?

R/utilities-data-combined.R
```diff
@@ -1,4 +1,344 @@
...
+  )
+
+  # The linear scaling is represented either of the following:
+  #
+  # - `"lin"` (in `DefaultPlotConfiguration`)
+  # - `"identity"` (in `tlf::PlotConfiguration`, because of `{ggplot2}`)
+  if (scaling %in% c("lin", "identity")) {
+    pairedData <- dplyr::mutate(pairedData, resValue = predValue - yValues)
+  } else {
+    pairedData <- dplyr::mutate(pairedData, resValue = log(predValue) - log(yValues))
```
Yuri05, line 237: log(x) is ln(x) in R, right?
Seems wrong to me, I would expect `log10(predicted)-log10(observed)` and not ln(predicted)-ln(observed)???

IndrajeetPatil: I am using the same calculations as in the reporting engine:

https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/3a3aa9ac469643a0a50db34ca4b4aafdeae71250/R/utilities-goodness-of-fit.R#L189-L194

But, yes, you are right that `log(x)` in R is `log(x, exp(1))`.

R/utilities-data-combined.R
IndrajeetPatil: Good idea. Done!

R/utilities-data-combined.R
IndrajeetPatil: Oh, and I am using `tlf::Scaling$lin` only for that one entry now.

R/utilities-data-combined.R
IndrajeetPatil: Done.

R/utilities-data-combined.R
Yuri05: should we introduce `tlf::Scaling$identity` then?

R/utilities-data-combined.R
Yuri05: you are right, RE calculates it like this. interesting...

R/utilities-data-combined.R
Yuri05: hmm, it's still optional?

IndrajeetPatil: Oh, this is internal function. I changed it in `calculateResiduals`, which is the exported function.

I guess I can also change it here.

R/utilities-data-combined.R
PavelBal: What kind of scaling is "Identity"?

R/utilities-data-combined.R
IndrajeetPatil: Done.

R/utilities-data-combined.R
IndrajeetPatil: Yes, good idea.

Created issue to track this: https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/368

IndrajeetPatil: `"identity"` is to `{ggplot2}` what `"lin"` is to `{tlf}`.

A linear scale is represented as `"identity"` in `tlf::PlotConfiguration$xAxis$scale`.

This is just for internal purposes. Users won't have to know about it.

Yuri05 (approved):

Issue #1087, opened 2022-08-16
Computing residuals when scaling is logarithmic


IndrajeetPatil:
log(x) is ln(x) in R, right?
Seems wrong to me, I would expect `log10(predicted)-log10(observed)` and not `ln(predicted)-ln(observed)`???

_Originally posted by @Yuri05 in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1085#discussion_r946571034_

Referring to the code here:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/1252a50a62bfb47db7e0d0816f85427ed5386e89/R/utilities-data-combined.R#L234-L238

IndrajeetPatil:
This was done to stay faithful to what is done in the reporting engine:

https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/3a3aa9ac469643a0a50db34ca4b4aafdeae71250/R/utilities-goodness-of-fit.R#L189-L194

But, yes, you are correct that `log(x)` in R is `log(x, exp(1))`, i.e. `ln(x)`. Therefore, the correct way to do this would be using `log10()`.

Yuri05:
@sfrechen @PavelBal Is calculating of logarithmic residuals using **ln** (and not **log10**) in RE correct?

sfrechen:
Natural logarithm is correct. The residuals should follow then a lognormal distribution.

PavelBal:
Then the PK-Sim implementation is not correct?

sfrechen:
Let's say: uncommon. And not **directly** comparable to a lognormal distribution.

Yuri05:
>Let's say: uncommon. And not directly comparable to a lognormal distribution.

Then I wonder, why in the "Histogram of residuals" in RE we plot **normal** distribution all the time??? @sfrechen 


<img src="https://user-images.githubusercontent.com/25061876/70235287-f3a2fb80-1762-11ea-8c0e-9307b79d6ef9.png">

sfrechen:
> Then I wonder, why in the "Histogram of residuals" in RE we plot **normal** distribution all the time???

Well, if the assumed error model is logarithmic, i.e. epsilon follows **log normal distribution** (which is usually the case), then plotting the logarithmized residuals against a **normal distribution** is correct (because `log(residual)` follows then a normal distribution while `residual `follows a log normal distribution)

[...]

PR #1090, opened 2022-08-16
Improvements to residual calculation function

### LINKED ISSUE #1075: plotObservedVsSimulated() - error when observed data time exceeds simulation time
[SimResults.csv](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/files/9310133/SimResults.csv)

With the following simulated and observed data:

```

simData <- importResultsFromCSV(sim, "d:/SimResults.csv")

obsData <- DataSet$new(name = "Observed")
obsData$setValues(xValues = c(1, 3, 3.5, 4, 5), yValues = c(1.9, 6.1, 7, 8.2, 1))
obsData$xUnit <- "min"

myDC <- DataCombined$new()
myDC$addSimulationResults(simData, groups = "myGroup")
myDC$addDataSets(obsData, groups = "myGroup")

plotIndividualTimeProfile(myDC)
```

image

`plotObservedVsSimulated()` fails:

```
> plotObservedVsSimulated(myDC)
Error in rep(seq(1, base - 1, length.out = ticks_per_base), reps) : 
  invalid 'times' argument
In addition: There were 11 warnings (use warnings() to see them)
```

### LINKED ISSUE #1092: Output of the `calculateResiduals()` function
I have some comments/questions about the `calculateResiduals()` function resp. its output.

- Observed values are stored in the colum `yValues`, simulated in `predictedValues`. Could we rename the columns to `yValuesObserved` and `yValuesSimulated`, to make it more obvious?
- Why do we need the columns  `yValuesLower` and `yValuesUpper` in this user-oriented function? I would be (and actually am) confused about what these columns are about.
- The dimensions `Concentration (molar)` and `Concentration (mass)` are substituted by `Concentration`, which actually not a valid dimension and therefore could not be further used by the user. I think this substitution should be performed by the functions that require it, and not in the output.
- I think including the name of the data set could be useful, especially when e.g. multiple observed data are mapped to one simulated data (or vice versa)
- `yErrorType` and `yErrorUnit` is probably also required by the plotting functions?

IndrajeetPatil:
Closes #1092
Closes #1075


R/utilities-data-combined.R
```diff
@@ -240,7 +243,7 @@ calculateResiduals <- function(dataCombined,
   #
   # To avoid this, just remove rows where any of the quantities are `0`s.
   if (scaling %in% c(tlf::Scaling$log, tlf::Scaling$ln)) {
-    pairedData <- dplyr::filter(pairedData, xValues != 0, yValues != 0)
+    pairedData <- dplyr::filter(pairedData, xValues != 0, yValues != 0, predictedValues != 0)
```
Yuri05, line 246: Something else comes in my mind now where I see this:

in PK-Sim/MoBi, log residuals are calculated via the defined there `LogSafe `function:
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fcd42c611764e4af44f266285175ef66f706fdde/src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculator.cs#L118-L124

with
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fcd42c611764e4af44f266285175ef66f706fdde/src/OSPSuite.Core/Domain/Constants.cs#L92

The log residual itself is then calculated as 
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fcd42c611764e4af44f266285175ef66f706fdde/src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculatorForOnlyObservedData.cs#L12-L20

This soves different problems, e.g. if simulated values are slightly negative (numerical zero) log would return NaN; if simulated values are positive but extremely small (numerical zero as well), residual might become infinity (or minus infinity) etc.

So how about implementing residual calculation in R in the same way (just with `ln` instead of `log10`)?

msevestre: Why aren't we implementing the same function in R and ospsuite.core? This does not make sense to me to have two different implementations for the same concept. If the one we have is not wrong, we should stick with this one for R also. I remember that we had long discussions about this in the past (also with @ju-rgen ) and the log10 was the defacto choice

Yuri05: well, seems there are different opinions on this: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1087

msevestre: I saw and the point is that the other implementation is also correct. (@sfrechen says so himself)
I think having two different implementations is simply wrong. We should keep with one to show that we can at least be cohesive within our suite. If that means changing the one in PK-Sim. So be it. But maybe @ju-rgen remembers why this was chosen back then.

msevestre: This was also checked with Roland at the time. The documentation also stipulates log10.... just as an FYI

IndrajeetPatil: Okay, I think then we should also switch to using `log10()` in both ospsuite and reporting engine packages.

I will update the PR for the same.

Yuri05: @IndrajeetPatil Please don't switch to `log10 `until we have the final decision on this.
It was `ln `in RE all the time (also in the previous MAtlab version) and `log10 `all the time in PK-Sim.
I agree that it would be good to have the same algo in all software pieces, but for now I could not give the final answer which of both we should use.

IndrajeetPatil: Okay, in that case, I would appreciate if we don't block the current PR, which is unrelated to the log issue, but rather related to when some predicted values shouldn't be computed.

Yuri05: Agree. I created an issue which can be adressed later: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1091

IndrajeetPatil:
@Yuri05 and @PavelBal Can one of you please approve this? 

This is blocking me from working on a few other things which involve the same files I have changed here.

Yuri05:
I cannot, because some of the code (`dplyr::mutate`) is beyond my understanding. 
@PavelBal please check and approve

PavelBal (approved):

PR #1097, opened 2022-08-19
Use Aciclovir data for testing individual time profile plot

IndrajeetPatil:
To be consistent with other tests.

Additionally, it's a bit simpler to deal with, making it easy to notice when the graphical output changes.

PR #1098, opened 2022-08-19
Simplify profile plot code

IndrajeetPatil:
A lot of this was meant to be just as a cautionary transition from the old to the new API for mapping to multiple datasets.

It is not necessary anymore since most of the wrinkles have already been ironed out.


IndrajeetPatil:
To be considered after #1097 is merged.

PavelBal (approved):
I assume this works as expected :)

Issue #1099, opened 2022-08-19
Clean up test code by saving serialized `R6` objects to `tests/data`


IndrajeetPatil:
Just an idea.

At least, while testing `DataCombined` and `plot*()` functions, the same `DataCombined` objects are being constructed across files. 
All this code distracts from the actual test code.

``` r
library(R6)

Person <- R6Class("Person", list(
  name = NULL,
  age = NA,
  initialize = function(name, age = NA) {
    self$name <- name
    self$age <- age
  },
  print = function(...) {
    cat("Person: \n")
    cat("  Name: ", self$name, "\n", sep = "")
    cat("  Age:  ", self$age, "\n", sep = "")
    invisible(self)
  }
))

indra <- Person$new("Indra", 34)
indra
#> Person: 
#>   Name: Indra
#>   Age:  34

saveRDS(indra, file = "indraObj")

savedIndra <- readRDS("indraObj")

savedIndra
#> Person: 
#>   Name: Indra
#>   Age:  34

waldo::compare(indra, savedIndra)
#> v No differences
```

<sup>Created on 2022-08-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
It's extremely important to also save the script used to create the objects alongside in the `tests/data` folder.

Yuri05:
How about extracting the functions for test objects creation into some `helper-for-tests.R` (or `helper-for-tests-datacombined.R`, `helper-for-tests-XYZ.R`, ...)

IndrajeetPatil:
I have made a proof of concept PR with tests changed only for `plotIndividualTimeProfile()`.

The test file is much cleaner now:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/1099_save_r6/tests/testthat/test-plot-individual-time-profile.R

Whoever is taking over my responsibilities can do the same change for other tests for plotting function, and preferably, also for `test-data-combined.R`

IndrajeetPatil:
> How about extracting the functions for test objects creation into some `helper-for-tests.R`

Not sure about that, since there are no functions involved. We are just creating objects and saving them. And we want to create them just once and keep using them thenceforth.

PavelBal:
@svavil Med prio.

msevestre:
I do not like the idea of serializing the object at all
[...]

PR #1100, opened 2022-08-19
Use saved `DataCombined` objects in `plotIndividualTimeProfile()` tests

IndrajeetPatil:
Related to #1099

Note that I haven't touched anything in the source code (`/R` folder). I have just renamed snapshots, and removed one snapshot test which was redundant.

The test file is much cleaner now:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/1099_save_r6/tests/testthat/test-plot-individual-time-profile.R

Whoever is taking over my responsibilities can do the same change for other tests for plotting function, and preferably, also for `test-data-combined.R`


IndrajeetPatil:
I am marking this as Ready for review.

You can either merge this now; or make the same change for all other plotting functions, and then merge. Up to you. I won't be working on this any further.

PavelBal (approved):

PR #1118, opened 2022-09-13
Add installation script

PavelBal:
A set of scirpts to install/update the osps r packages.

After sourcing the file, the user has to call

`installPackagesGlobally(updatePackages = TRUE, pkSimPath = pkSimPath, suppressOutput = TRUE, developerVersion = FALSE)`

The script will:

- Install required packages from CRAN:
  install.packages(c("R6", "stringr", "readr", "hash", "readxl", "shiny", "shinyjs", "vctrs", "writexl", "dplyr", "tidyr", "ggplot2", "FME", "patchwork"),
                   dependencies = TRUE)
- Check for installation of `rtools` (I am not sure if this is required...)
- Install `rClr` - works only for windows!
- Install `ospsuite.utils`, `tlf`, and `ospsuite` from latest releases. If `developerVersion = TRUE`, latest developer versions from AppVeyor will be installed
- If `updatePackages = TRUE`, try to update ALL packages in the library.
- Test if connection to PK-Sim can be established (`initPKSim`)

Known issues:
- When starting a fresh R session and trying to install some packages, R asks for "Some packages are in use, want to restart R session?". Actually this should not happen as nothing is loaded, but it happens. If the user clicks "OK restart", the script breaks... so "No" is the correct answer here. Don't know yet how to overcome this.
- When a version of ospsuite is already installed, sometimes the script could not update the packages because it was loaded and it could not remove the lock from DLLs... Manually deleting the package before running the script solves the problem, and I have no idea why.
- No Linux support


svavil:
`rtools` is required to install `rClr`, so it's a required step in this process. Should we try removing the `ospsuite` package each time before installing it?

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,163 @@
...
+#' @return
+#' @export
+#'
+#' @examples
+installOSPPackages <- function(rtoolsPath = NULL, rclrVersion = "0.9.2",
+                               suppressOutput = TRUE,
+                               developerVersion = FALSE){
+  # Install dependencies from CRAN
+  displayProgress("Installing CRAN packages", suppressOutput = suppressOutput)
+  install.packages(c("R6", "stringr", "readr", "hash", "readxl", "shiny", "shinyjs", "vctrs", "writexl", "dplyr", "tidyr", "ggplot2", "FME", "patchwork"),
```
PavelBal, line 82: Installing CRAN packages

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,163 @@
...
+                   dependencies = TRUE)
+
+  displayProgress("Checking RTOOLS", suppressOutput = suppressOutput)
+  if (Sys.which("make") == "") { # rtools is not found
```
PavelBal, line 86: Check rtools

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,163 @@
...
+      displayProgress("Checking RTOOLS", success = FALSE, message = packageInstallationMessages$installRTools, suppressOutput = suppressOutput)
+      return()
+    }
+  }
+  if (Sys.which("make") == "") {
+    displayProgress("Checking RTOOLS", success = FALSE, message = packageInstallationMessages$RToolsNotFound, suppressOutput = suppressOutput)
+    return()
+  }
+
+  displayProgress("Installing rClr", suppressOutput = suppressOutput)
```
PavelBal, line 99: Install rClr - currently only for Windows

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,163 @@
...
+  if (version$major == "3") {
+    install.packages(paste0("https://github.com/Open-Systems-Pharmacology/rClr/releases/download/v", rclrVersion, "-R3/rClr_", rclrVersion, ".zip"), repos = NULL)
+  }
+
+  packagePaths <- .releasePaths
+  if (developerVersion) {
+    packagePaths <- .developPaths
+  }
+
+  displayProgress("Installing ospsuite.utils", suppressOutput = suppressOutput)
```
PavelBal, line 112: Install osp packages

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,163 @@
...
+installPackagesGlobally <- function(updatePackages = TRUE, pkSimPath = NULL,
+                                    rtoolsPath = NULL, rclrVersion = "0.9.2",
+                                    suppressOutput = TRUE,
+                                    developerVersion = FALSE){
+  installOSPPackages(rtoolsPath = rtoolsPath, rclrVersion = rclrVersion,
+                     suppressOutput = suppressOutput,
+                     developerVersion = developerVersion)
+
+  # Update all installed packages from CRAN
+  if (updatePackages) {
```
PavelBal, line 142: This will update all CRAN packages in the library.

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,163 @@
...
+    update.packages(ask = FALSE)
+  }
+
+  displayProgress("Testing PK-Sim connection", suppressOutput = suppressOutput)
+  flagConnection <- FALSE
+  try(flagConnection <- testPKSIMConnection(pkSimPath = pkSimPath))
```
PavelBal, line 148: Try to load PK-Sim dll

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,177 @@
...
+  RToolsNotFound = function(rToolsPath) {
+    paste0("Rtools not found at ", rToolsPath, ", cannot continue")
+  },
+  PKSimLoadFails = "PK-Sim fails to load. The installation might be incompatible with your current version of PK-Sim"
+)
+
+# Download paths of released package versions
+.releasePaths <- list(
+  ospsuite.utils = "https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/releases/download/v1.3.17/ospsuite.utils_1.3.17.zip",
+  tlf = "https://github.com/Open-Systems-Pharmacology/TLF-Library/releases/download/v1.4.89/tlf_1.4.89.zip",
```
Yuri05, line 13: let's use .tar.gz instead of zip where possible.
a) packages which are not OS-specific (like tlf or ospsuite.utils) can be installed under Linux then without problems
b) tests can be installed on demand

Regarding the last point: I would convert this script into **function** and provide optional argument for the installation of the package tests

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,177 @@
...
+  ospsuite = "https://github.com/Open-Systems-Pharmacology/OSPSuite-R/releases/download/v11.0.123/ospsuite_11.0.123.zip"
```
Yuri05, line 14: ospsuite package must be selected OS-dependent

inst/extdata/install-packages.R
Yuri05: I miss the following packages:
* **purrr**
* **jsonlite**
* **knitr**
* **rmarkdown**

On the other side, why are the packages bellow installed? None of them is in the dependencies list of ospsuite.utils, tlf or ospsuite-R
* **hash**
* **shinyjs**
* **vctrs**
* **writexl**
* **FME**

inst/extdata/install-packages.R
Yuri05: This must be extendened for all currently supported systems (Ubuntu18, CentOS7)

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,177 @@
...
+  if (developerVersion) {
+    packagePaths <- .developPaths
+  }
+
+  displayProgress("Installing ospsuite.utils", suppressOutput = suppressOutput)
+  install.packages(packagePaths$ospsuite.utils, repos = NULL)
+  displayProgress("Installing tlf", suppressOutput = suppressOutput)
+  install.packages(packagePaths$tlf, repos = NULL)
+  displayProgress("Installing ospsuite", suppressOutput = suppressOutput)
+  install.packages(packagePaths$ospsuite, repos = NULL)
```
Yuri05, line 129: here, all osp packages should be installed with tests (if the option was set by user), e.g.
`install.packages(packagePaths$ospsuite.utils, repos = NULL, INSTALL_opts = "--install-tests")`

inst/extdata/install-packages.R
svavil: `hash` and `FME` are (currently) required for parameter identification

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,177 @@
...
+#' the path is deduced from system environment variables.
+#' @param rclrVersion Version of rClr package. Default is 0.9.2 for Windows R4.
+#' @param developerVersion If `FALSE` (default), release verions of the packages
+#' will be installed. If `TRUE`, latest developer builds of the osps packages
+#' will be installed
+#' @param suppressOutput
+installPackagesGlobally <- function(updatePackages = TRUE, pkSimPath = NULL,
+                                    rtoolsPath = NULL, rclrVersion = "0.9.2",
+                                    suppressOutput = TRUE,
+                                    developerVersion = FALSE) {
```
Yuri05, line 148: What is this function good for and why is `installOSPPackages ` not sufficient?
Also I don't see any differences between these 2 functions in terms where the packages are installed (globally or not globally)

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,177 @@
...
+#' will be installed. If `TRUE`, latest developer builds of the osps packages
+#' will be installed
+#' @param suppressOutput
+#'
+#' @return
+#' @export
+#'
+#' @examples
+installOSPPackages <- function(rtoolsPath = NULL, rclrVersion = "0.9.2",
+                               suppressOutput = TRUE,
```
Yuri05, line 89: I would prefer the default setting `suppressOutput = FALSE,`

inst/extdata/install-packages.R
Yuri05: ok, but parameter identification is standalone package and not (yet) part of the ospsuite package.
Would suggest to remove them.

inst/extdata/install-packages.R
Yuri05: would prefer the default setting `updatePackages = FALSE`

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,177 @@
...
+  if (updatePackages) {
+    update.packages(ask = FALSE)
+  }
+
+  displayProgress("Testing PK-Sim connection", suppressOutput = suppressOutput)
+  flagConnection <- FALSE
+  try(flagConnection <- testPKSIMConnection(pkSimPath = pkSimPath))
+  if (!flagConnection) {
+    stop(message = packageInstallationMessages$PKSimLoadFails)
+  }
```
Yuri05, line 165: `testPKSIMConnection `should be executed on Windows only

Yuri05: Also in case of failed connection I would produce just a warning, not error.
Otherwise it looks for the user like the whole installation was failed - which is in fact not the case!

msevestre:
Quick question. Can the user decide NOT to install anything? E.g. same as before?

inst/extdata/install-packages.R
PavelBal: - purrr is probably some dependency from other package....
- jsonlite is not a dependency of ospsuite
- **knitr** and **markdown** are only required for building vignettes but not for simply install and work with the package
- I removed the unnecessary packages from installation

Yuri05: let's install **purrr** explicitely. Otherwise if it will be not a dependnecy of other CRAN packages one day - the installation script will become invalid.
For the other packages - agree. They are used by the `ospsuite.reportingengine` and must be installed there, not as part of ospsuite-R.

inst/extdata/install-packages.R
PavelBal: Done.

inst/extdata/install-packages.R
PavelBal: Completely removed `installPackagesGlobally()`, it is redundand with `installOSPPackages()`

inst/extdata/install-packages.R
PavelBal: Is now an optional task that can be triggered by the user.

PavelBal:
> Quick question. Can the user decide NOT to install anything? E.g. same as before?

Erm.. he just not executes the function? :D

PavelBal:
@Yuri05 I have adressed the following points:

- Add "purr" explicitly to be installed from CRAN
- Remove "installPackagesGlobally", only use "installOSPPackages"
- installOSPPackages - argument "suppressOutput" defaults changed to "FALSE"
- testPKSimConnection must be executed separately

What remains is to extend the script for installation on Linux. As we do not have a Linux installation (and therefore no experience), this task should be taken over by someone else.

Yuri05:
just checked, that **jsonlite** is a prereq for **tlf** (and not only for RE), so it must be installed here

inst/extdata/install-packages.R
```diff
@@ -0,0 +1,146 @@
...
+}
+
+# Remove functions created by this script from the environment
+cleanEnvironment <- function() {
+  rm(
+    packageInstallationMessages,
+    .releasePaths,
+    .developPaths,
+    testPKSIMConnection,
+    cleanEnvironment,
```
Yuri05, line 47: suicide :)

inst/extdata/install-packages.R
Yuri05: issue created

inst/extdata/install-packages.R
Yuri05: issue created

Yuri05:
@PavelBal OK, I created additional issues for Linux, tests etc.
After you add **jsonlite** - can be merged from my point of view. The remaining issues can be addressed in the future PRs

PavelBal:
@Yuri05 Added jsonlite

Yuri05 (approved):

PR #1155, opened 2022-10-20
`plotObservedVsSimulated` fold distance cannot be <1

### LINKED ISSUE #1146: `plotObservedVsSimulated` fold distance should not be <1
Do not allow values below 1,  throw an error.

Improve the documentation of the funnction, explaining what the parameter means.

PavelBal:
Fixes #1146

### DIFF
--- a/R/plot-observed-vs-simulated.R
+++ b/R/plot-observed-vs-simulated.R
@@ -2,11 +2,11 @@
 #'
 #' @inheritParams plotIndividualTimeProfile
 #' @param foldDistance A vector for plotting lines at required fold distances
-#'   The vector can include only fold distance values different from `>=1`. An
+#'   The vector can include only fold distance values `>1`. An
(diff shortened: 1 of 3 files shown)


R/plot-observed-vs-simulated.R
```diff
@@ -81,6 +84,11 @@ plotObservedVsSimulated <- function(dataCombined,
   # - For logarithmic scale: `0`
   defaultFoldDistance <- ifelse(is_any_scale_linear, 0, 1)
 
+  # foldDistance should not be lower than 1
+  if (any(foldDistance < 1)) {
+    stop(messages$plotObservedVsSimulatedWrongFoldDistance("foldDistance", foldDistance))
```
Yuri05, line 89: I think fold distance = 1 does not make sense as well

PavelBal: Restricted to >1.

Yuri05 (approved):

PR #1161, opened 2022-11-16
Make use of safe log for residuals calculation

### LINKED ISSUE #1160: Add `LogSafe` function for residuals calculation.

PavelBal:
Fixes #1160

### DIFF
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -324,6 +324,10 @@ DataCombined <- R6::R6Class(
     #'   scale factors to apply to raw values. The default offset is `0`, while
     #'   default scale factor is `1`, i.e., the data will not be modified. If a
     #'   list is specified, it should be the same length as `forNames` argument.
+    #' @param reset IF `TRUE`, only data transformations that are specified will
+    #'   be retained. Not specified transformations will be reset to their defaults.
+    #'   Default behavior is `FALSE`, e.g., setting only `xOffsets` will not reset
+    #'   `xScaleFactors` if those have been set previously.
     #'
     #' @details
     #'
--- a/R/ospsuite-env.R
+++ b/R/ospsuite-env.R
@@ -54,6 +54,9 @@ ospsuiteEnv$isPKSimLoaded <- FALSE
 # NetTask `DimensionTask` cached for performance benefits. Created the first time it is requested.
 ospsuiteEnv$dimensionTask <- NULL
 
+# Small value added to zero when calculating log
+ospsuiteEnv$LOG_SAFE_EPSILON <- 1e-20
+
 #' Names of the settings stored in ospsuiteEnv. Can be used with `getOSPSuiteSetting()`
 #' @include utilities.R
 #' @export
--- a/R/plot-observed-vs-simulated.R
+++ b/R/plot-observed-vs-simulated.R
@@ -115,6 +115,16 @@ plotObservedVsSimulated <- function(dataCombined,
     return(NULL)
   }
 
+  # In logarithmic scale, if any of the values are `0`, plotting will fail.
+  #
+  # To avoid this, just remove rows where any of the quantities are `0`s.
+  if (obsVsPredPlotConfiguration$yAxis$scale %in% c(tlf::Scaling$log, tlf::Scaling$ln)) {
+    pairedData <- dplyr::filter(
+      pairedData,
+      yValuesObserved != 0, yValuesSimulated != 0
+    )
+  }
+
   # Add minimum and maximum values for observed data to plot error bars
   pairedData <- dplyr::mutate(
     pairedData,
--- a/R/utilities-data-combined.R
+++ b/R/utilities-data-combined.R
@@ -214,12 +214,18 @@ calculateResiduals <- function(dataCombined,
       observedData$xValues
     )$y
   )
-
   # Residual computation will depend on the scaling.
   if (scaling %in% c(tlf::Scaling$lin, tlf::Scaling$identity)) {
     pairedData <- dplyr::mutate(pairedData, residualValues = yValuesSimulated - yValuesObserved)
   } else {
-    pairedData <- dplyr::mutate(pairedData, residualValues = log(yValuesSimulated) - log(yValuesObserved))
+    # Epsilon for safe log calculation should be converted to the units of the values
+    epsilon <- toUnit(
+      quantityOrDimension = pairedData$yDimension[[1]],
+      values = ospsuiteEnv$LOG_SAFE_EPSILON,
+      targetUnit = pairedData$yUnit[[1]],
+      molWeight = 1
+    )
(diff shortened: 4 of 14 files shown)


R/data-combined.R
```diff
@@ -324,6 +324,10 @@ DataCombined <- R6::R6Class(
     #'   scale factors to apply to raw values. The default offset is `0`, while
     #'   default scale factor is `1`, i.e., the data will not be modified. If a
     #'   list is specified, it should be the same length as `forNames` argument.
+    #' @param reset IF `TRUE`, only data transformations that are specified will
```
PavelBal, line 327: @svavil Documentation of the new argument was missing.

R/ospsuite-env.R
```diff
@@ -54,6 +54,9 @@ ospsuiteEnv$isPKSimLoaded <- FALSE
 # NetTask `DimensionTask` cached for performance benefits. Created the first time it is requested.
 ospsuiteEnv$dimensionTask <- NULL
 
+# Small value added to zero when calculating log
+ospsuiteEnv$LOG_SAFE_EPSILON <- 1e-20
```
PavelBal, line 58: Using the same value as in Core

R/plot-observed-vs-simulated.R
```diff
@@ -115,6 +115,16 @@ plotObservedVsSimulated <- function(dataCombined,
     return(NULL)
   }
 
+  # In logarithmic scale, if any of the values are `0`, plotting will fail.
+  #
+  # To avoid this, just remove rows where any of the quantities are `0`s.
+  if (obsVsPredPlotConfiguration$yAxis$scale %in% c(tlf::Scaling$log, tlf::Scaling$ln)) {
```
PavelBal, line 121: This was previously done during residuals calculation, but is only relevant for plotting.

R/utilities-data-combined.R
```diff
@@ -214,12 +214,18 @@ calculateResiduals <- function(dataCombined,
...
     )$y
   )
-
   # Residual computation will depend on the scaling.
   if (scaling %in% c(tlf::Scaling$lin, tlf::Scaling$identity)) {
     pairedData <- dplyr::mutate(pairedData, residualValues = yValuesSimulated - yValuesObserved)
   } else {
-    pairedData <- dplyr::mutate(pairedData, residualValues = log(yValuesSimulated) - log(yValuesObserved))
+    # Epsilon for safe log calculation should be converted to the units of the values
+    epsilon <- toUnit(
```
PavelBal, line 222: The epsilon should only be valid for values in base unit. So have to convert it here to the actually used unit.

R/utilities-data-combined.R
```diff
@@ -214,12 +214,18 @@ calculateResiduals <- function(dataCombined,
...
+      quantityOrDimension = pairedData$yDimension[[1]],
+      values = ospsuiteEnv$LOG_SAFE_EPSILON,
+      targetUnit = pairedData$yUnit[[1]],
+      molWeight = 1
+    )
+    pairedData <- dplyr::mutate(pairedData, residualValues = .log_safe(yValuesSimulated, epsilon = epsilon) - .log_safe(yValuesObserved, epsilon = epsilon))
```
PavelBal, line 228: Using log_safe now.

R/utilities-data-combined.R
```diff
@@ -229,40 +235,30 @@ calculateResiduals <- function(dataCombined,
...
-  x <- ospsuite.utils::toMissingOfType(x, type = "double")
-
-  if (is.na(x)) {
-    return(NA_real_)
-  } else if (x < epsilon) {
-    return(log(epsilon, base = base))
-  } else {
-    return(log(x, base = base))
-  }
+.log_safe <- function(x, base = 10, epsilon = ospsuiteEnv$LOG_SAFE_EPSILON) {
```
PavelBal, line 249: Also now using log_10 and not ln to produce the same results as pksim/mobi

R/utilities-data-combined.R
```diff
@@ -229,40 +235,30 @@ calculateResiduals <- function(dataCombined,
...
+  x <- sapply(X = x, \(element){
```
PavelBal, line 250: Also vectorized the function.

tests/testthat/test-utilities-data-combined.R
```diff
@@ -6,15 +6,15 @@ simFilePath <- system.file("extdata", "Aciclovir.pkml", package = "ospsuite")
 sim <- loadSimulation(simFilePath)
 simData <- withr::with_tempdir({
   df <- dplyr::tibble(
-    IndividualId = c(0, 0, 0),
-    `Time [min]` = c(0, 2, 4),
-    `Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood) [µmol/l]` = c(0, 4, 8)
+    IndividualId = c(0, 0, 0, 0),
```
PavelBal, line 9: Extended the tests to account for the new behavior for log_safe(0)

R/plot-observed-vs-simulated.R
msevestre, line 121: can it be NOT in log or ln?

R/ospsuite-env.R
msevestre: The plan it to use the one from cor eventually right? (the logsafe function)

R/utilities-data-combined.R
msevestre: hum..good point

R/utilities-data-combined.R
```diff
@@ -229,40 +235,30 @@ calculateResiduals <- function(dataCombined,
...
+    element <- ospsuite.utils::toMissingOfType(element, type = "double")
+    if (is.na(element)) {
+      return(NA_real_)
+    } else if (element < epsilon) {
```
msevestre, line 254: you don't need all the else here (just FYI no need to chnagE)

msevestre (commented):
Looks good to me
I just have the feeling that I saw somewhere a method asking to use the method from .net. Is this temporary or what's the difference?

R/plot-observed-vs-simulated.R
PavelBal: It can be linear `tlf::Scaling$lin`.

PavelBal:
> The plan it to use the one from cor eventually right? (the logsafe function)

I don't think we should make .NET calls for such simple operation. If logsafe were more complex, yes, but here I would prefer to have this function implemented in R.

PavelBal:
> I just have the feeling that I saw somewhere a method asking to use the method from .net. Is this temporary or what's the difference?

We discussed about using residuals calculation from Core. Still under discussion, I just wanted to maker sure that the current implementation is robust.

PavelBal:
@msevestre The build is passing, so feel free to merge if you don't have any other remarks.

PR #1166, opened 2022-12-01
Update core to 11.1.65

### LINKED ISSUE #1165: Package installation fails under R3
This code seems to be unsupported under R3:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/5deaf3696f85562ee726b97080e3908ecdddfc6a/R/utilities-data-combined.R#L247-L262

When trying to execute the code on Windows (R 3.6.3):
```
> .log_safe <- function(x, base = 10, epsilon = 1e-20) {
+     x <- sapply(X = x, \(element){
Error: unexpected input in:
".log_safe <- function(x, base = 10, epsilon = 1e-20) {
    x <- sapply(X = x, \"
```

When trying to install under Ubuntu (R 3.6.3)
grafik

PavelBal:
Remove usage of R4 anonymous function syntax (Fixes #1165)


R/utilities-data-combined.R
```diff
@@ -247,7 +247,7 @@ calculateResiduals <- function(dataCombined,
 #' @keywords internal
 #' @noRd
 .log_safe <- function(x, base = 10, epsilon = ospsuiteEnv$LOG_SAFE_EPSILON) {
-  x <- sapply(X = x, \(element){
+  x <- sapply(X = x, function(element){
```
Yuri05, line 250: I must say, for me `function(element) `is so much clearer than `\(element)`...

Yuri05 (approved):

PR #1173, opened 2023-01-04
Fixes #1171 nightly build fails

### LINKED ISSUE #1171: Nightly builds fail
Since couple of days nigthly builds started failing (without any code changes)

grafik

rwmcintosh:


rwmcintosh:
This works because R version 4.2 has dropped support for 32-bit arch. The error in the build is caused by the fact that our testing step is checking that the package can be installed in both 64 and 32 bit varieties (under R 4.1.1)

I just updated the R version used for the checks, and updated rClr to match.

rwmcintosh:
If acceptable, I'll update the nightly yml file too. Is there an appveyor job that runs R version 3? Could we remove this appveyor file if not?

svavil:
I have some reservations about this, because we have been using `rClr` 0.9.1 (this specific version) in scripts that set up the workflow, and other versions of `rClr` were conflicting with other packages. I don't mind switching to `R` 4.2.2, but switching to `rClr` 0.9.2 will require more attention. 

As other ways of solving the same problem, we should try specifying in the `appveyor.yml` to only test for `x64` architecture and not `i386`, or installing an earlier version of `purrr` with `i386` support. If I am reading the control flow correctly, `purrr` is currently installed by calling `travis-tool install_deps` in `appveyor.yml`, which in turns calls [`devtools::install_deps()`](https://rdrr.io/cran/devtools/man/install_deps.html), which looks into the `DESCRIPTION` file to see what packages (and their versions) are needed.

rwmcintosh:
@svavil Let's fix the purrr revision tomorrow together. I'm just a little out of my depth in the R world.
I think it's also beneficial if we can somehow disable 32 bit testing.

PavelBal:
rClr 0.9.2. is the same as 0.9.1. but built with r 4.2. safe to use with
all r4 versions.

Robert McIntosh ***@***.***> schrieb am Mi., 4. Jan. 2023,
16:48:

rwmcintosh:
rClr 0.9.2 should be marked as a release if it's ready for that - currently labeled pre-release

https://github.com/Open-Systems-Pharmacology/rClr/releases/tag/v0.9.2

What I have done now is reverted to R 4.1.1 and rClr 0.9.1. The latest purrr will be used from CRAN, with only 64-bit version at the time of writing this. Even if they release a 32-bit version in the future we won't be testing it in our CI anymore.

So, this can be merged as-is if the default is to keep using the most up-to-date packages.

_This change only disables the 32-bit installation check._

appveyor.yml
```diff
@@ -23,7 +23,8 @@ environment:
   USE_RTOOLS: true
   NOT_CRAN: true
   KEEP_VIGNETTES: false
-  R_BUILD_ARGS: --no-build-vignettes --no-manual"
```
rwmcintosh, line 26: This line difference is only the quote at the end. I think it was there accidentally before.

rwmcintosh:
Should this appveyor-3.6.yml be discontinued? Or should it be more similar to the appveyor.yml file with only a difference in the R version? 

It hasn't been built successfully in quite some time.

msevestre:
all of the differences are normally to make it work for 3.6. We just want to run the tests I think

PR #1204, opened 2023-02-06
Improve performance of toUnit by exiting early

### LINKED ISSUE #1203: `toUnit()` - exit early
To improve performance

PavelBal:
Fixes #1203

Some benchmarks:

```
# Same unit without source specified
# user  system elapsed
# 41.18    1.75   74.96
# New version
# 14.75    1.21   18.73

# Same unit with source specified
# user  system elapsed
# 39.67    0.74  107.83

# New version
# 15.42    1.35   18.58
```

When both units are specified and conversion actually happens, performance is almost the same (even better in some cases, though I don't really know why).

### DIFF
--- a/R/utilities-units.R
+++ b/R/utilities-units.R
@@ -95,38 +95,22 @@ getBaseUnit <- function(dimension) {
 #' @export
 toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL, molWeightUnit = NULL) {
   validateIsOfType(quantityOrDimension, c("Quantity", "character"))
-  validateIsNumeric(values, nullAllowed = TRUE)
-  validateIsNumeric(molWeight, nullAllowed = TRUE)
-  unit <- .encodeUnit(unit)
+
+  # Get the base unit of the dimension and call `toUnit()`
   dimension <- quantityOrDimension
-  dimensionTask <- .getNetTask("DimensionTask")
-
-  # covers all NULL or NA
-  if (all(is.na(values))) {
-    return(values)
-  }
-
-  # ensure that we are dealing with an list of values seen as number (and not integer)
-  values <- as.numeric(c(values))
-
   if (isOfType(quantityOrDimension, "Quantity")) {
     dimension <- quantityOrDimension$dimension
   }
+  baseUnit <- getBaseUnit(dimension)
 
-  if (all(is.na(molWeight))) {
-    molWeight <- NULL
-  }
-
-
-  if (is.null(molWeight)) {
-    return(rClr::clrCall(dimensionTask, "ConvertToBaseUnit", dimension, unit, values))
-  }
-
-  # Convert molWeight value to base unit if a unit is provided
-  if (!is.null(molWeightUnit)) {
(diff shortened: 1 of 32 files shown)


R/utilities-units.R
```diff
@@ -95,38 +95,22 @@ getBaseUnit <- function(dimension) {
 #' @export
 toBaseUnit <- function(quantityOrDimension, values, unit, molWeight = NULL, molWeightUnit = NULL) {
```
PavelBal, line 96: Lot's of code duplication in `toBaseUnit()` - I have changed it to using `toUnit()` (which now also exists earlier if the target unit equals the source unit)

R/utilities-units.R
```diff
@@ -95,38 +95,22 @@ getBaseUnit <- function(dimension) {
...
-  }
-
-  # Convert molWeight value to base unit if a unit is provided
-  if (!is.null(molWeightUnit)) {
-    molWeight <- rClr::clrCall(dimensionTask, "ConvertToBaseUnit", ospDimensions$`Molecular weight`, molWeightUnit, molWeight)
-  }
-  rClr::clrCall(dimensionTask, "ConvertToBaseUnit", dimension, unit, values, molWeight)
+  baseUnit <- getBaseUnit(dimension)
+
+  toUnit(
```
PavelBal, line 106: Just get the base unit of the dimension and call `toUnit()`

R/utilities-units.R
```diff
@@ -163,41 +147,58 @@ toUnit <- function(quantityOrDimension,
   validateIsOfType(quantityOrDimension, c("Quantity", "character"))
   validateIsNumeric(values, nullAllowed = TRUE)
   validateIsNumeric(molWeight, nullAllowed = TRUE)
-  targetUnit <- .encodeUnit(targetUnit)
```
PavelBal, line 166: Just moved them below so these calls are not made unless required.

R/utilities-units.R
msevestre: Nice

PavelBal:
@msevestre Could you please finish the review then, if you already started :)

msevestre:
Yep. I am on my phone and I need a bigger screen. Coming soon

R/utilities-units.R
```diff
@@ -163,41 +147,58 @@ toUnit <- function(quantityOrDimension,
...
 
+  dimension <- quantityOrDimension
   if (isOfType(quantityOrDimension, "Quantity")) {
     dimension <- quantityOrDimension$dimension
   }
+  baseUnit <- getBaseUnit(dimension)
+
+  # Return early
+  # If no source unit is defined and target is the base unit
+  if (is.null(sourceUnit) && targetUnit == baseUnit) {
```
msevestre, line 175: can sourceUnit be null?

msevestre: oh yeah. Interesting

PavelBal: > If `NULL` (default), the values are assumed to be in base unit.

msevestre (approved):
Who would have thoughts that this would be a bottle neck!?

PR #1205, opened 2023-02-06
1176 Decoupling the R package from PK-Sim

georgeDaskalakis:


Yuri05:
@georgeDaskalakis Is this still WIP? Otherwise I would have some questions...

rakefile.rb
```diff
@@ -109,6 +114,52 @@ def download_file(project_name, file_name, uri)
...
+  
+  Utils.run_cmd('7z', %W[e #{portable_zip_package}])
+
+  pksim_minimal_dir = "C:/projects/ospsuite-r/pksim_minimal.zip"
+  command_line = %W[
+    a 
+    #{pksim_minimal_dir} 
+    Castle.Core.dll 
+    Castle.Windsor.dll 
+    DevExpress.Data.v21.2.dll
```
Yuri05, line 134: can we use wildcards here (`DevExpress.Data.v*.dll `) in order to avoid adjusting the version in the future?

georgeDaskalakis:
> @georgeDaskalakis Is this still WIP? Otherwise I would have some questions...
@Yuri05 yeap this is ready for review :D

rakefile.rb
msevestre: What is this?!? It's not what we discussed @georgeDaskalakis

georgeDaskalakis: @Yuri05 wildcard added.
@msevestre quick call?

msevestre: I think it's good to document the rationale here so that we can discuss. We said we wanted to upload the zip as a separate package. Instead, we are doing something else. Why

georgeDaskalakis: But we are uploading as a seperate file no? The only difference here is that we are extracting the PKSim portable to create a new zip that has even less files (PKSim minimal) and that is what we are providing as an artifact.

appveyor-nightly.yml
```diff
@@ -65,6 +65,8 @@ test_script:
 after_test:
   - rake "create_linux_build[%APPVEYOR_BUILD_VERSION%, %APPVEYOR_BUILD_FOLDER%, ubuntu18]"
   - rake "create_linux_build[%APPVEYOR_BUILD_VERSION%, %APPVEYOR_BUILD_FOLDER%, centOS7]"
+  - rake "download_portable[develop]"
+  - appveyor PushArtifact "C:/projects/ospsuite-r/pksim_minimal.zip"
```
Yuri05, line 69: I don't quite understand the procedure. 
I thought, the required PK-Sim DLLs would be added to the R package (in the same way as the OSPSuite.Core DLLs are added). Instead, they are combined into **pksim_minimal.zip** and pushed as an additional artifact on AppVeyor?

msevestre: The first idea was to specify the portable PKsim associated with the build so that it could be downloaded as artefacts and even used

msevestre: Right now we create a folder that has no meaning whatsoever
Then, if this is still too cumbersome, we said we would create another package with all the DLLs. 

I don't think we said we would update our standard package with this

Yuri05: >we said we would create another package with all the DLLs.
I don't think we said we would update our standard package with this

Fine for me - less testing :)

PavelBal:
We want to provide a "minimal" version of PK-Sim - i.e., all the files (DB and DLLs) that are required to run all the functionalities of the ospsuite-r. This minimal version is to be provided as a zip together with the build so the user always can download the correct version together with the package. This is how I understood our discussion and also what I see here.

@msevestre please correct me if I understood the discussion wrong. The solution with the "minimal" PK-Sim provided with each ospsuite-r build is a good compromise from my perspective.

msevestre:
It is not for me because it requires changing the ospsuitr-r build everytime there is a new dependency in PKSim. That's why I would simply upload the portable version (all of it)

PavelBal:
@msevestre @Yuri05  Any objection to include this also to the default AppVeyor build (this PR only adds it to the nightly)?

msevestre:
We do not build the package on every build

PavelBal (approved):

PR #1211, opened 2023-02-20
With multiple data sets and groups, legend symbols are overriden

### LINKED ISSUE #1206: `plotIndividualTimeProfile` - wrong legend symbols
```
manyObsDC <- readRDS(getTestDataFilePath("manyObsDC"))
manyObsDC$removeGroupAssignment(names = "Vergin 1995.Iv")

dataSet3 <- DataSet$new(name = "Data set 3")
dataSet3$setValues(xValues = 1, yValues = 1, yErrorValues = 1)
dataSet4 <- DataSet$new(name = "Data set 4")
dataSet4$setValues(xValues = 1, yValues = 1, yErrorValues = 1)

manyObsDC$addDataSets(c(dataSet3, dataSet4))

plotIndividualTimeProfile(manyObsDC)
```

image

svavil:
Fixes #1206

### DIFF
--- a/R/plot-individual-time-profile.R
+++ b/R/plot-individual-time-profile.R
@@ -159,8 +159,13 @@ plotIndividualTimeProfile <- function(dataCombined,
     plotConfiguration = timeProfilePlotConfiguration
   )
 
+  # If there are multiple data sets and multiple groups, shapes in the color legend have to be overriden
+  uniqueShapesForGroups <- dplyr::distinct(obsData[c("name", "group")])
+  uniqueShapesForGroups$shape <- tlf:::.getAestheticValuesFromConfiguration(n = nrow(uniqueShapesForGroups), plotConfigurationProperty = profilePlot$plotConfiguration$points, propertyNames = "shape")$shape
+  uniqueShapesForGroups <- uniqueShapesForGroups[!duplicated(uniqueShapesForGroups$group), ]
+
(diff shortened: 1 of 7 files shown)


Felixmil (commented):
I think this should be implemented in {tlf} instead. Here we can see that the proposed changes have unwanted side effects on the legends:

| Before | After |
|--------|-------|
|image|image|
|image|image|

PavelBal:
> I think this should be implemented in {tlf} instead. Here we can see that the proposed changes have unwanted side effects on the legends:

Could you try to fix it in TLF, then?

Felixmil:
> Could you try to fix it in TLF, then?

@PavelBal is is actually related to the way the default dataMapping is constructed in OSPSR for this kind of plot. More info / discussions in [original issue](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1206#issuecomment-1486454900)

PavelBal:
Why closed? Is this fixed? @svavil

svavil:
@PavelBal my understanding was that we are not going to push these particular changes due to side effects (shaded regions on markers), and we need another approach here? The issue #1206 that was the reason for the PR is still there, and if we come up with a working solution, we could then close the issue, but not with this PR. I might be missing something?

PavelBal:
Oh sorry, I thought it es an issue and not a PR. All good then.

Issue #1112, opened 2022-08-30
`plotResidualsVsSimulated`, `plotResidualsVsTime`: argument for residuals type


PavelBal:
The functions `plotResidualsVsSimulated`, `plotResidualsVsTime` should have the `scaling` argument identical to that used in `calculateResiduals`. Otherwise, it is not straightforward how to select linar or log residuals calculation. This is currently only possible by using a custom `DefaultPlotCongiruation` with x-axis scaling set to `log` or lin. But 1) this is not intuitive and 2) in log scaling it would plot the x-axis in log even if the user does not want it...

PavelBal:
https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/332

PavelBal:
Update the vignettes once implemented.

PR #1213, opened 2023-03-02
Implement #1112

Felixmil:
For #1112

NEW:
- feature that add scaling for residuals (lin or log)
- unit tests

UPDATED:
- Documentation 
- Vignette

I didn't know if I had to render the whole pkgdown website, in double, I just re-rendered the impacted vignette and commited the `.html` file and new `.png` only.


R/plot-residuals-vs-simulated.R
```diff
@@ -43,20 +44,25 @@
 #' myPlotConfiguration$caption <- "My Sources"
 #'
 #' # plot
-#' plotResidualsVsSimulated(myDataCombined, myPlotConfiguration)
+#' plotResidualsVsSimulated(myDataCombined, scaling = "log", defaultPlotConfiguration = myPlotConfiguration)
 #'
 #' @export
 plotResidualsVsSimulated <- function(dataCombined,
-                                     defaultPlotConfiguration = NULL) {
+                                     defaultPlotConfiguration = NULL,
```
Yuri05, line 51: why is it actually called **default**PlotConfiguration and not just plotConfiguration?

PavelBal: Because `PlotConfiguration` is already used in `TLF`...

PavelBal:
@Felixmil You must run the tests locally and accept the changed snapshot (if they should be accepted...) and then push the updated snapshots, otherwise, the built will fail.

R/plot-residuals-vs-simulated.R
Yuri05: ?? it's a function argument, you can call it whatever you want.
```

plotResidualsVsSimulated <- function(dataCombined,
                                     plotConfiguration = NULL,
                                     scaling="lin") {
```

Not a big deal, we can keep it AS IS. Just wondering.

PavelBal: Ah no I understand, you mean the argument name. I prefer to have the class name so it is clear why type is required.

PavelBal:
@Yuri05 @msevestre What kind of sorcery is this? The latest TLF version from the `develop` branch is 1.5.124:

image

However when installing the package as is done in the AppVeyor script:

`install.packages('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/tlf-library/artifacts/tlf.zip', repos = NULL, type = 'binary')`

the installed version is 1.5.125:

image

HOW??? What is this 1.5.125 build?? With this version, plots look different and this is why this PR fails.

PavelBal (approved):

Yuri05:
we have one open PR on the tlf and it seems that the AppVeyor permalink takes the artifact from the PR.
This should not be the case!
After checking the [AppVeyor docu](https://www.appveyor.com/docs/packaging-artifacts/#permalink-to-the-last-successful-build-artifact) on this topic: it is possible to exclude the PRs by adding `?pr=false` to the permalink.

grafik

will create issues for this in all R repos

grafik

Issue #1212, opened 2023-03-01
Us `.logSafe()` from utils


PavelBal:
Once https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/issues/126 is implemented.

PR #1218, opened 2023-03-07
- Use logSafe from utils package

### LINKED ISSUE #1091: Residual calculation
In PK-Sim/MoBi, log residuals are calculated via the defined there `LogSafe `function:
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fcd42c611764e4af44f266285175ef66f706fdde/src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculator.cs#L118-L124

with
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fcd42c611764e4af44f266285175ef66f706fdde/src/OSPSuite.Core/Domain/Constants.cs#L92

The log residual itself is then calculated as `LogSafe(simulatedValue) - LogSafe(observedValue)`:
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fcd42c611764e4af44f266285175ef66f706fdde/src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculatorForOnlyObservedData.cs#L12-L20

This solves different problems, e.g. 
* if simulated values are slightly negative (numerical zero) residual would return NaN otherwise
* if simulated values are positive but extremely small (numerical zero as well), residual might become extremely low (negative) value, causing algorithmic problems etc.

So I would suggest implementing residual calculation in R in the same way?

_Originally posted by @Yuri05 in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1090#discussion_r947213016_

PavelBal:
- Calculate residuals using ln instead of log10
Fixes #1091, #1087, #1212

### DIFF
--- a/tests/testthat/test-utilities-data-combined.R
+++ b/tests/testthat/test-utilities-data-combined.R
@@ -83,9 +83,16 @@ test_that(
 )
 
 test_that(
-  "calculateResiduals returns a correct vector of log residuals on example data",
+  "calculateResiduals returns a correct vector of log residuals on example data",{
+  pairedData <- calculateResiduals(myDC, scaling = "log")
+  expectedResiduals <- sapply(seq_along(pairedData$yValuesObserved), function(idx){
+    ospsuite.utils::logSafe(pairedData$yValuesSimulated[[idx]]) -
+    ospsuite.utils::logSafe(pairedData$yValuesObserved[[idx]])
+    })
+
   expect_equal(calculateResiduals(myDC, scaling = "log")$residualValues,
-    c(0, 0.022276400, -0.007178578, 0.00000000, -0.010723855, -20),
+               expectedResiduals,
     tolerance = 1e-5
   )
+  }
 )
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,13 @@
 
 ## Major Changes
 
+* Function `calculateResiduals()` uses natural logarithm (`log()`) when calculating 
+residuals with `scaling = "log"` as opposed to `log10()` as it was done before. This also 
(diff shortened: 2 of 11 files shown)


PavelBal:
Natural logarithm is what expected by most people when calculating the residuals, so I guess it is safe to make this switch here.

tests/testthat/test-utilities-data-combined.R
```diff
@@ -83,9 +83,16 @@ test_that(
...
+  "calculateResiduals returns a correct vector of log residuals on example data",{
+  pairedData <- calculateResiduals(myDC, scaling = "log")
+  expectedResiduals <- sapply(seq_along(pairedData$yValuesObserved), function(idx){
+    ospsuite.utils::logSafe(pairedData$yValuesSimulated[[idx]]) -
+    ospsuite.utils::logSafe(pairedData$yValuesObserved[[idx]])
+    })
+
   expect_equal(calculateResiduals(myDC, scaling = "log")$residualValues,
-    c(0, 0.022276400, -0.007178578, 0.00000000, -0.010723855, -20),
+               expectedResiduals,
```
svavil, line 94: Do we need to define `expectedResiduals` before using? Might have been left undefined.

PavelBal: Line 88 is defining the expected residuals, or what do you mean?

Felixmil (approved):

PavelBal:
@Yuri05 Would like to have your "go" on this before merging.

Yuri05 (approved):

PR #1225, opened 2023-03-27
Update snapshots

PavelBal:


Yuri05:
Build still failing...

Yuri05:
@PavelBal Seems that the submodule update is wrong (both submodules)

PR #1226, opened 2023-03-29
Snapshot update

PavelBal:
Add installation of package `showtext`


Felixmil:
🥳

DESCRIPTION
```diff
@@ -35,7 +35,9 @@ Suggests:
     rmarkdown,
     testthat (>= 3.0.3),
     vdiffr (>= 1.0.0),
-    withr
+    withr,
+    showtext,
```
Felixmil, line 39: Suggest are not supposed to propagate to depending package but well... I guess this is a special case.

Felixmil (approved):

DESCRIPTION
PavelBal: > Suggest are not supposed to propagate to depending package but well... I guess this is a special case.

I was desperate... but it did actually help. Could you please add it to suggest of esqlabsR?

PavelBal:
@Felixmil could you approve once more, please? I removed the explicit installation of the aforementioned packages from the AppVeyor script (works just fine)

Felixmil (approved):

PR #1234, opened 2023-04-06
Fixes #1233 Adjust install_dependencies.R

### LINKED ISSUE #1233: Adjust install_dependencies.R
Remove downloaded osp packages after installation (otherwise they need to be added to .Rbuildignore)
Add patchwork (missing according to the last nightly build)

Yuri05:


install_dependencies.R
```diff
@@ -1,4 +1,8 @@
-install.packages(c('testthat', 'vdiffr', 'spelling','covr'), repos = 'http://cran.us.r-project.org')
+install.packages(c('testthat', 'vdiffr', 'spelling','covr','patchwork'), repos = 'http://cran.us.r-project.org')
```
Felixmil, line 1: Why install patchwork while it is nor an Import or Suggest package ?

Yuri05: Because: https://ci.appveyor.com/project/open-systems-pharmacology-ci/OSPSuite-R/branch/develop
grafik

Felixmil: > Maybe installing the osp packages **with dependencies** 

Yes I think so.

Yuri05: will try devtools::install_local() instead

Felixmil: I stumbled on the same issue. `install_local()` should do it indeed even though I I have an error from RTools. But I think this is an issue with my configuration.

install_dependencies.R
```diff
@@ -1,4 +1,8 @@
-install.packages(c('testthat', 'vdiffr', 'spelling','covr'), repos = 'http://cran.us.r-project.org')
+install.packages(c('testthat', 'vdiffr', 'spelling','covr','remotes'), repos = 'http://cran.us.r-project.org')
 install.packages('https://github.com/Open-Systems-Pharmacology/rClr/releases/download/v0.9.1/rClr_0.9.1.zip', repos = NULL, type = 'binary')
-download.file('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/ospsuite-rutils/artifacts/ospsuite.utils.zip?pr=false', destfile = 'ospsuite.utils.zip', mode='wb'); install.packages('ospsuite.utils.zip', repos = NULL, type = 'binary')
-download.file('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/tlf-library/artifacts/tlf.zip?pr=false', destfile = 'tlf.zip', mode='wb');  install.packages('tlf.zip', repos = NULL, type = 'binary')
+download.file('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/ospsuite-rutils/artifacts/ospsuite.utils.zip?pr=false', destfile = 'ospsuite.utils.zip', mode='wb')
+install.packages('ospsuite.utils.zip', repos = NULL, type = 'binary')
```
Yuri05, line 4: There is currently no source package artifact without version, so I cannot use the same method as for the tlf.
Think it's ok - dependencies of ospsuite.utils would change very rarely

Felixmil: Then which part of the code installs ospsuite.utils dependencies ? (purrr and R6)

https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/blob/88152af44e32e714a1d5c80237c239105ed85120/DESCRIPTION#L25-L27

Yuri05: >Then which part of the code installs ospsuite.utils dependencies ? (purrr and R6)

SkyNet 😄

Yuri05: grafik
which is triggered by 
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/a8d5220a1636357f962df9e6be8ae8372aeae0e6/appveyor.yml#L34

installs the dependencies of ospsuite-R
Purrr is one of the deps, R6 is probably an indirekt dependency of some packages required by the ospsuite-R

Yuri05: yep, R6 is a dependency of dplyr
https://github.com/tidyverse/dplyr/blob/29307bf661a8c52b892c28afa531b38461aababb/DESCRIPTION#L32

Felixmil: Ok. Even if the packages are already installed, maybe we should add their installation explicitly in this script ? 

```
# install ospsuite.utils dependencies
install.packages(c('r6', 'purrr'), repos = 'http://cran.us.r-project.org')
```

install_dependencies.R
Felixmil: I think we could come to the same result using `remotes::install_github("Open-Systems-Pharmacology/TLF-Library")` which installs the dependency and also targets the latest `develop` commit.

install_dependencies.R
Felixmil: I think we could come to the same result for TLF using `remotes::install_github("Open-Systems-Pharmacology/TLF-Library")` which installs the dependencies and also targets the latest `develop` commit.

So maybe something like this could be used for `ospsuite-rutils` ?

PavelBal (approved):
Build fails but due to changed snapshots. I am merging.

PR #1237, opened 2023-04-07
Cach .Net Tasks for better performances

Felixmil:
## Summary
While profiling the code for performance improvement, I noticed that it was actually the `.getNetTask` function that was called thousand times by the `toUnit` function, even for "simple" plots.
What  I propose here is a way to cache .Net Task in the package environment the first time it is created.

## Symptom 🩺
Plots can take very long time to generate. 

## Problem 🪲

This is what I saw during profiling of a project's plot:

Pasted image 20230407115430

Here I focus on the **`toUnit()` function that took 6.6 seconds** for a plot.

When tested on its own, this function is actually quick, even with large numeric vectors. So I counted the number of time this function was called to generate **1 plotGrid**... **8934** times !

Let's try to call the function 8934 times for converting 1 vallue from L to mL:
```r
> simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
> sim <- loadSimulation(simPath)
> par <- getParameter("Organism|Liver|Volume", sim)
> bench::mark(toUnit(par, 1, "ml"), iterations = 8934)
```
```
# A tibble: 1 × 13
  expression                min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    memory              time               gc                  
  <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    <list>              <list>             <list>              
1 toUnit(par, 1, "ml")   1.13ms   1.17ms      731.    47.3KB     2.38  8905    29      12.2s <dbl [1]> <Rprofmem [47 × 3]> <bench_tm [8,934]> <tibble [8,934 × 3]>
```

**12 seconds**

But the culprit is actually `.getNetTask("DimensionTask")` which is called by `toUnit()`:
```
> bench::mark(.getNetTask(.getNetTask("DimensionTask")), iterations = 8934)
# A tibble: 1 × 13
  expression                        min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    memory             time               gc                  
  <bch:expr>                   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    <list>             <list>             <list>              
1 .getNetTask("DimensionTask")    827µs    847µs     1023.      328B    0.803  8927     7      8.72s <cobjRef> <Rprofmem [1 × 3]> <bench_tm [8,934]> <tibble [8,934 × 3]>
```
**8.72 seconds !**

So, **creating new "DimensionTask" repeatedly is what takes time** in this part of the code.

## Proposition 💡

Create .Net Tasks just once: when we need them for the first time. Then, they are stored in the `ospsuiteEnv` and reused. This is not true for Simulation tasks, a new one is created for each simulation.

## Results 🚀

➡️  **82 % less time for the toUnit function**

```
bench::mark(toUnit(par, 1, "ml"), toUnit_old(par, 1, "ml"), iterations = 8900)
# A tibble: 2 × 13
  expression                    min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    memory               time               gc                  
  <bch:expr>               <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    <list>               <list>             <list>              
1 toUnit(par, 1, "ml")        208µs  226.7µs     3458.     254KB    10.9   8872    28      2.56s <dbl [1]> <Rprofmem [522 × 3]> <bench_tm [8,900]> <tibble [8,900 × 3]>
2 toUnit_old(par, 1, "ml")   1.14ms    1.2ms      741.     205KB     3.51  8858    42     11.95s <dbl [1]> <Rprofmem [474 × 3]> <bench_tm [8,900]> <tibble [8,900 × 3]>
```

image

The plot I used to debug this still takes time: **16s but previously it was 24s**! The biggest part is still to improve: `private$wrapExtensionMethod` from `data-column.R`  takes a total of 8 seconds and I'm sure we can do something about it.

## Quality ✅

```r
══ Results ═══════════════════════════════════════════════════════════════════════
Duration: 191.4 s

[ FAIL 0 | WARN 10 | SKIP 0 | PASS 1232 ]
```

## Potential side effects ❓
I am not familiar with the Task principle on the `.Net` side. Maybe @georgeDaskalakis  & @rwmcintosh have any comment on this ?


PavelBal:
This is great and I think it should be cafe to cache net tasks. I would like @msevestre to review to be sure.

msevestre (changes requested):
A few things to note
1- in general it should be ok to cache the tasks because calling a service should be idem potent. However, I need to verify that this is indeed the case to be sure
2. With this implementation, we should not cache the containerTask separately. It was done for the same reason, but getContainerTask should return the getNetTask("ContainerTask") and we were done with it
3. Why rename from get to create? This is semantically wrong as I am expecting a new task to be created when I call a create method. In this case, we are actually not doing this at all and always caching => rename back please
4. Looking at the code. One comment. The else ought to be removed because the function is not consistent. Once we return implicitly, once explicitly. If you simply say if("""), add to cache and in any case return the value from the cache, the code is explicit all the time
So all in all, this should be good. We can merge when the proposed changes are looked at. But I need to see if we need to exclude explicitly one task from caching.

Felixmil:
Thanks for the extended review.

1. It does not seem to be the case for simulation tasks (see comment on 3 below)
2. Fixed in last commit
3. > 3. Why rename from get to create? This is semantically wrong as I am expecting a new task to be created when I call a create method. In this case, we are actually not doing this at all and always caching => rename back please

    Originaly, I just changed `.getNetTask()` to add the implemented logic BUT it led to errors for simulation tasks:
    ```
     ── Error ('test-utilities-simulation.R:290:3'): It can run a simulation batch by varying some parameters and molecules ──
    Error in `rClr::clrCall(simulationRunner, "RunConcurrently")`: Type:    OSPSuite.Utility.Exceptions.OSPSuiteException
    Message: You already have Simulation and SimulationBatch objects and should not mix, please invoke Clear to start adding 
    objects from a fresh start
    ```
    So I needed to start simulation objects "from fresh start"  by running the original `.getNetTask`  each time it would be needed     (with `.createNetTask`) just like it was before. Maybe `getNewNetTask` would be better naming ?
4. Fixed in last commit

R/get-net-task.R
```diff
@@ -15,16 +15,27 @@
...
 
-  ospsuiteEnv$containerTask
+#' @title .getNetTask
+#'
+#' @param taskName
+#'
+#' @keywords internal
+.getNetTask <- function(taskName) {
+  if (is.null(ospsuiteEnv[[taskName]])) {
+    ospsuiteEnv[[taskName]] <- .getNewNetTask(taskName)
```
PavelBal, line 38: Let's define a separate cache for tasks. Otherwise, the user could have access to them via `getOSPSuiteSetting()` which is not optimal.

msevestre:
I think we need to make sure we return multiple instances and not only one. Creating two instances in R or .net is probably not what we need but let me check it out and I'll let you know

R/get-net-task.R
Felixmil: I created a `tasksEnv` in latest commits.

msevestre:
@Felixmil is looking at the dimensiontask that can be cached. I am talking about the ConcurrentSimulationRunner. Creating time is going to be << running time so in this case, this caching is not required. I am just wondering why this crashes in fact and I think that those tasks are not stateless in 
.NET. I had no time to look at this but I will to understand at least
As long as the PR doesn't cache them all , we should be good. 
In fact, global caching is the source of many evil. I'll suggest to only cache when we have too as opposed to caching all the time . DimensionTask, ContainerTask are good candidates.

PavelBal:
We are creating the following tasks:

- `DataRepositoryTask` (propose to cache)
- `DataImporterTask` (propose to cache)
- `OutputIntervalFactory`
- `PKAnalysisTask`
- `PKParameterTask`
- `PopulationTask`
- `FullPathDisplayResolver`
- `SensitivityAnalysisRunner`
- `SimulationResultsTask`
- `SimulationPersister`
- `SimulationRunner`
- `ConcurrentSimulationRunner`
- *`DimensionTask` (already cached but not used everywhere!)*
- *`ContainerTask` (already cached)*

The problem of the `toUnit()` function was that it was creating new dimension tasks instead of loading it from cache!


So I would propose the following:

- Function `.getNetTask()` remains as is and returns a new tasks (for compatibility reasons)
- Function `.getNetTaskFromChache()` returns a cached version. Cached versions should be stored in the `tasks-env` environment
- Remove functions `.getContainerTask()` and `.getDimensionTask()`. Instead, everywhere where they were used use `.getNetTaskFromChache()`. This way we will not have to create new functions every time we decide to cache a task.
- Make sure that the cached tasks are NEVER created with `.getNetTask()` but always with `.getNetTaskFromChache()`.
- I propose to also cache the tasks `DataRepositoryTask` and `DataImporterTask`.  For other tasks, its up to @msevestre to decide.

Felixmil:
All @PavelBal suggestions were implemented in last commits.  This is ready for review @msevestre.

PR #1257, opened 2023-06-21
Datacombined  optimizations

Felixmil:
Remove duplicated steps and optimized functions.

### DIFF
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -596,7 +596,7 @@ DataCombined <- R6::R6Class(
         missingNames <- specifiedNames[!specifiedNames %in% currentNames]
 
         message(messages$printMultipleEntries(
-          header = messages$datasetsToGroupNotFound(),
+          header  = messages$datasetsToGroupNotFound(),
           entries = missingNames
         ))
       }
@@ -772,9 +772,9 @@ DataCombined <- R6::R6Class(
 
     # private fields ----------------------------------------
 
-    .dataCombined = NULL,
-    .groupMap = NULL,
-    .names = NULL,
+    .dataCombined        = NULL,
+    .groupMap            = NULL,
+    .names               = NULL,
     .dataTransformations = NULL
   ),
 


R/data-combined.R
```diff
@@ -653,42 +650,30 @@ DataCombined <- R6::R6Class(
...
+        xScaleFactors = tidyr::replace_na(xScaleFactors, 1),
+        yScaleFactors = tidyr::replace_na(yScaleFactors, 1)
       )
 
-      # Update data frame using given transformation parameters
-      private$.dataTransformations <- private$.updateDataFrame(private$.dataTransformations, dataArg)
+      # Update private$.dataTransformations with new transformation settings
+      private$.dataTransformations <-
+        private$.updateDataFrame(private$.dataTransformations,
+                                 dataArg) %>%
```
msevestre, line 664: I am unsure about this formatting. Why is data Arg on the second line suddenly?

R/data-combined.R
```diff
@@ -653,42 +650,30 @@ DataCombined <- R6::R6Class(
...
-          .cols = matches("scalefactors$"), # relevant only for columns matching this pattern
-          .fns = function(x) tidyr::replace_na(x, 1)
-        )
-      )
+      # they should be removed and set with the new/default values.
+      data <-
+        data %>%
+        dplyr::select(-tidyr::any_of(colnames(private$.dataTransformations)[-1])) %>%
+        dplyr::left_join(private$.dataTransformations,
+                         by = "name")
```
msevestre, line 676: same with this. line 705, the arrange method is called and it is organized differently. I think we need to write the code one way or the other. In the same PR, we HAVE to be consistent . Also specifying named parameter onkly for one parameter..is this usual in R? Why not for the FIRST PARAMETER?

msevestre (changes requested):
@Felixmil I don;t understand enough to say if this makes sense or not
I see two issues however
1-cohesion of style within the same PR. For me, it looks like mixed potatoes here. Either we name parameter or we don't. Either we align them or we don't. Between files, I understand that this is hard to enforce, in the same fie, it should be enforce but this is hard. BUT IN ONE CODE CHANGE....I don't see why this is happening
2-This is the bigger problem. There is a code change that DOES NOT reference an issue. This is for me a big no no unless code changes are trivial. Here, the code is not trivial, I do not know what triggered a change. Performance ? bug? New feature? 

/cc @PavelBal

R/data-combined.R
Felixmil: I used {styler} that autoformatted this. I will make it consistend in next commits.

And yes, usually, first and sometimes second arguments are not named.
> In general, use positional matching only for the first one or two arguments; they will be the most commonly used, and most readers will know what they are. Avoid using positional matching for less commonly used arguments/
> source: https://adv-r.hadley.nz/functions.html?q=functions#prefix-form

Felixmil:
> @Felixmil I don;t understand enough to say if this makes sense or not I see two issues however 1-cohesion of style within the same PR. For me, it looks like mixed potatoes here. Either we name parameter or we don't. Either we align them or we don't. Between files, I understand that this is hard to enforce, in the same fie, it should be enforce but this is hard. BUT IN ONE CODE CHANGE....I don't see why this is happening 2-This is the bigger problem. There is a code change that DOES NOT reference an issue. This is for me a big no no unless code changes are trivial. Here, the code is not trivial, I do not know what triggered a change. Performance ? bug? New feature?
> 
> /cc @PavelBal

1. Should look a bit less like mixed potatoes by now.
2. To provide some context, we have been actively seeking performance improvements, and @PavelBal has expressed interest in addressing the dataCombined aspect for some time now. I'm confident in stating that these changes not only enhance the performance of dataCombined, but also improve the overall maintainability and robustness of the codebase. With good test coverage and successful passing of all tests, we can be assured of the reliability and stability of the changes made.

If you have any other specific concerns or suggestions for improvement, I would be more than happy to address them promptly.

msevestre:
Nope this looks great.

Just to give some context to my remark on #2

Let's make sure to create an issue before fixing anything in the code base unless it's trivial (and that's not even something we should do too often. We only do it when we update dependencies in PKSim and MoBi or perform typo corrections. R meets the exact same requirements). 
The issue should outline what is to be changed, and why. (<=this is something we don't always do that well but we try)

In the PR, for review, a little bit of context would help inexperienced reviewers also. In this specific PR, there are chunks of code removed. Explain why. That way, I can follow the thought process. I am not looking at the specific syntax. I know that I am underqualified: ). But the overall "algorithm" change may be explained in one sentence., Maybe it's there and I missed it too. :)

PR #1261, opened 2023-06-26
Fixes #1260 Update core

### LINKED ISSUE #1260: Update core

rwmcintosh:
Update core #1260


msevestre:
The build does not work
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r-r99ji/builds/47399091#L1849

There are still references to BuildConfiguration

```
i Testing ospsuite
v | F W S  OK | Context
v |         4 | AgingData                                                                      
x | 1       0 | Application [1.2s]                                                             
-----------------------------------------------------------------------------------------------
Error (test-application.R:4:1): (code run outside of `test_that()`)
Error in `clrCallStatic(clrFacadeTypeName, "GetFieldOrProperty", objOrType, 
    name)`: Type:    System.ArgumentException
Message: Field or property name 'BuildConfiguration' not found on object of type 'Simulation', for binding flags 'Instance, Public'
Method:  System.Object internalGetFieldOrProperty(System.Type, System.String, System.Reflection.BindingFlags, System.Object)
Stack trace:
   at Rclr.ClrFacade.internalGetFieldOrProperty(Type t, String name, BindingFlags b, Object obj_or_null)


Backtrace:
 1. ospsuite:::loadTestSimulation("S1")
      at test-application.R:4:0
 2. ospsuite::loadSimulation(...)
      at C:\dev\OSPSuite-R\tests\testthat\helper-for-tests.R:15:2
 3. Simulation$new(netSim, filePath)
      at OSPSuite-R/R/utilities-simulation.R:60:2
 4. ospsuite (local) initialize(...)
 5. rClr::clrGet(self$ref, "BuildConfiguration")
      at OSPSuite-R/R/simulation.R:54:6
 6. rClr::clrCallStatic(...)
 7. ```

msevestre:
which comes from here
    private$.buildConfiguration <- rClr::clrGet(self$ref, "BuildConfiguration")

this does not exist anymore

Instead we have the BuildConfiguration and then the "builder" that is created and allocated once for cachign purposes

image

So somehow this needs to be instantiated and cahced IN R instead of the cached configuration. The actual methods are then the same (AllPresentEndogenousStationaryMoleculeNames, AllPresentXenobioticFloatingMoleculeNames etc..)

msevestre (changes requested):
The code needs to be changed as the API surface of ospsuite.core, also used by ospsuite-R has changed

PavelBal:
@msevestre Can you take care of that? Without knowing the changes in Core, it would be hard for me to do.

rwmcintosh:
> @msevestre Can you take care of that? Without knowing the changes in Core, it would be hard for me to do.

I should be able to have a go.

msevestre:
@rwmcintosh Yes, this should be failry straithforward to get it to go. HOWEVER, we never thought about R during this change so maybe, just maybe, some functions we are exporting do not make sense anymore. I can't tell

PR #1263, opened 2023-06-27
plotpopulationtimeprofile supports other aggregations

Felixmil:


pchelle (changes requested):
Nice !
Could you create an enum that provides the list of aggregations for users ?

Felixmil:
> Nice ! Could you create an enum that provides the list of aggregations for users ?

Thanks.

Here is the documentation listing the available aggregations:
image

They are also introduced in the examples.

Is this list supposed to be used elsewhere ? isn't it enough to list them in doc for now ?

NEWS.md
```diff
@@ -12,6 +12,8 @@ residuals with `scaling = "log"` as opposed to `log10()` as it was done before.
 affects figures created by `plotObservedVsSimulated()`, `plotResidualsVsSimulated()`, and 
 `plotResidualsVsTime()`. See discusssions https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1091
 and https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1087.
+- `plotPopulationTimeProfile()` has a new `aggregation` argument that allow the user to choose between quantiles, 
+  arithmetic and geometric aggragations of y.
```
Yuri05, line 16: typo: `aggragations `

Felixmil, line 16:
```suggestion
  arithmetic and geometric aggregations of y.
```

Yuri05:
I agree with @pchelle - we should introduce an enum.

>Is this list supposed to be used elsewhere ? isn't it enough to list them in doc for now ?

The function will be called by the end users, who need to know all possible values for the aggregation argument.
Of course, the user could look up in the docu, but having the enum is more convenient and less error-prone.
Also, the function should have a check, if the aggregation value is one of the enum values

R/plot-individual-time-profile.R
```diff
@@ -58,7 +58,8 @@ plotIndividualTimeProfile <- function(dataCombined,
 #' @noRd
 .plotTimeProfile <- function(dataCombined,
                              defaultPlotConfiguration = NULL,
-                             quantiles = NULL) {
+                             aggregation = NULL,
+                             ...) {
   # validation -----------------------------
```
Yuri05, line 63: Here we should check that `aggregation `is one of the allowed enum values

Felixmil: also answering for: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1263#issuecomment-1611096224

This is done at a lower level, directly in the .extractAggregatedSimulatedData() function here: 
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/68b73fb6d03a7a7c34a7a9f56054d2d532a7ddf4/R/utilities-plotting.R#L149

Here's what it looks like:

```r
plotPopulationTimeProfile(myDataComb,
                            aggregation = "anotherAggregation")
# Error in `.extractAggregatedSimulatedData()` at OSPSuite-R/R/plot-individual-time-profile.R:111:4:
#   ! `aggregation` must be one of "quantiles", "arithmetic", or "geometric", not "anotherAggregation".
# Run `rlang::last_trace()` to see where the error occurred.
```

Yuri05: I see. Still, the enum is preferable from my point of view.

R/plot-population-time-profile.R
```diff
@@ -23,21 +21,27 @@
...
+#' plotPopulationTimeProfile(myDataComb,
+#'                           aggregation = "arithmetic")
+#
+#' # plot with arithmetic mean and 3sd above and below average line
+#' plotPopulationTimeProfile(myDataComb,
+#'                           aggregation = "arithmetic",
+#'                           n=3)
 #'
 #' @export
 plotPopulationTimeProfile <- function(dataCombined,
```
PavelBal, line 41: I really do hate the `...` in public functions and would like to avoid it as much as possible. In this case, let's get the `quantiles` argument back. Otherwise, this would be a breaking change which is not necessary here.

Also, the `n` argument is a bit confusing, am I not aware of use cases where someone would want to plot n-times sd instead of `sd` only.

PavelBal (changes requested):
See comment about the `...` argument.

Felixmil:
> See comment about the `...` argument.

I don't really like `...` either but **when it is not well documented**. While it's understandable to have personal preferences, it's crucial to recognize the significance of the '...' argument in R functions.  It makes the code flexible for the user, compatible with future changes and also cleaner. 
In this case, with the documentation, it is clear what the user can achieve (i.e customize the quantile, mean or sd executions). Also the functions calls chain here is rather simple and should be quite easy to understand.

**New proposition: I added a piece of code that ensure the previous `quantiles` argument still works (while showing a deprecate warning) so it is not a breaking change.**

What do you think @PavelBal  ? And also @Yuri05, @pchelle  ?

Is it not useful be able to choose to display either 68%, 95%, 99.7% of data as range ?

R/plot-individual-time-profile.R
Felixmil: done in [be9d183](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1263/commits/be9d1836253d77ba30804e17a77f813be895ebaa). 

I created a `SimulatedDataAggregationMethods` `ospsuite.utils::enum` and used `ospsuite.utils::validateEnumValue` for validation

R/utilities-plotting.R
```diff
@@ -683,9 +707,74 @@
...
+#' Quantile Range
+#'
+#' @param x numeric vector to compute quantile range from
+#' @param probs numeric vector of length 3 containing probabilities with values in [0,1].
+#' @param na.rm logical; if true, any NA and NaN's are removed from x before the quantiles are computed.
+#' @param ... further arguments passed to stats::quantile.
+#'
+#' @return numeric vector of length 3 representing the computed quantiles of x.
+#' @noRd
+.quantRange <- function(x, probs = c(0.05, 0.5, 0.95), na.rm = FALSE, ...){
```
PavelBal, line 736: Just for my understanding... There is no `@keyword internal` in the roxygen code here, but the function is still not exported (which is good). How does this work?

R/utilities-plotting.R
```diff
@@ -683,9 +707,74 @@
...
+      !is.null(generalPlotConfiguration[[specificSetting]])) {
     if (generalPlotConfiguration[[specificSetting]] != specificPlotConfiguration[[specificSetting]]) {
       specificPlotConfiguration[[specificSetting]] <- generalPlotConfiguration[[specificSetting]]
     }
   }
 }
+
+#' Names of aggregation available for .extractAggregatedSimulatedData
+#'
+#' @keywords internal
```
PavelBal, line 719: Let's call it `DataAggregationMethods`, as the aggregation can also be applied to the observed data.

PavelBal (dismissed):

R/utilities-plotting.R
Felixmil: This is because of `@noRd`

`Keyword internal` means there will be a manual page for the function but it won't be in the documentation index. [https://roxygen2.r-lib.org/articles/rd.html#indexing ](https://roxygen2.r-lib.org/articles/rd.html#indexing)" `@keywords internal` is useful because it removes the function from the documentation index; it’s useful for functions aimed primarily at other developers, not typical users of the package."

`@NoRd` means there won't even be a manual page so the `roxygen2` doc of that function only lives in the R script.

PavelBal: I mean that it also cannot be found with `ospsuite::.quantRange`

Felixmil: There is no `@export` tag either ;)

PavelBal: True :D

R/utilities-plotting.R
PavelBal: After this it can be merged

PavelBal:
@pchelle Could you re-review?

PavelBal (dismissed):

tests/testthat/_snaps/plot-population-time-profile/arithmetic-mean-with-3-sd.svg
Yuri05, line 1: Looking at this and other snapshot plots: plot legend is unclear. Is it just because the legend was not properly set up **in the tests** or is it a problem with `PlotPopulationTimeProfile`?
E.g. here:
* It is not clear what the **lines** refer to (mean, median,...)
* It is not clear, what the **range** refers to (mean+-SD, x-y-perecentile range, ...)
* Dashed line displayed as solid in the legend

PavelBal: General problem, created a new issue.

PavelBal:
@pchelle @msevestre can we have this merged soon?

pchelle:
> @pchelle @msevestre can we have this merged soon?

No, the enum is not exported, so the user cannot easily access it if needed.
`@keywords internal` should be replaced by `@export`

PavelBal:
Fixed.

pchelle (approved):

R/plot-population-time-profile.R
```diff
@@ -23,21 +21,51 @@
...
+                            average for `arithmetic` or `geometric`."
+    )
+    names(dots)[names(dots) == "quantiles"] <- "probs"
+  }
 
-  .plotTimeProfile(dataCombined, defaultPlotConfiguration, quantiles)
+  rlang::inject(.plotTimeProfile(dataCombined,
+                                 defaultPlotConfiguration,
+                                 aggregation,
+                                 !!!dots))
```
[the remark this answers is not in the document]
PavelBal: I read the documentation on this but did not really understand :D
https://rlang.r-lib.org/reference/splice-operator.html

R/utilities-plotting.R
```diff
@@ -685,9 +708,74 @@
...
+#' Quantile Range
+#'
+#' @param x numeric vector to compute quantile range from
+#' @param probs numeric vector of length 3 containing probabilities with values in [0,1].
+#' @param na.rm logical; if true, any NA and NaN's are removed from x before the quantiles are computed.
+#' @param ... further arguments passed to stats::quantile.
+#'
+#' @return numeric vector of length 3 representing the computed quantiles of x.
+#' @noRd
+.quantRange <- function(x, probs = c(0.05, 0.5, 0.95), na.rm = FALSE, ...){
```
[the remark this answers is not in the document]
PavelBal: @pchelle Honestly I have no idea, what would be the use case where the data would contain NAs?

At any case, looking at the code I believe this was the default behavior before, so nothing changed here.

R/plot-population-time-profile.R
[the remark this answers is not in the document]
Yuri05: I had no idea about this `!!!` operator but find it pretty cool.
While I didn't understand all the use cases in the docu as well, here it is used to split the list into single arguments.
So the original `...` arguments are combined to the list, the argument "quantiles" (if found) is renamed to "probs" and the new argument list is then splitted into single arguments back again by `!!!`. Funky :)

PavelBal:
image

@pchelle In this case I belive that less is more, and we should only expose and document the arguments that are intuitive and commonly used. So I removed the `...` and added `quantiles` arguments back. It is great that the underlying implementation allows to easily extend the arguments list, but e.g. in case of `CI` I would wait until we have a use case for that to better understand how the documentation and signatures should look like. What do you think?

PavelBal (approved):

PR #1278, opened 2023-08-17
Fix Aggregation computations

Felixmil:
Previously, the default quantiles argument was overwriting .quantRange and .geoRange arguments and would produce wrong plots or errors


tests/testthat/_snaps/plot-population-time-profile/geometric-mean.svg
PavelBal, line 1: This must be wrong, geometric mean should not be that far away from arithmetic mean..

This is how geometric mean for these results is drawn by PK-Sim:

image

PavelBal (changes requested):

R/plot-population-time-profile.R
```diff
@@ -48,12 +48,13 @@
...
-  probs <- quantiles
+                                      quantiles = c(0.05, 0.5, 0.95),
+                                      ...) {
   .plotTimeProfile(
     dataCombined,
     defaultPlotConfiguration,
     aggregation,
-    probs
+    probs = quantiles,
+    ...
```
PavelBal, line 58: I would say the `...` is not necessary at all.

PavelBal:
Could you please add tests for normRange and geoRange? I mean, yes, the functions are simple, but just to be confident about what it does. Something like this should suffice:

```
meanVals <- c(mean(randu$x) - sd(randu$x), mean(randu$x), mean(randu$x) + sd(randu$x))
meanVals - .normRange(randu$x)


geomeanVals <- c(geomean(randu$x) - geosd(randu$x), geomean(randu$x), geomean(randu$x) + geosd(randu$x))
geomeanVals - .geoRange(randu$x)
```

PavelBal (approved):

PR #1281, opened 2023-08-30
Moved the dimension retrieval inside the getBaseUnit() function

svavil:
Currently, the `toBaseUnit()` function can take either a `Quantity` or a `Dimension` as an input, and `getBaseUnit()` function strictly requires a `Dimension` as an input. With the proposed change, `getBaseUnit()` behavior will be more consistent with the `toBaseUnit()`.

### DIFF
--- a/R/utilities-units.R
+++ b/R/utilities-units.R
@@ -65,11 +65,16 @@ validateUnit <- function(unit, dimension) {
 
 #' Get base unit of a dimension
 #'
-#' @param dimension Dimension (string name) for which the base unit is returned.
+#' @param quantityOrDimension Instance of a quantity from which the dimension will be retrieved or name of dimension
 #'
 #' @return String name of the base unit.
 #' @export
-getBaseUnit <- function(dimension) {
+getBaseUnit <- function(quantityOrDimension) {
+  if (isOfType(quantityOrDimension, "Quantity")) {
+    dimension <- quantityOrDimension$dimension
+  } else {
+    dimension <- quantityOrDimension
+  }
   validateDimension(dimension)
   dimensionTask <- .getNetTaskFromCache("DimensionTask")
   rClr::clrCall(dimensionTask, "BaseUnitFor", enc2utf8(dimension))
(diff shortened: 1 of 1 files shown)


PavelBal (changes requested):
- Test failing
- Re-build documentation
- Add this change to NEWS and mention that this is a breaking change (argument name  has changed!)

Yuri05:
> * and mention that this is a breaking change (argument name  has changed!)

why is "argument name  has changed" a breaking change?

R/utilities-units.R
```diff
@@ -65,11 +65,16 @@ validateUnit <- function(unit, dimension) {
...
-#' @param dimension Dimension (string name) for which the base unit is returned.
+#' @param quantityOrDimension Instance of a quantity from which the dimension will be retrieved or name of dimension
 #'
 #' @return String name of the base unit.
 #' @export
-getBaseUnit <- function(dimension) {
+getBaseUnit <- function(quantityOrDimension) {
+  if (isOfType(quantityOrDimension, "Quantity")) {
+    dimension <- quantityOrDimension$dimension
+  } else {
```
Yuri05, line 75: please add a new test for this new use case.

PavelBal:
> > * and mention that this is a breaking change (argument name  has changed!)
> 
> why is "argument name has changed" a breaking change?

Because `getUnit(dimension = "Amount")` will not work any more (this is why one of the tests is failing).

Yuri05:
> > why is "argument name has changed" a breaking change?
> 
> Because `getUnit(dimension = "Amount")` will not work any more (this is why one of the tests is failing).

ok

NEWS.md
```diff
@@ -4,6 +4,7 @@
 - Lower limit of quantification (LLOQ) is plotted as dotted lines for `plotIndividualTimeProfile()` and `plotPopulationTimeProfile()`.
 
 ## Breaking Changes
+- The single argument `getBaseUnit()` function is now named `quantityOrDimension` (was `dimension` previously). It can now use objects of class `Quantity` as inputs, to be consistent with the `toBaseUnit()` function.
```
PavelBal, line 7:
```suggestion
- The single argument of the `getBaseUnit()` function is now named `quantityOrDimension` (was `dimension` previously). It can now use objects of class `Quantity` as inputs, to be consistent with the `toBaseUnit()` function.
```

PavelBal (dismissed):

PavelBal (approved):

PR #1286, opened 2023-09-03
Fixes #1283 Speed up build time

### LINKED ISSUE #1283: Speed up build time
* extend `R_CHECK_ARGS` with `--no-build-vignettes` both in **appveyor.yml** and **appveyor-nigthly.yml** (compare RE: https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/c7642c66e0da8a45a75431623ad8280721c8b40b/appveyor.yml#L24)
* install CRAN package dependencies as **binary**

Yuri05:


Yuri05:
little bit better :)
grafik

appveyor-nightly.yml
```diff
@@ -27,7 +27,8 @@ environment:
   KEEP_VIGNETTES: true
   R_ARCH: x64
   R_VERSION: 4.1.0
-  R_CHECK_ARGS: --no-multiarch --no-manual --as-cran
+  R_CHECK_ARGS: --no-multiarch --no-manual --as-cran --no-build-vignettes
```
Yuri05, line 30: I don't see any gain in this step. Vignettes are just built and then are tested again in the installed package.

The only use case I see where this could fail: an example file, which is used in the vignettes, is explicitly excluded from the package via **.RBuildIgnore**. 

In all other cases: if vignette **creation** was successful - vignette **check** will be successful too.

appveyor-nightly.yml
```diff
@@ -27,7 +27,8 @@ environment:
...
+  R_PKGTYPE: binary
```
Yuri05, line 31: force **binary** package installation in `travis-tool.sh install_deps`

appveyor.yml
```diff
@@ -24,12 +24,13 @@ environment:
   NOT_CRAN: true
   KEEP_VIGNETTES: false
   R_BUILD_ARGS: --no-build-vignettes --no-manual
-  R_CHECK_ARGS: --no-multiarch --no-manual --as-cran
+  R_CHECK_ARGS: --no-multiarch --no-manual --as-cran --no-build-vignettes
```
Yuri05, line 27: Here vignettes check does not make any sense anyway, because vignettes are not even build.
It just produces false alarms in the build log.

appveyor.yml
```diff
@@ -24,12 +24,13 @@ environment:
...
   R_ARCH: x64
   R_VERSION: 4.1.0
   #We use this variable to skip some long lasting tests using "skip_on_ci"
   CI: true
-
+  R_PKGTYPE: binary
```
Yuri05, line 32: Same change as in the nightly build.

install_dependencies.R
```diff
@@ -1,4 +1,4 @@
-install.packages(c('testthat', 'vdiffr', 'spelling','covr','remotes'), repos = 'http://cran.us.r-project.org')
+install.packages(c('testthat', 'vdiffr', 'spelling','covr','remotes'), repos = 'http://cran.us.r-project.org', type='win.binary')
```
Yuri05, line 1: Force **binary** package installation for the manually installed packages as well

Felixmil (commented):
Nice improvements !

Issue #1287, opened 2023-09-04
`simulationResultsToDataFrame()` performance


PavelBal:
I have re-written `simulationResultsToDataFrame()` with the intention of improving the performance.

Old version:

```
sim <- loadTestSimulation("MinimalModel")
simResults <- importResultsFromCSV(
  simulation = sim,
  filePaths = getTestDataFilePath("Stevens_2012_placebo_indiv_results.csv")
)

system.time({
  for (i in 1:1000) {
    simulationResultsToDataFrame_old(simResults)
  }
})
```

Elapsed time:
```
   user  system elapsed 
  51.95    2.03   59.46 
```

New version:

```
gc()

system.time({
  for (i in 1:1000) {
    simulationResultsToDataFrame(simResults)
  }
})
```

Elapsed time:
```
user  system elapsed 
  17.17    0.29   19.08 
```

All tests are passing. PR is coming.

PR #1288, opened 2023-09-04
`simulationResultsToDataFrame()` speed up

PavelBal:


R/utilities-simulation-results.R
```diff
@@ -223,59 +223,30 @@ simulationResultsToDataFrame <- function(simulationResults,
...
-    dplyr::tibble(
-      "TimeUnit" = simList$metaData$Time$unit[[1]],
-      "TimeDimension" = simList$metaData$Time$dimension[[1]]
+  units <- lapply(simList$metaData, function(path) {
+    path$unit
+  })
+  dims <- lapply(simList$metaData, function(path) {
+    path$dimension
+  })
+  molWeights <- lapply(unique(simData$paths), function(path) {
```
PavelBal, line 232: @Felixmil any better way to do this?

R/utilities-simulation-results.R
```diff
@@ -223,59 +223,30 @@ simulationResultsToDataFrame <- function(simulationResults,
...
-  # with `nest()`, the computation only be carried for the same number of
-  # times as the number of `paths` present.
-  simData <- simData %>%
-    dplyr::group_by(paths) %>%
-    tidyr::nest() %>%
-    # Add a new column for molecular weight.
-    # When you call `molWeightFor()`, it returns the value in the base unit -
-    # which is `kg/µmol`. This is not the unit the user would expect, so we
-    # convert it first to the common unit `g/mol`.
+  simData <- dplyr::group_by(simData, paths) %>%
```
PavelBal, line 241: @Felixmil Also here - any speed up ideas?

tests/testthat/test-utilities-simulation-results.R
```diff
@@ -166,12 +166,12 @@ test_that("simulationResultsToDataFrame works as expected - minimal pkml", {
   expect_equal(unique(df2$paths), "Organism|A")
 
   # names
-  expect_equal(
-    names(df1),
-    c(
-      "paths", "IndividualId", "Time", "simulationValues", "unit",
-      "dimension", "TimeUnit", "TimeDimension", "molWeight"
-    )
+  expect_true(
```
PavelBal, line 169: Changed the test so it does not test for the exact order of columns. The order has changed to 

`"IndividualId"     "Time"             "paths"            "simulationValues" "TimeDimension"    "TimeUnit"         "dimension"        "unit"             "molWeight"  `

PavelBal:
Fixes #1287

R/utilities-simulation-results.R
Felixmil: This should be better performance and readability wise:

```r
units <- purrr::map(simList$metaData, "unit")

dims <- purrr::map(simList$metaData, "dimension")

molWeights <- unique(simData$paths) %>%
  purrr::set_names() %>%
  purrr::map(~ospsuite::toUnit(
    quantityOrDimension = ospDimensions$`Molecular weight`,
    values              = simulationResults$simulation$molWeightFor(.x),
    targetUnit          = ospUnits$`Molecular weight`$`g/mol`
  ))
```
\+ remove `names(molWeights) <- unique(simData$paths)`

R/utilities-simulation-results.R
Felixmil: this tibble operation should not be very long, however we can still improve it a bit for this kind of column mutation using `rowwise`:
```r
  simData <-
    simData %>%
    dplyr::rowwise() %>%
    dplyr::mutate(
      TimeDimension = dims$Time,
      TimeUnit = units$Time,
      dimension = dims[[paths]],
      unit = units[[paths]],
      molWeight = molWeights[[paths]]
    )
```

R/utilities-simulation-results.R
PavelBal: Yepp a bit better.

R/utilities-simulation-results.R
PavelBal: This actually DOUBLED the time needed :D

tests/testthat/test-utilities-simulation-results.R
svavil: This tests that there are no additional columns in `names(df1)`, but not that all columns are present. Maybe 
```
expect_true(all(names(df1) %in% c(...))
expect_true(all(c(...) %in% names(df1))
```
or
```
expect_equal(sort(names(df)), sort(c(...)))
```

R/utilities-simulation-results.R
Felixmil: Here is the solution using `{data.table}`:

```r
  simData <-  data.table::as.data.table(simData)
  simData <- simData[,`:=`(TimeDimension =  dims$Time,
                           TimeUnit =  units$Time,
                           dimension =  dims[[paths]],
                           unit =  units[[paths]],
                           molWeight = molWeights[[paths]]),
                     by = paths]
```
Result: ~2 times faster. This should be even faster with large datasets
```
Unit: milliseconds
                                     expr    min      lq     mean median      uq     max neval
     og(simData, dims, units, molWeights) 4.3466 4.89990 5.936743 5.0822 5.40190 22.0154   100
    new(simData, dims, units, molWeights) 3.9706 4.34070 5.003569 4.5498 4.91425 18.0941   100
 new_dt(simData, dims, units, molWeights) 1.4282 2.07735 3.029173 2.3516 2.65680 16.4108   100
```

Felixmil: Results with a simData with `2 946 000` rows:
```r
Unit: milliseconds
                                         expr      min       lq     mean   median       uq      max neval
     og(big_simData, dims, units, molWeights) 204.5205 219.9517 326.0187 322.4694 430.7714 449.7938    10
    new(big_simData, dims, units, molWeights) 157.8323 180.3917 232.1167 187.7086 320.0275 374.1119    10
 new_dt(big_simData, dims, units, molWeights) 101.9321 119.9672 180.7313 123.3568 303.0376 340.7760    10
```

Felixmil (commented):
implement either new dplyr or data.table solutions

PavelBal:
@Felixmil nice, feel free to push.

PavelBal:
Hmm but the reference performance will be different for each machine?

svavil:
> Hmm but the reference performance will be different for each machine?

@PavelBal that's why I'm setting a 2-fold margin (maybe set a 3-fold or something), but basically, I don't want the performance to suddenly become worse in a year because some changes were accumulating.

Felixmil (approved):

Yuri05:
> > Hmm but the reference performance will be different for each machine?
> 
> @PavelBal that's why I'm setting a 2-fold margin (maybe set a 3-fold or something), but basically, I don't want the performance to suddenly become worse in a year because some changes were accumulating.

 If we add benchmark tests, I suggest to define a precondition which is FALSE as per default.
E.g. "benchmark test starts only if environment variable OSPSUITE-BENCHMARK-TESTS is defined".

svavil:
@PavelBal we can make it more machine-agnostic, if you are fine with some arbitrary benchmarks, for example: 
```
sim <- loadTestSimulation("MinimalModel")
  simulation = sim,
  filePaths = getTestDataFilePath("Stevens_2012_placebo_indiv_results.csv")
)
# This is an example benchmark that takes around 4 seconds
benchmark_1e8 <- system.time({for (i in 1:1e8) {res <- 2**0.5}})
# In the current implementation, 100 runs of simulationResultsToDataFrame work
# approximately twice faster than the benchmark above
test_that("The simulationResultsToDataFrame() performance is not degraded, compared to a reference performance", {
  benchmark <- system.time({
    for (i in 1:100) {
      simulationResultsToDataFrame(simResults)
    }
  })
  expect_lt(benchmark[["elapsed"]], benchmark_1e8[["elapsed"]])
})
```

Of course, wrapped in a flag that Yuri proposed.

PavelBal:
@svavil Feel free to make a PR for benchmark tests. In the meantime, I am merging this.

PR #1290, opened 2023-09-06
Data combined refactor

PavelBal:


PavelBal:
Re-written `DataCombined` and tried to minimize data frame modifications. This is surely not the fastest possible solution as I am not that familiar with the whole `dplyr` syntax, but I did my best. Feel free to profile the branch and try to find any further optimization possibilities.

## Comparison of current `dev` and this PR:

```
sim <- loadSimulation("D:\\Work\\OSPS\\Repos\\OSPSuite-R\\tests\\data\\MinimalModel.pkml")
  simulation = sim,
  filePaths = "D:\\Work\\OSPS\\Repos\\OSPSuite-R\\tests\\data\\Stevens_2012_placebo_indiv_results.csv"
)

```
### 1. Simply convert simulation results to data frame 1000x

```
system.time({
  for (i in 1:1000) {
    simulationResultsToDataFrame(simResults)
  }
})
```
- **`dev`:** 15 sec
- **this PR:** 14 sec

As expected, the performance is equal, as this does not include `DataCombined` yet.

### 2. add simulation results 1000x 

```
gc()
dc <- DataCombined$new()
system.time({
  for (i in 1:1000) {
    dc$addSimulationResults(simResults, silent = TRUE)
  }
})
```

- **`dev`:** 32 sec
- **this PR:** 19 sec

This is basically just adding the results, so in best case the performance should be equal with use case 1 (converting results to df). For `dev` more than 2x time is required, current PR is only 5 seconds slower than just converting the results (however, I would prefer it to be almost as fast as converting the results... This could be optimized)

### 3. add simulation results 1000x and call `toDataFrame` at the end

```
gc()
dc <- DataCombined$new()
system.time({
  for (i in 1:1000) {
    dc$addSimulationResults(simResults, silent = TRUE)
  }
  df <- dc$toDataFrame()
})
```

- **`dev`:** 33 sec
- **this PR:** 18 sec

Almost identical with the previous call, as `toDataFrame()` is called only once and can be neglected

### 4. add simulation results and call `toDataFrame()` 1000x

```
gc()
dc <- DataCombined$new()
system.time(
  for (i in 1:1000) {
    dc$addSimulationResults(simResults, silent = TRUE)
    df <- dc$toDataFrame()
  }
)
```

- **`dev`:** 48 sec
- **this PR:** 32 sec

Obviously lots of things happening (and can be improved), but overall the performance increase is not bad.

NEWS.md
```diff
@@ -19,6 +19,13 @@ and https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1087.
 
 ## Minor Changes
 
+- The order of columns of the data frame returned by `DataCombined$toDataFrame()` has changed.
+- `DataCombined$toDataFrame()` shows the data in the order as the data sets were 
+added and not alphabetically sorted.
+- `DataCombined$removeGroupAssignment()` does not show a warning if specified name is
```
PavelBal, line 25: Why should it, don't see a reason (btw, I dont like warnings at all)

NEWS.md
```diff
@@ -19,6 +19,13 @@ and https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1087.
...
+not present in the `DataCombined`.
+- `DataCombined$removeGroupAssignment()` does not produce an error if specified
```
PavelBal, line 27: Why should it?

R/data-combined.R
```diff
@@ -108,24 +109,23 @@ DataCombined <- R6::R6Class(
...
       self$setDataTransformations(names)
 
-      if (!is.null(groups)) {
-        self$setGroups(names %||% datasetNames, groups)
-      }
+      private$.updateGroups(names, groups)
 
+      # Set data type
+      for (name in names) {
+        private$.dataType[[name]] <- "observed"
```
Yuri05, line 127: don't we have an enum for data types?

R/data-combined.R
```diff
@@ -191,17 +184,20 @@ DataCombined <- R6::R6Class(
...
       self$setDataTransformations(names)
 
-      if (!is.null(groups)) {
-        self$setGroups(names %||% pathsNames, groups)
-      }
+      private$.updateGroups(names, groups)
 
+      # Set data type
+      for (name in names) {
+        private$.dataType[[name]] <- "simulated"
```
Yuri05, line 199: same: enum

R/data-combined.R
```diff
@@ -589,202 +613,53 @@ DataCombined <- R6::R6Class(
...
+      data <- dplyr::group_by(data, name) %>%
+        # using `unique(name)` because we have grouped by the name, so it is
+        # always unique
         dplyr::mutate(
-          xValues = (xRawValues + xOffsets) * xScaleFactors,
-          yValues = (yRawValues + yOffsets) * yScaleFactors,
-          yErrorValues = yRawErrorValues * abs(yScaleFactors)
+          xValues = (xValues + dataTransformations[dataTransformations$name == unique(name), ]$xOffsets) * dataTransformations[dataTransformations$name == unique(name), ]$xScaleFactors,
+          yValues = (yValues + dataTransformations[dataTransformations$name == unique(name), ]$yOffsets) * dataTransformations[dataTransformations$name == unique(name), ]$yScaleFactors,
+          yErrorValues = yErrorValues * abs(dataTransformations[dataTransformations$name == unique(name), ]$yScaleFactors)
```
Yuri05, line 648: is it worth to define `dataTransformations[dataTransformations$name == unique(name), ]` as a function?
only in 3 lines above it is used 5 times

R/data-combined.R
PavelBal: No, and at the moment I wouldn't know where the user coud use the enum.

R/data-combined.R
PavelBal: I don't know how this works together with `mutate`, maybe @Felixmil can help.

Yuri05: not a big deal, was just a question.

Felixmil: I tried some other strategy for this part of the code:

## Join approach: 
Use a join to bring values from the `self$dataTransformation` table to the original data and reference them using just their names:
```r
      data <- data %>%
        dplyr::left_join(self$dataTransformations, by = "name") %>%
        dplyr::group_by(name) %>%
        # using `unique(name)` because we have grouped by the name, so it is
        # always unique
        dplyr::mutate(
          xValues = (xValues + xOffsets) * xScaleFactors,
          yValues = (yValues + yOffsets) * yScaleFactors,
          yErrorValues = yErrorValues * abs(yScaleFactors)
        )
```

This way, the code is way easier to understand

## `{data.table}` approach

Perform a join with 
```r
      dataTransformations <- self$dataTransformations
      # Transform values in data frame
      data <-
        setDT(data)[setDT(dataTransformations),
                          `:=`(xValues = (xValues + xOffsets)*xScaleFactors,
                               yValues = (yValues + yOffsets) * yScaleFactors,
                               yErrorValues = yErrorValues * abs(yScaleFactors)),
                          on=.(name)]
      data <- tibble::as_tibble(data)
```
Here the transformation code is also easy to understand, but the `{data.table}` syntax is not.  However, as often when using `{data.table}`, this should increase performances.

## Performances
I expanded the test table from Pavel's example 1000 times, so `data` is a `1 250 000`  rows table. Here are the results:

```
Unit: milliseconds
                              expr      min        lq      mean   median       uq      max neval
      private$.dataTransform(data)  96.0567 108.04465 130.04788 116.3166 125.0462 374.7492   200
 private$.dataTransform_join(data) 132.6254 165.77815 237.44497 181.6150 376.0195 430.5048   200
   private$.dataTransform_dt(data)  42.8544  46.43925  76.90983  53.0328  62.7446 299.8599   200
```

image

Shall I push the `{data.table}` implementation ? 😃

Yuri05: wow.
Well, from the _performance_ point of view data.table is clearly the winner.
From the _code understanding_ point of view - @PavelBal should judge :)

And I really wonder why we have this "bimodal" distributions of the execution time in every scenario.
Just out of curiosity: if you execute the functions not 200 but e.g. 10000 times - how do the graphs look like then?

Yuri05: pretty interesting

tests/testthat/_snaps/plot-population-time-profile/multiple-sim-and-obs.svg
Felixmil, line 1: Comparison to this snapshot failed for some reason. 

The diff says it changed but I cannot see any difference. @PavelBal  Do you know what happened ?

PavelBal: Lol, the order of ggplot elments changes.

Felixmil: On my side it is *all* the snapshots that need to be updated :/

PavelBal:
Ok, merging?

Felixmil:
> Ok, merging?

On my side it is all the snapshots that need to be updated :/

PavelBal:
> On my side it is all the timeprofile snapshots that need to be updated.

Weird. AppvVeyor build is passing.

Felixmil:
Yes, it also fails in the`develop` branch on my computer. The problem is on my side. You can merge.

Felixmil (approved):

PR #1314, opened 2023-09-27
setQuantityValuesByPath() returns a better error when provided unit is not valid for the quantity

PavelBal:
Same PR but base on dev


PavelBal:
> Also should't we change the error coming from Core directly? Instead of wrapping it in OSPSuite.R?

@msevestre  First I also thought it, but the error is coming when we are trying to convert unit using 

```
           value <- toBaseUnit(
              quantityOrDimension = dimension,
              values = value,
              unit = units[[i]],
              molWeight = mw
            )
```

that doesn't have the path of the quantity.

PavelBal:
Hi

## Before:

```
Error in rClr::clrCall(dimensionTask, "ConvertToBaseUnit", dimension,  : 
  Type:    OSPSuite.Utility.Exceptions.OSPSuiteException
Message: Unit '' is not defined in dimension 'Volume'.
Method:  Double[] convertToBaseUnit(OSPSuite.Core.Domain.UnitSystem.IDimension, System.String, System.Nullable`1[System.Double], Double[])
Stack trace:
   at OSPSuite.R.Services.DimensionTask.convertToBaseUnit(IDimension dimension, String displayUnit, Nullable`1 molWeight, Double[] valuesInDisplayUnit)
   at OSPSuite.R.Services.DimensionTask.ConvertToBaseUnit(String dimensionName, String displayUnit, Double valueInDisplayUnit)
4.
rClr::clrCall(dimensionTask, "ConvertToBaseUnit", dimension,
sourceUnit, values) at utilities-units.R#192
3.
toUnit(quantityOrDimension = quantityOrDimension, values = values,
targetUnit = baseUnit, sourceUnit = unit, molWeight = molWeight,
molWeightUnit = molWeightUnit) at utilities-units.R#107
2.
toBaseUnit(quantityOrDimension = dimension, values = value, unit = units[[i]],
molWeight = mw) at utilities-quantity.R#165
1.
setQuantityValuesByPath(quantityPaths = parameterPath, values = 100,
simulation = sim, units = "")
```

## After

```
Error in value[[3L]](cond) : 
  Unit '' is not valid for quantity with path 'Organism|Liver|Intracellular|Volume'
Original error message:
Type:    OSPSuite.Utility.Exceptions.OSPSuiteException
Message: Unit '' is not defined in dimension 'Volume'.
Method:  Double[] convertToBaseUnit(OSPSuite.Core.Domain.UnitSystem.IDimension, System.String, System.Nullable`1[System.Double], Double[])
Stack trace:
   at OSPSuite.R.Services.DimensionTask.convertToBaseUnit(IDimension dimension, String displayUnit, Nullable`1 molWeight, Double[] valuesInDisplayUnit)
   at OSPSuite.R.Services.DimensionTask.ConvertToBaseUnit(String dimensionName, String displayUnit, Double valueInDisplayUnit)
6.
stop(paste0(messages$wrongUnitForQuantity(quantityPath = path,
unit = units[[i]]), "\n", "Original error message:\n", e$message)) at utilities-quantity.R#176
5.
value[[3L]](cond)
4.
tryCatchOne(expr, names, parentenv, handlers[[1L]])
3.
tryCatchList(expr, classes, parentenv, handlers)
2.
tryCatch({
value <- toBaseUnit(quantityOrDimension = dimension, values = value,
unit = units[[i]], molWeight = mw)
}, error = function(e) { ... at utilities-quantity.R#168
1.
setQuantityValuesByPath(quantityPaths = parameterPath, values = 100,
simulation = sim, units = "")
```

msevestre:
ok wondering if we actally need the whole stack trace in this case?

R/utilities-quantity.R
```diff
@@ -162,11 +162,28 @@ setQuantityValuesByPath <- function(quantityPaths, values, simulation, units = N
...
+              molWeight = mw
+            )
+          },
+          error = function(e) {
+            stop(paste0(
+              messages$wrongUnitForQuantity(
+                quantityPath = path,
+                unit = units[[i]]
+              ),
+              "\n",
```
msevestre, line 182: is this message useful? it's cleary a conversion error no?

msevestre (commented):
Wodnering if we should in this case make the error a bit less scary
What about the other way around (toUnit?)

R/utilities-quantity.R
PavelBal: I was setting parameter values from excel file using `setParameterValuesByPath()` and it was failing because I forgot to define the unit. With the previous message

> Message: Unit '' is not defined in dimension 'Volume'.

I had to debug to find what exactly was wrong. If I've had the path of the quantity (what is now in the improved message), I would have have much better idea where this is coming from.

PavelBal:
> What about the other way around (toUnit?)

We don't know the path of the quantity in `toUnit()` or `toBaseUnit()`,  so the error message must be issued in a calling function.

> Wodnering if we should in this case make the error a bit less scary

Why is it scary? :D

> ok wondering if we actally need the whole stack trace in this case?

True, I removed the stack and updated the message. Now:

```
Error in value[[3L]](cond) : 
  Unit 'mol' is not valid for quantity with path 'Organism|Liver|Intracellular|Volume' and dimension 'Volume'
```

PR #1315, opened 2023-09-27
Use master branch of PK-Sim

PavelBal:
Since we are using the released DLLs, use PK-Sim from master branch.

Is there anything else to change?

### DIFF
--- a/appveyor-3.6.yml
+++ b/appveyor-3.6.yml
@@ -3,6 +3,7 @@ image: Visual Studio 2019
 # uncomment to use global ospsuite version (and comment line underneath)
 # version: '{build}'
 version: '11.2.{build}'
+pkSim_branch: 'master'
 
 # Download script file from GitHub
 init:
@@ -34,7 +35,7 @@ before_build:
   - nuget sources add -name core -source https://ci.appveyor.com/nuget/ospsuite-core
   - nuget sources add -name centos -source https://ci.appveyor.com/nuget/ospsuite-centos
   - nuget install packages.config -OutputDirectory packages # cannot use automatic artifact publish because it's not recognized in a C++/CLI project
-  - rake "prepare_for_build[%APPVEYOR_BUILD_VERSION%]"
+  - rake "prepare_for_build[%APPVEYOR_BUILD_VERSION%, %pkSim_branch$%]"
 
 build_script:
   - travis-tool.sh install_deps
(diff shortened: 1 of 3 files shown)


msevestre:
Do not change the default in the rake file. Rather pass the. Branch as second parameter (in this case master )

Yuri05:
I don't understand the PR.
With this, nightly builds would be tested with the previously released version of PK-Sim ( so currently 11.2). 
I think this is wrong???

PavelBal:
@Yuri05 I made a PR yesterday to use 11.2 Core version, so it must be 11.2 PK-Sim. We should only switch to new core when the R package would require any new feature (or 12 is released, of course). https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1312

appveyor-3.6.yml
```diff
@@ -34,7 +35,7 @@ before_build:
   - nuget sources add -name core -source https://ci.appveyor.com/nuget/ospsuite-core
   - nuget sources add -name centos -source https://ci.appveyor.com/nuget/ospsuite-centos
   - nuget install packages.config -OutputDirectory packages # cannot use automatic artifact publish because it's not recognized in a C++/CLI project
-  - rake "prepare_for_build[%APPVEYOR_BUILD_VERSION%]"
+  - rake "prepare_for_build[%APPVEYOR_BUILD_VERSION%, %pkSim_branch$%]"
```
PavelBal, line 38: @msevestre Does it work like this?

Yuri05:
> @Yuri05 I made a PR yesterday to use 11.2 Core version, so it must be 11.2 PK-Sim. We should only switch to new core when the R package would require any new feature (or 12 is released, of course). #1312

There were a lot of bugfixes in core since 11.2
E.g. [this fixed issue](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2098) is for me already reason enough to use 12

PavelBal:
Damn. Hmm, obviously, there is no test for it in the R package, otherwise, it would fail. Meeeh, that's not good... Maybe the OSPSuite.R libraries should be a separate project so it is possible to fix something for the R package without using the whole new core?

msevestre:
How can it be a separate project since it's whole purpose is to be a wrapper over core functionalities?

Yuri05:
Not sure what the problem with the usage of Core 12 was/is.
But for now, please roll back the core downgrading and continue using Core 12.
E.g. Reporting Engine relies on features/bug fixes in Core 12 and will not properly work with Core 11.
(this PR would become obsolete then)

PavelBal:
> Not sure what the problem with the usage of Core 12 was/is.

Simulations saved to PKML in R cannot be opened with MoBi V11 any more. And MoBi 12 is too raw to work with.

msevestre:
I think in the future, we should distinguish PR that are R only and R +.net only and we have develop always using develop of core. The other branch is X.Y. ongoing where we only merge the first type of PR

Yuri05:
or maybe we just create **additional nightly build**, which uses master/released versions of Core+PK-Sim but the current (dev) version of the R code.

PR #1316, opened 2023-09-27
Back to using core 12

PavelBal:
Back to 12..


packages.config
```diff
@@ -1,13 +1,13 @@
...
-  <package id="OSPSuite.Infrastructure" version="11.2.155"/>
-  <package id="OSPSuite.Infrastructure.Import" version="11.2.155"/>
-  <package id="OSPSuite.Infrastructure.Autofac" version="11.2.155"/>
-  <package id="OSPSuite.R" version="11.2.155"/>
+  <package id="OSPSuite.Assets" version="12.0.218"/>
+  <package id="OSPSuite.Core" version="12.0.218"/>
+  <package id="OSPSuite.Infrastructure" version="12.0.218"/>
+  <package id="OSPSuite.Infrastructure.Import" version="12.0.218"/>
+  <package id="OSPSuite.Infrastructure.Autofac" version="12.0.218"/>
+  <package id="OSPSuite.R" version="12.0.218"/>
```
Yuri05, line 8: let's use the latest **12.0.226** 😉

dimensions
Yuri05, line 1: why downgrading the dimensions version?

pkparameters
Yuri05, line 1: why downgrading the pk parameters version?

PavelBal: I actually do not understand this change... How do you even know it's a downgrade? The folders "pkparameters" and "dimensions" were marked as "changed" forever and now they slipped into this PR...

packages.config
PavelBal: Is it already used by PK-Sim?

pkparameters
Yuri05: > How do you even know it's a downgrade?

The latest commit in the dimensions repo is currently used by ospsuite-R.
Every change can be only a downgrade :)

PavelBal: Yes that did the trick.. I thought "Update submodules..." does exactly this, but apparently not...

packages.config
Yuri05: PK-Sim currently uses 12.0.223 
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/0b33736c8b7efe83c2faf743fd12a54587d3e4bb/src/PKSim.Core/PKSim.Core.csproj#L31-L39

PavelBal: Updated to 223

Yuri05 (approved):

PR #1318, opened 2023-10-02
Clean dataSetToDataFrame

Felixmil:
+ devtools::document()

I tried to improve this function but it is already well optimized. 

I tried:
- making the lapply loop one level higher,
- use vectors instead of lists,
- make use of data.table.
But all solutions were slower than the original implementation.

So I'm only submitting the cleaning I did.

``` r
devtools::load_all()
#> ℹ Loading ospsuite
#> 


configurationPath <- getTestDataFilePath("dataImporterConfiguration_noSheets.xml")
xlsFilePath <- getTestDataFilePath("CompiledDataSet_oneSheet.xlsx")
importerConfiguration <- loadDataImporterConfiguration(configurationPath)

dataSets <- loadDataSetsFromExcel(xlsFilePath = xlsFilePath, importerConfigurationOrPath = importerConfiguration, importAllSheets = TRUE)


mbm <- microbenchmark::microbenchmark(
  new = dataSetToDataFrame(dataSets),
  old = dataSetToDataFrame2(dataSets),
  times = 20,
  check = "equal"
)

mbm
#> Unit: milliseconds
#>  expr      min       lq     mean   median       uq      max neval
#>   new 220.2441 228.7395 235.0186 230.8486 237.8563 270.5133    20
#>   old 223.6831 231.0054 240.9414 235.9187 251.9565 266.5035    20

ggplot2::autoplot(mbm)
```

<sup>Created on 2023-10-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


tests/testthat/_snaps/plot-individual-time-profile/lloq.svg
PavelBal, line 1: Why did the snapshot change? Was it wrong before?

Felixmil: The difference comes from the last point which now is below 0.
This is identical on the `develop/` branch...

PavelBal (approved):

PR #1326, opened 2023-11-09
Clean R CMD Check warnings

### LINKED ISSUE #1325: R CMD Check Warnings
## Symptoms

Here is the warning that is repeated several time during `R CMD Check` execution:

```
 Error: .onLoad failed in loadNamespace() for 'ospsuite', details:
   call: rClr::clrCallStatic("OSPSuite.R.Api", paste0("Get", taskName))
   error: Type:    RDotNet.EvaluationException
 Message: Error: object '.slotNames' not found
 
 Method:  RDotNet.SymbolicExpression Parse(System.String, System.Text.StringBuilder, RDotNet.REnvironment)
 Stack trace:
    at RDotNet.REngine.Parse(String statement, StringBuilder incompleteStatement, REnvironment environment)
    at RDotNet.REngine.<Defer>d__59.MoveNext()
    at System.Linq.Enumerable.LastOrDefault[TSource](IEnumerable`1 source)
    at RDotNet.REngine.Evaluate(String statement, REnvironment environment)
    at RDotNet.S4Object..ctor(REngine engine, IntPtr pointer)
    at RDotNet.SymbolicExpressionExtension.AsS4(SymbolicExpression expression)
    at Rclr.RDotNetDataConverter.CreateClrObj(Object obj)
    at Rclr.RDotNetDataConverter.ConvertToR(Object obj)
    at Rclr.ClrFacade.marshallDataToR(Object obj, Boolean tryUseConverter)
    at Rclr.ClrFacade.Cal
```

This is a problem because this will stop `R CMD Check` verifications steps before it can detect **real** warnings and errors.

## Cause

I investigated (with the help of @georgeDaskalakis) and found out that this is caused by this function:
```r
.initializeDimensionAndUnitLists <- function() {
  # This initializes the two lists in the parent environment which is the package environments
  ospDimensions <<- .getDimensionsEnum()
  ospUnits <<- .getUnitsEnum()
}
```
In particular the `<<-` used here.

.getDimensionsEnum() and .getUnitsEnum() both call `rClr::clrCall()` at some point:

```r
.getDimensionsEnum <- function() {
  enum(allAvailableDimensions())
}

allAvailableDimensions <- function() {
  dimensionTask <- .getNetTaskFromCache("DimensionTask")
  rClr::clrCall(dimensionTask, "AllAvailableDimensionNames")
}
```

During `R CMD Check` the package is loaded several times. Each time the package is loaded, `ospDimensions` and `ospUnits` are initialized by calling a .NetTask and **overwrites** the previous values. 
I'm not really sure of what happens here, this problem 

### Make them functions instead of variables

I was able to avoid this warning by turning these variables into function. This way, the .NetTask is only called when necessary (instead of each time the package is loaded).

```r
ospDimensions <- function(){
  .getDimensionsEnum()
}

ospUnits <- function(){
  .getUnitsEnum()
}
```

The downside is that we'd now need to use `ospDimensions ()` and `ospUnits()` (notice the `()`) to access the values _i.e_ ` ospUnits$Time$h` would become ` ospUnits()$Time$h` and this would be a breaking change for other packages and existing scripts.

### Make use of package internal data

`ospUnits` and `ospDimensions` can be defined as internal data with:
```r
ospDimensions <- .getDimensionsEnum()
ospUnits <- .getUnitsEnum()
usethis::use_data(ospDimensions, ospUnits, internal = TRUE)
```
This will save the object as `sysdata.rda` in `R/` folder.
Then, when the package is loaded, these objects will be available in the package namespace just like before.
The downside is that this small script should be run each time the .dll core is updated (with potential new units / correction)

Felixmil:
Fixes #1325 and get rid of several Warnings and Notes from R CMD Check


Felixmil:
I think the build fails because in AppVeyor:
```
using session charset: ISO8859-1
```

while on local:
```
─  using session charset: UTF-8
```

R/ospsuite-env.R
```diff
@@ -27,7 +27,7 @@ ospsuiteEnv$numberOfCores <- function() {
 ospsuiteEnv$showProgress <- FALSE
 
 # Specifies the symbol used for µ. This will be set by the .NET layer
```
Yuri05, line 29: _µ_ in the comment line above is OK for R CMD check?

Felixmil: Apparently... yes: https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r-r99ji/builds/48493684#L1205

BinaryFiles
Yuri05, line 1: where is this new file used?

Felixmil: It is used by `R CMD Check`

Felixmil: This file is directly used by `R CMD Check`

R/utilities-units.R
```diff
@@ -372,26 +372,105 @@ getDimensionByName <- function(name) {
...
+      # if unit_name equals "" replace by Unitless
+      if (unit_name == "") {
+        unit_name <- "Unitless"
+      }
+      unit_list[[unit_name]] <- unit_name
+    }
+    ospUnits[[dim_name]] <- unit_list
+  }
+
+  ospUnits[["Dimensionless"]] <- list("Unitless" = "Unitless")
```
Yuri05, line 460: why do we actually replace the empty unit with this funny "Unitless"?
it seems to be unused at the moment (at least I could not find any usage of "Unitless" within the package), but if we would pass e.g. `ospUnits$Dimensionless$Unitless` to OSPSuiteCore it would crash(??)

Yuri05: currently:
```
> ospUnits$Dimensionless$Unitless
[1] "Unitless"
> ospUnits$Fraction$Unitless
[1] "Unitless"
```

my expectation:
```
> ospUnits$Dimensionless$Unitless
[1] ""
> ospUnits$Fraction$Unitless
[1] ""
```

BinaryFiles
Yuri05: ok, thx

R/utilities-units.R
Felixmil: I really only replicated the output of the previous `.getUnitsEnum()`.

But my guess would be that for plots, displaying "" could be misunderstood for a bug when the user needs to show units.

Yuri05: ok, created a new issue for that: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1328

R/ospsuite-env.R
Yuri05: hmm, ok

Yuri05 (approved):

Issue #1246, opened 2023-04-27
Print parameter: small values are displayed as 0.00


Yuri05:
[simulation.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/files/11345697/simulation.zip)
```
> library(ospsuite)
> sim=loadSimulation("sim.pkml")
> p=getParameter("Neighborhoods|Brain_pls_Brain_int|Alfentanil|P (plasma<->interstitial)", sim)
> p
Parameter: 
   Path: Neighborhoods|Brain_pls_Brain_int|Alfentanil|P (plasma<->interstitial) 
   Value: 0.00 [dm/min]   <------------------------ here I would expect e.g. the scientific notation, as in p$value below
   isFormula: TRUE 
   formula: P/2 
   Value overrides formula: FALSE 
   isStateVariable: FALSE 
> p$value
[1] 1.549603e-05
```

PavelBal:
Fixes with https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1330

Issue #1123, opened 2022-09-15
Weird LLOQ printing


pchelle:
When trying to set and check an LLOQ value in `DataSet` objects, 
The value is oddly printed

```r
> dataSet <- DataSet$new(name = "test")
> dataSet$LLOQ <- 1
> print(dataSet)

DataSet: 
   Name: test 
   X dimension: Time 
   X unit: h 
   Y dimension: Concentration (mass) 
   Y unit: mg/l 
   Error type: 
   Error unit: 
   Molecular weight: 
   LLOQ: 0.999999997475243
   Meta data: 
list()
```

while the value seems correct
```r
> print(dataSet$LLOQ)
[1] 1
```

PavelBal:
Lol, it's the way how R (studio?) prints things in the console.

The **actual** value of `dataSet$LLOQ` is 0.99999... (see ".Last.value), print to console rounds it up. 

![image](https://user-images.githubusercontent.com/25903040/190593182-7086da7b-ee2c-4d4d-ae18-5dbc809f404b.png)

0.9999... instead of 1 arises from unit conversion. When the user sets 1 as LLOQ, it is supposed to be "mg/l", and the value is internally converted to the base unit of the dimension... with numerical imperfection. Nothing we can do here.

Yuri05:
With the default printing setting of 7 (R default) I would still expect that the print function of the dataset rounds correctly.
E.g. with the standrad print function:

```
> getOption("digits")
[1] 7
> print(0.999999997475243)
[1] 1
```

But with the dataset still:
```
> print(dataSet)
DataSet: 
   ...
   LLOQ: 0.999999997475243 
   ...
```

Why?

Yuri05:
I think the error is in the **printline** function:

https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/blob/409d2382eb49f04e9e332a126452b89b0cc2a6e5/R/printable.R#L40-L42

It just prints the value via `cat`. 
But it should check first if the value is numeric and e.g. convert it to string using the `format `function. 
Then the output would be as expected.

```
> format(dataSet$LLOQ)
[1] "1"
```

PR #1330, opened 2023-11-17
Update print of quantity to include scientific notation in some cases

### LINKED ISSUE #1329: Tutorial examples for setParametersValue not "explicit"
As I was following the introduction to the R package, I found two examples not "helpfull" to understand as a beginner: 

1) 
image
From what I understand, we don't see any change, as the dose of 350 mg is too small to be shown in kg. However, I would suggest: 

- A first example where we can see the impact of the function 
- A comment saying why we don't see any difference 

It might be also interesting to specify that it is equivalent to simply assigning a new $value 
ex: `doseParam$value <- toBaseUnit(doseParam, 350, "mg")`

2)
image

As the dose is set at 0.00 kg at first, the "scaling" gives us a value of still 0.00
I think it is not super explicit to understand the behavior of this function 

Another remark is that I find (as a beginner) confusing to have a rounded value appear. For example, I would have expected to find something like 3.5* 10E-4 kg in the first case. If the value is rounded to 0.00 kg, then there is no way to know if we are really at 0, or just smaller than 0.01 kg.

Laura-Villain:
Fixes #1329 #1246 and #1123

### DIFF
--- a/R/quantity.R
+++ b/R/quantity.R
@@ -111,7 +111,7 @@ Quantity <- R6::R6Class(
       self$printQuantityValue(self$name)
     },
     #' @description
-    #' Print the the value and unit of the quantity
+    #' Print the the value (in scientific notation with 2 digits when needed) and unit of the quantity
     #' @param  caption Text to prepend to the value
     printQuantityValue = function(caption) {
       if (self$value >= 10000 | self$value < 0.01) {
--- a/R/quantity.R
+++ b/R/quantity.R
@@ -111,7 +111,7 @@ Quantity <- R6::R6Class(
       self$printQuantityValue(self$name)
     },
     #' @description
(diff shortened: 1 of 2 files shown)


R/quantity.R
```diff
@@ -111,7 +111,7 @@ Quantity <- R6::R6Class(
       self$printQuantityValue(self$name)
     },
     #' @description
-    #' Print the the value and unit of the quantity
+    #' Print the the value (in scientific notation with 2 digits when needed) and unit of the quantity
```
PavelBal, line 114:
```suggestion
    #' Print the value (in scientific notation with 2 digits when needed) and unit of the quantity
```

PavelBal (commented):
-Please reference issues that will be fixed by this PR (Fixes #xxxx)

PavelBal (approved):

Issue #307, opened 2020-02-21
Get the "negative values allowed" property


PavelBal:
Get access to the value of "negative value allowed" flag of molecules.

PR #1341, opened 2023-12-20
Steady state

PavelBal:
Adds Steady state calculation/export (compare discussions in https://github.com/Open-Systems-Pharmacology/Forum/discussions/1450) originally implemented in `esqlabsR`.

We also have a function to export such a "steady state" to excel file that can be loaded in MoBi, would it be also of interest for the main package?

```
#' Export steady-state to excel
#'
#' @details Simulates a given model to its steady-state and creates an
#' Excel-file with the end values of molecules amounts in all containers and
#' parameter values that have a right-hand-side (state variable parameters).
#' @param resultsXLSPath Path to the xls-file where the results will be written
#'   to. If the file does not exist, a new file is created. If no path is
#'   provided, the file will be created in the same directory where the model
#'   file is located. The name of the file will be `<SimulationFileName>_SS`.
#' @param simulation A `Simulation` object that will be updated with the steady
#' state
#' @inheritParams ospsuite.parameteridentification::getSteadyState
#' @import ospsuite.utils ospsuite.parameteridentification
#' @export
exportSteadyStateToXLS <- function(simulation,
                                   quantitiesPaths = NULL,
                                   resultsXLSPath = "",
                                   steadyStateTime = 1000,
                                   ignoreIfFormula = TRUE,
                                   stopIfNotFound = TRUE,
                                   lowerThreshold = 1e-15,
                                   simulationRunOptions = NULL)
```


R/utilities-simulation.R
```diff
@@ -749,3 +749,230 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+#' simPath <- system.file("extdata", "Aciclovir.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' steadyState <- getSteadyState(simulations = sim)
+#' # Set initial values for steady-state simulations
+#' setQuantityValuesByPath(
+#'   quantityPaths = steadyState[[sim$id]]$paths,
+#'   values = steadyState[[sim$id]]$values, simulation = sim
+#' )
+getSteadyState <- function(quantitiesPaths = NULL,
+                           simulations,
```
Yuri05, line 800: I would change the order of arguments: first `simulations`, then `quantitiesPaths `

Felixmil (dismissed):
Approved once `pkgdown::build_site()` is run

R/utilities-simulation.R
```diff
@@ -749,3 +749,230 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+                           steadyStateTime = 1000,
+                           ignoreIfFormula = TRUE,
```
Yuri05, line 802: Should we try to select the default simulation time based on the start time of the last application?
Instead of setting  the default value to 1000: set it NULL.
And then estimate the default simulation time as Time-of-last-application + Delta
Where Delta = 1 day or so (I personally would set it even to 3 days)

R/utilities-simulation.R
```diff
@@ -749,3 +749,230 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+    simulation$outputSchema$addTimePoints(timePoints = steadyStateTime[[idx]])
+    # If no quantities are explicitly specified, simulate all outputs.
+    if (is.null(quantitiesPaths)) {
+      quantitiesPathsMap[[idx]] <- ospsuite::getAllStateVariablesPaths(simulation)
+    } else {
+      quantitiesPathsMap[[idx]] <- quantitiesPaths
+    }
+    names(quantitiesPathsMap)[[idx]] <- simId
+    ospsuite::clearOutputs(simulation)
+    ospsuite::addOutputs(quantitiesOrPaths = quantitiesPathsMap[[idx]], simulation = simulation)
```
Yuri05, line 843: It would be nice to have the function `setOutputs(quantityPaths)` in the ospsuite-R, which would just execute the 2 functions above (clearOutputs and addOutputs )

R/utilities-simulation.R
```diff
@@ -749,3 +749,230 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+#' setQuantityValuesByPath(
+#'   quantityPaths = steadyState[[sim$id]]$paths,
+#'   values = steadyState[[sim$id]]$values, simulation = sim
+#' )
+getSteadyState <- function(quantitiesPaths = NULL,
+                           simulations,
+                           steadyStateTime = 1000,
+                           ignoreIfFormula = TRUE,
+                           stopIfNotFound = TRUE,
+                           lowerThreshold = 1e-15,
```
Yuri05, line 804: In the Ospsuite.Core we calculate a Comparison Threshold for each state variable and each observer dynamically, based on the AbsTol and ScaleFactor for the state variables and on the monitor formula for observers; also having special handling for Fractions.
Values below that comparison tolerance are considered numerical noise.
I wonder if we should export this property to R and use as a default for each variable if not passed by user

R/utilities-simulation.R
```diff
@@ -749,3 +749,230 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+#'   molecules and state variable parameters are considered. The same list is
+#'   applied for all simulations.
+#' @param simulations `Simulation` object or a list of `Simulation` objects
+#' @param ignoreIfFormula If `TRUE` (default), species and parameters with
+#'   initial values defined by a formula are not included.
+#' @param stopIfNotFound Boolean. If `TRUE` (default), an error is thrown when
+#'   results for certain species were not generated. This may happen when due to
+#'   numerical problems some values cannot be calculated, though the whole
+#'   simulation converges. Setting this argument to `FALSE` allows to ignore
+#'   such errors. Check the outputs for empty values when using this option.
```
Yuri05, line 777: I don't get this: how can the values for SOME species be not available if the simulation run was successful?

R/utilities-simulation.R
```diff
@@ -749,3 +749,230 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+#' @param lowerThreshold Numerical value (in default unit of the output).
+#' Any steady-state values below this value are considered as numerical noise
+#' and replaced by 0. If `lowerThreshold` is `NULL`, no cut-off is applied.
```
Yuri05, line 780: Strictly speaking, we should also check if the variable allows negative values and replace not the values below the threshold by 0 but only the values in the interval [-Threshold..Threshold]

R/utilities-simulation.R
```diff
@@ -749,3 +749,230 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+    # Get the end values of all outputs
+    endValues <- lapply(quantitiesPathsMap[[simId]], function(path) {
+      # Check if the quantity is defined by an explicit formula
+      isFormulaExplicit <- ospsuite::isExplicitFormulaByPath(
+        path = enc2utf8(path),
+        simulation = simulation,
+        stopIfNotFound = stopIfNotFound
+      )
+
+      if (ignoreIfFormula && isFormulaExplicit) {
```
Yuri05, line 875: I don't get this: how is a state variable supposed to be defined by explicit formula???

R/utilities-simulation.R
PavelBal: Hmm, I don't see an easy way to do this. How would you get all applications? There is a method `allApplicationsFor()`, but then we need the list of all molecules first, and then retrieve all applications and compare... I don't think the complexity is worth it.

What I can imagin for the future, is to run simulation until the derivatives become near 0.

Yuri05: Get all molecules; for each molecule get all applications; for each application get its start time.
Get max of all start times.
```
> sim=loadSimulation("c:/Temp/S_C1C2.pkml")
> sim$allXenobioticFloatingMoleculeNames()
[1] "C1" "C2"
> sim$allApplicationsFor("C1")
[[1]]
Application: 
   Start time: 0.00e+00 [min] 
   Infusion time: 1.00 [min] 
   DrugMass: 182.50 [µmol] 

[[2]]
Application: 
   Start time: 720.00 [min] 
   Infusion time: 1.00 [min] 
   DrugMass: 182.50 [µmol] 

> sim$allApplicationsFor("C2")
[[1]]
Application: 
   Start time: 0.00e+00 [min] 
   Infusion time: 1.00 [min] 
   DrugMass: 182.50 [µmol] 

[[2]]
Application: 
   Start time: 480.00 [min] 
   Infusion time: 1.00 [min] 
   DrugMass: 182.50 [µmol] 
```

R/utilities-simulation.R
PavelBal: PR created

R/utilities-simulation.R
PavelBal: If I remember correctly, it was some observer that was `NaN` the whole time... there was a case for this.

R/utilities-simulation.R
PavelBal: What we are dealing here with is the **start value** of the parameter, which can be an explicit formula.

R/utilities-simulation.R
PavelBal: You are right. For this, https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/307 must be implemented.

R/utilities-simulation.R
Yuri05: or, so with "species" you mean also observers? (I thought only ODE state variables were meant here).
Then this could happen indeed.

R/utilities-simulation.R
Yuri05: ok... but I still don't understand the use case. So if a parameter is a state variable with initial value defined by formula - why would I need to ignore this parameter in the steady state?

R/utilities-simulation.R
Yuri05: Issue create: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1348

R/utilities-simulation.R
PavelBal: Done.

R/utilities-simulation.R
PavelBal: I am not sure this works properly... This is what I wanted to avoid:

image

Amount of an orally administered drug in stomach.

R/utilities-simulation.R
PavelBal: I sometimes define the start value of the parameter as a formula that calculates the steady-state of this parameter based on other intputs. In this case, I would not like to overwrite the start value.

R/utilities-simulation.R
Yuri05: Hmm, it's theoretically done in SimModel:
https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/blob/7092dd785819f9904abbc1f40cc33c8e4be05a30/src/OSPSuite.SimModelNative/src/DESolver.cpp#L350-L364

However when looking into the implementation: only the values in `[-Threshold .. 0]` are set to zero, but **not the values in** `[0 .. Threshold]`
https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/blob/7092dd785819f9904abbc1f40cc33c8e4be05a30/src/OSPSuite.SimModelNative/src/SimulationTask.cpp#L186C1-L193C1

I assume originally the values in [-Threshold .. Threshold] were set to zero, but we got some kind of problems with this (but I don't remember which exactly..). Will create an issue for this in SimModel

Yuri05: s. https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/issues/162

R/utilities-simulation.R
Yuri05: While I still don't understand the use case (maybe you can explain it to me during one of the meetings) - I wonder if we should set the default value of `ignoreIfFormula `to **FALSE**?
It seems pretty dangerous for me when the steady state of an ODE species is ignored because its initial value is not const. but defined by the formula.

PavelBal: I have been thinking about it for some time now, and I am still not convinced, maybe we should discuss it in a meeting.

The way I see it, if a start value of an ODE parameter (and the same goes for molecules, e.g. start values of proteins) is defined by a formula, then this formula represents the steady-state. Otherwise, there is something off with the equation, don't you think?

Yuri05:
@PavelBal Let's create issues for:
* https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1341#pullrequestreview-1794487002
* https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1341#pullrequestreview-1794496717
* https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1341#pullrequestreview-1794510622
After that I would say we can merge this PR and address those issues in further PRs if required

Yuri05 (approved):

PavelBal:
@Yuri05 Can you re-review? I have removed the `stopIfNotFound` argument and had to update the snapshots to match the latest TLF version.

R/utilities-simulation.R
```diff
@@ -749,3 +749,247 @@ getSimulationTree <- function(simulationOrFilePath, quantityType = "Quantity") {
...
+#' @examples
+#' simPath <- system.file("extdata", "Aciclovir.pkml", package = "ospsuite")
+#' sim <- loadSimulation(simPath)
+#' steadyState <- getSteadyState(simulations = sim)
+#' # Set initial values for steady-state simulations
+#' setQuantityValuesByPath(
+#'   quantityPaths = steadyState[[sim$id]]$paths,
+#'   values = steadyState[[sim$id]]$values, simulation = sim
+#' )
+getSteadyState <- function(simulations,
```
Yuri05, line 797: Is there also `setSteadyState()` or `setInitialState()` function somewhere?

Yuri05 (approved):

R/utilities-simulation.R
PavelBal: No, the workflow would be to call

```
#' setQuantityValuesByPath(
#'   quantityPaths = steadyState[[sim$id]]$paths,
#'   values = steadyState[[sim$id]]$values, simulation = sim
#' )
```

as described in the example of `getSteadyState()`. Unless we want to have additional checks or whatever, I do not see a need for a separate function, be we can implement something if required.

PR #1363, opened 2024-02-20
Add function `exportSteadyStateToXLS()`

PavelBal:


DESCRIPTION
```diff
@@ -32,7 +32,8 @@ Imports:
     rlang,
     tlf (>= 1.5.0),
     data.table,
-    xml2
```
PavelBal, line 35: Are we using some package for writing excel files somewhere?

R/utilities-simulation.R
```diff
@@ -993,3 +993,140 @@ getSteadyState <- function(simulations,
...
+  data <- list("Molecules" = speciesInitVals, "Parameters" = parameterInitVals)
+  # If the provided path to the output file targets a non-existent directory,
+  # try to create the directory
+  resultsDir <- dirname(resultsXLSPath)
+  if (!file.exists(resultsDir)) {
+    dir.create(resultsDir, recursive = TRUE)
+  }
+
+  # Write the data into an excel file.
+  writexl::write_xlsx(data, path = resultsXLSPath, col_names = TRUE)
```
PavelBal, line 1131: This can be changed if other package for writing excel is prefered.

Felixmil:
I would avoid exporting data to **Excel** files to keep the package truely open (you need a private licence/subscription to use excel files). Especially if no tabs are required.

Instead I'd suggest a simple .csv export that is truely open, less prone to formating errors and is still compatible to be opened in excel (or any text editor).

PavelBal:
The main purpose of this function is to export the values so they can be loaded in MoBi, and MoBi loads initial values from Excel files.

Felixmil (dismissed):
Approved if writexl package is part of the validated ones on your side @Yuri05

Yuri05:
> Approved if writexl package is part of the validated ones on your side @Yuri05

as for now: it is NOT
I will check if it is planned for the near future. If not: we will add a request for this. 
But for now: please don't use the package.

PavelBal:
> > Approved if writexl package is part of the validated ones on your side @Yuri05
> 
> as for now: it is NOT I will check if it is planned for the near future. If not: we will add a request for this. But for now: please don't use the package.

Is there any other package for writing excel that is part of valid r?

msevestre:
I don’t get why we don’t ask user to save the csv as excel file instead 
Or just import from csv in mobi

PavelBal:
> I don’t get why we don’t ask user to save the csv as excel file instead

Why force the user to do some additional manual steps only because the R package for writing excel files is not validated yet? The current workflow is very straightforward - run the function, load in MoBi. If I manually have to save the csv as excel, that would require way too many steps (90% of cases excel will mess up the csv and simply saving a file to a new location from excel is a pain in the new versions).

Yuri05:
> Is there any other package for writing excel that is part of valid r?

The 3 packages below allow writing Excel files and are already included into the validated R. 
I would vote for openxlsx (s. more info on the packages below)
* **openxlsx**
* xlsx 
* gdata 

<details><summary><b>Information on R packages which allow writing excel files</b></summary>
<p>

1. **`write.xlsx`** from the `openxlsx` package
   - Provides functionality for writing both .xlsx files and .xlsm files with formatting.
   
2. **`write.xlsx`** from the `xlsx` package
   - This package allows for writing .xlsx files but requires Java.

3. **`write_xlsx`** from the `writexl` package
   - A package that does not have any dependencies, such as Java, which is needed for the `xlsx` package.

4. **`write_excel_csv`** from the `readr` package
   - While not strictly an Excel format, `readr` can write CSV files that Excel can open.

5. **`write.xls`** from the `gdata` package
   - This function can write .xls files and also requires Perl to be installed.

6. **`write.xlsx2`** from the `openxlsx` package
   - Another function from `openxlsx`, `write.xlsx2` is designed to be faster than `write.xlsx` for large datasets but with fewer features.

7. **`write_xlsx`** from the `rio` package
   - The `rio` package is an all-in-one package that can import/export data in many file formats, including .xlsx.

8. **`write`** function from the `R2Excel` package
   - This package provides various functions to write data to Excel files, and to add information to existing files.

9. **`XLConnect`** package
   - Offers comprehensive functionality for reading, writing, and manipulating Excel files. It requires Java.

10. **`excel.link`** package
    - Provides an interface to communicate with Excel using COM (Component Object Model) interface, allowing for reading and writing of Excel files among other functionalities.

Please note that the availability and functionality of these packages may have changed since my last update, and new packages might have been introduced. Always check the CRAN repository or R documentation for the most up-to-date information on R packages.

</p>
</details>

PavelBal:
Using `openxlsx` now.

PavelBal:
@Yuri05 Is it okay to merge now?

PavelBal:
Build passing again :)

Yuri05 (approved):

Issue #1365, opened 2024-02-23
`OutputSchema$endTime` does not consider single points


PavelBal:
`OutputSchema$endTime` should return the end time of the simulation. However, it does not consider single points, only time intervals.

Load simulation and inspect the time schema:
```
sim <- loadSimulation(system.file("extdata", "Aciclovir.pkml", package = "ospsuite"))
sim$outputSchema
```
```
OutputSchema: 
Interval: 
   Name: Simulation interval high resolution 
   Start time: 0.00e+00 [min] 
   End time: 15.00 [min] 
   Resolution: 1.00 [pts/min] 
Interval: 
   Name: Simulation Interval 1 
   Start time: 15.00 [min] 
   End time: 1440.00 [min] 
   Resolution: 0.33 [pts/min] 
Interval: 
   Name: Simulation Interval 2 
   Start time: 120.00 [min] 
   End time: 1440.00 [min] 
   Resolution: 0.07 [pts/min] 
```

Print end time:

```
sim$outputSchema$endTime
[1] 1440
```

Add a time point 1445 minutes and print end time again:

```
sim$outputSchema$addTimePoints(1445)
sim$outputSchema$endTime
```

The output is still 1440, which is not correct.

PR #1366, opened 2024-02-23
Do not clear output intervals in steady-state simulatoin

PavelBal:
I realized that clearing output intervals might break some events, as their conditions will not be checked.

- Added workaround for https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1365
- Added internal function `.setEndSimulationTime()`.
- Steady state simulation do not clear output intervals but sets the end time to the provided. Otherwise, some events might not fire.


PavelBal:
@msevestre @Yuri05 an someone review please :)

Yuri05:
@PavelBal Can you also adjust the test snapshots? Builds are failing for a while because of that
(s. https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/507#issuecomment-1970702261)

PavelBal:
rutils cannot be downloaded for whatever reason... @Felixmil do you have an idea?

Felixmil:
@PavelBal I'm pretty sure the artifact is outdated and was deleted by Appveyor: last build was 3 months ago and I think this is the lifespan of artifact in there.

PavelBal:
True... @Yuri05 , can you trigger the build of develop branch again?

Yuri05:
> True... @Yuri05 , can you trigger the build of develop branch again?

@PavelBal [Done 🙂](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/assets/25061876/0d12e5e4-89a1-4c33-ac1f-e46ea2e8acf2)

PavelBal:
@Yuri05 Danke! Build is passing now ;)

R/utilities-output-schema.R
```diff
@@ -77,3 +77,50 @@ clearOutputIntervals <- function(simulation) {
...
+          endTime = min(outputInterval$endTime$value, endTime),
+          resolution = outputInterval$resolution$value,
+          intervalName = outputInterval$name
+        )
+      }
+    }
+
+    # Don't forget to add latest time
+    simulation$outputSchema$addTimePoints(c(timePoints, endTime))
+  }
```
Yuri05, line 125: Here you could add a check that either at least 1 output interval or at least 1 (2?) single output time points are in the final output schema (otherwise the simulation will not produce any results).

Also what would happen if the simulation has e.g. the only output interval [10h .. 20h] and I set the end time to 5h?

PavelBal: > Here you could add a check that either at least 1 output interval or at least 1 (2?) single output time points are in the final output schema (otherwise the simulation will not produce any results).

I think this is not required because `endTime` will always be a valid numeric value, and it is always added to the time schema. But I added a validation that it is >0.

> Also what would happen if the simulation has e.g. the only output interval [10h .. 20h] and I set the end time to 5h?

The output interval is removed, only one point 5h is retained. Added a test for this.

Yuri05 (approved):

PR #1377, opened 2024-03-13
Change data to cover wider log scale

### LINKED ISSUE #1375: Some snapshots broke after TLF update
E.g. this snapshot lost its axes: 

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/tests/testthat/_snaps/plot-observed-vs-simulated/lloq-vertical.svg

It was in this PR: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1368

Felixmil:
partially fixes #1375 

It appears that axis lines are not drawn if the plot does not contain any ticks labels.

It is possible that there are no ticks labels because `{tlf}` forces ticks labels for log scale (1, 10 ...) as vanilla ggplot2 would simply adjust the labels depending on the data in the plot.

In the case of these snapshots, the data points were to far away of `{tlf}` predefined log scale labels so they weren't plotted, inducing the missing axis lines.

I simply extended the test data range to restore the axis but it this bug was to be fixed, it should be done in `{tlf}` directly.


PavelBal:
I would rather not change anything in this package. Let's keep the old data so we are aware of this bug.

PR #1382, opened 2024-03-29
Deprecate `runSimulation()`

### LINKED ISSUE #1380: Having both `runSimulation` and `runSimulations` is confusing
Several user feedback this is confusing and not really usual in the vectorized paradigm of R. 

I think `runSimulations()` should be default and always return a named list of results but with better names because right now most use cases I saw use `[[index]]` to access simulation results, which makes code more complicated to understand.

Also, are these `id` really essential ? Couldn't we just use the name of the Simulation object to name the list ?

Here is an example of what I suggest:

```r
simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
sim1 <- loadSimulation(simPath)
sim2 <- loadSimulation(simPath)

results <- runSimulations(sim1)
results
#> $sim1
#> SimulationResults: 
#>    Number of individuals: 1 

results <- runSimulations(list(sim1, sim2))
results
#> $sim1
#> SimulationResults: 
#>    Number of individuals: 1 
#> 
#> $sim2
#> SimulationResults: 
#>    Number of individuals: 1 
```

Felixmil:
fixes #1380 

``` r
devtools::load_all()
#> ℹ Loading ospsuite
#> 
sim <- loadSimulation(system.file("extdata", "Aciclovir.pkml", package = "ospsuite"))

res <- runSimulation(sim)
#> Warning: `runSimulation()` was deprecated in ospsuite 12.0.0.
#> ℹ Please use `runSimulations()` instead.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
```

<sup>Created on 2024-03-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


Felixmil:
Once review is done:
- [ ] add NEWS entry

DESCRIPTION
```diff
@@ -33,7 +33,8 @@ Imports:
     tlf (>= 1.5.0),
     data.table,
     xml2,
-    openxlsx
+    openxlsx,
+    lifecycle
```
PavelBal, line 37: @Yuri05 is this part of validated R?

PavelBal:
Approved if the new package is allowed.

We could go further and make all functions that require a `SimulationResults` as input also work with lists, sun `simulationResults[[1]]` is not required. Functions like `toDataFrame` could proceed all simulation results and also return a list, other functions/methods should check if the list contains only one result or multiple.

Yuri05:
> Approved if the new package is allowed.
> 
> We could go further and make all functions that require a `SimulationResults` as input also work with lists, sun `simulationResults[[1]]` is not required. Functions like `toDataFrame` could proceed all simulation results and also return a list, other functions/methods should check if the list contains only one result or multiple.

would be nice.
writing` [[1]] `all the time for simResults is annoying

Felixmil:
@Yuri05 is `{lifecycle}` approved for you internal use ?

Felixmil:
CI fails because of #1384

PavelBal:
Okay, we'll have to wait until this has been resolved...

PavelBal (approved):

PR #1389, opened 2024-04-11
Allow rClr versions >0.9.1

### LINKED ISSUE #1388: 11.3 release cannot be installed under R >=4.2
This is because it requires rClr 0.9.2 but the 11.3 release is bound to rClr 0.9.1

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/e5508a7e575cbfe08bd315572df0e31011bffc7d/DESCRIPTION#L22

PavelBal:
Fixes #1388

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ BugReports:
     https://github.com/open-systems-pharmacology/ospsuite-r/issues
 Depends: 
     R (>= 3.6),
-    rClr (== 0.9.1)
+    rClr (>= 0.9.1)
 Imports: 
     dplyr (>= 1.0.0),
     ospsuite.utils (== 1.4.23),


Yuri05:
@PavelBal Can you please also change the R version to 4.3.1 in appveyor.yml and appveyor-nightly.yml
That's why we missed the problem
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/5187c1276463dd135ef4219f1ae8502853f69bbd/appveyor-nightly.yml#L29

PavelBal:
@Yuri05 should I also delete the 3.6 appveyor file?

Yuri05:
> @Yuri05 should I also delete the 3.6 appveyor file?

let's keep it in the hotfix; we can delete later in develop

PR #1392, opened 2024-04-12
update core

### LINKED ISSUE #1387: Update Core
@rwmcintosh you are good at this.

rwmcintosh:
Fixes #1387


rwmcintosh:
Shouldn't the Appveyor.yml file indicate a v12 build? This is still at 11.2

PavelBal:
When I run the tests locally, I get errors where the individual or population seed is expected to be greater than 0, e.g. here:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/392a13423c7b05c6a4741ee457a1a1d986ce71da/tests/testthat/test-utilities-individual.R#L98

@Yuri05 @msevestre This was never the case before. Did something change in the way how seed is calculated? Is it even correct to expect the seed to be >0?

PavelBal:
I don't get it... Is PK-Sim not installed in at all??

Yuri05:
> Shouldn't the Appveyor.yml file indicate a v12 build? This is still at 11.2

absolutely! actually both appveyor and appveyot-nigthly should use the global ospsuite version now 
(so remove 11.2 / 12.0 and uncomment those lines)

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/392a13423c7b05c6a4741ee457a1a1d986ce71da/appveyor.yml#L4-L6
...
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/392a13423c7b05c6a4741ee457a1a1d986ce71da/appveyor.yml#L10-L11

appveyor-nightly.yml
```diff
@@ -1,13 +1,13 @@
 image: Visual Studio 2019
 
 # uncomment to use global ospsuite version (and comment line underneath)
```
Yuri05, line 3: the comment does not fit anymore 😉

appveyor-nightly.yml
```diff
@@ -1,13 +1,13 @@
...
-# version: '{build}'
-version: '12.0.{build}'
+version: '{build}'
+#version: '12.0.{build}'
 
 # Download script file from GitHub
 init:
 # uncomment to use global ospsuite version
```
Yuri05, line 9: same

PavelBal:
@Felixmil @Yuri05  These are the vignettes that are failing, but why? Vignettes should not be built in the non-nightly build. I don't understand what changed.

Also for example this error:

```
Execution halted
when running code in 'set-values.Rmd'
  ...
   Path: Organism|StateVariable_Parameter 
   Value: 10.00 
   isConstant: TRUE 
   isStateVariable: FALSE 
> stateVariableParam$isStateVariable <- TRUE
  When sourcing 'set-values.R':
Error: Creating a RHS Formula is not supported at the moment. This should be done in MoBi.
Execution halted
when running code in 'table-parameters.Rmd'
```

The code is supposed to produce an error in the vignette, and it runs locally but fails on AppVeyor..

Yuri05:
The error comes from the R CMD Check step **checking running R code from vignettes**:



This step was executed for non-nightly in the past as well and worked before. So it seems that something has changed in the package code/files??


So the 1st question is how to fix (because it fails now in the nightly in the same way).
The 2nd question is if we need to check vignettes in non-nightly. If no: seems that `--no-build-vignettes` is not enough and we must extend **R_CHECK_ARGS** with **--ignore-vignettes** argument.

<details><summary><b>R CMD check options</b></summary>

grafik

</details>

Felixmil:
Here is the git blame traceback.

1. The error comes from the vignette `set-values.Rmd` which did not change in the last 4 years:
image

Looking at the comment just above ("# Switching it on is not supported"), it seems accurate that the vignette should fail.

2. `isStateVariable` is defined is the `parameter.R` file and did not change a long time also:
image

3. messages are defined in ospsuite.utils and this did not change for the last 2 years also:
image

My conclusion is that something changed in some settings (AppVeyor, Knitr) rather than in the code.

A simple fix is to change the chunk options from:
`{r setStateVariableValue}`
to 
`{r setStateVariableValue, error=TRUE}`

This will print the error instead of interrupting execution.

PavelBal:
Could you please do the same for the remaining error (unit conversion).

Felixmil:
The other errors seems to be related to PKSIM not being correctly installed. Everything works fin the in the github actions (see https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1394).

Maybe it's time to move to github actions instead ? ping @Yuri05

PavelBal:
What about the linux builds?

I am strongly in favor of using GA (especially as AppVeyor fails now), but let @Yuri05 review if everything we need is covered by this PR.

Yuri05:
I would suggest moving to GH actions after rSharp is fully integrated (both Windows and Linux versions).

Felixmil:
I can also setup a Linux build through github actions, but I would need a bit of guidance regarding the Linux environment setup.

appveyor.yml
```diff
@@ -19,17 +17,20 @@ install:
...
-  NOT_CRAN: true
-  KEEP_VIGNETTES: false
   R_BUILD_ARGS: --no-build-vignettes --no-manual
   R_CHECK_ARGS: --no-multiarch --no-manual --as-cran --no-build-vignettes
-  R_ARCH: x64
-  R_VERSION: 4.3.1
+  R_PKGTYPE: binary
+  NOT_CRAN: true
+  # Do not update vignettes after each commit. Vignettes are only generated for nightly builds.
+  KEEP_VIGNETTES:
```
PavelBal, line 31: LOL, this parameter must be EMPTY in order to skip vignettes!! https://github.com/krlmlr/r-appveyor

PavelBal (approved):

PR #1394, opened 2024-04-15
Add OSP github actions

Felixmil:
These actions will:
- Run tests and perform test coverage
- Run RCMD Check and generate a windows built package
- Render the Pkgdown website and if the action is triggered from a merge into develop branch, it will upload to dedicated branch (and update the website).

During these, the R environment setup process will install rClr 0.9.2 and latest development versions of OSPsuite R packages dependencies (ospsuite.utils and tlf), the other packages are retrieved from CRAN.

This comes together with the proposition to adopt the releasing process described here: https://dev.open-systems-pharmacology.org/r-development-resources/collaboration_guide#releasing-versions

Appveyor workflow is disabled and website output folder is deleted (PR will not have to contain website update no more).


PavelBal:
@Yuri05 I don't see a reason why this should not be used now... It works ;)

Yuri05:
Questions / Remarks (I must admit I did not have time for the full review, so sorry if the questions are answered by the PR):
1. Is only **windows** package generated or also the linux (ubuntu) package?
2. How is the build number incrementation realized? (which was done by patching the DESCRIPTION file on AppVeyor)?
3. For the website generation, we should create it in a separate branch (e.g. "gh-pages") - otherwise reviewing the PRs will be problematic in the future. Please check how website is created in the reporting engine, which has some advantages:
* Website is created in the separate "gh-pages" branch
* Dev / Release website are automatically created based on the target branch
  * push to develop-branch => update dev website
  * push to main-branch => update release website

PavelBal:
@Felixmil Wher can the pk-sim minimal artefact be stored?

Felixmil:
> @Felixmil Wher can the pk-sim minimal artefact be stored?

Not needed if we implement #1422

Felixmil:
What is missing to merge this PR ?

Yuri05:
> What is missing to merge this PR ?

1. Linux package creation (a MUST)
2. Decision on the website creation (based on the still open discussion "main branch only vs. main and develop branch" - s. above)
3. I would also like to discuss the build number incrementation (would prefer automated build number increase by the GH action (patching of the DESCRIPTION file) vs. the proposed "manual" incrementation)

Felixmil:
> 1. Linux package creation (a MUST)

I can work on that and add a parallel job that check and build linux version.

> 2\. Decision on the website creation (based on the still open discussion "main branch only vs. main and develop branch" - s. above)

One branch `gh-pages` can contain both prod and dev documentation but it relies on the `.9000` version suffix.

> 3\. I would also like to discuss the build number incrementation (would prefer automated build number incrementation by the GH action (patching of the DESCRIPTION file) vs. the proposed "manual" incrementation)

With the proposed process, we only increment version manually when we decide to create a release. Each merged PR still gets a unique id through the git sha. Since with this process we can clearly identify develop versions, released version while still being able to target each "build through unique git sha, I don't see what is the benefit of having a build number. But maybe I am missing something ?

Yuri05:
>One branch gh-pages can contain both prod and dev documentation but it relies on the .9000 version suffix.

I really don't like this `.9000` thing. Every time when you need a release docu, you need to create 2 additional PRs for nothing (remove .9000 to get the release docu, then insert .9000 again to prevent overwriting the release docu with the dev docu next time). From my point of view it's better to have a **separate build triggers** for release docu vs. dev docu. E.g.: 
* Docu mode (dev vs. release) is taken from an environment variable (e.g. like [here](
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/e8a9ccaf75a02c0bf48a2a5067de9d4d8d7a42db/_pkgdown.yml#L6-L7); requires also [this](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/e8a9ccaf75a02c0bf48a2a5067de9d4d8d7a42db/.github/workflows/update-dev-website.yml#L33))
* Website creation GH-actions is triggered by push, pull request and for example by release published request
* The environment variable used by pkgdown is set in the GH-Action either to `devel `(if trigger = push or PR) or to `release `(if trigger = release published) **automatically**

It does not matter than, if we have develop branch or not. But the decision "release vs devel docu" is then done automatically without additional PRs.

Alternatively, the GH-Action could patch the DESCRIPTION file and remove .9000 if the trigger = release published.
But still, it's done automatically then, not manually.

>With the proposed process, we only increment version manually when we decide to create a release. Each merged PR still gets a unique id through the git sha. Since with this process we can clearly identify develop versions, released version while still being able to target each "build through unique git sha, I don't see what is the benefit of having a build number. But maybe I am missing something ?

When I install a dev version of the package - I do not see the SHA of the build; I just see e.g. Version = 12.0 for every dev version of the package, correct? And so if I report a bug, I cannot tell which exactly version version of the package I used - which makes the communication between testers and devs harder (e.g. the bug was fixed in the later build of the package, which I missed to install etc.)

Felixmil:
> When I install a dev version of the package - I do not see the SHA of the build; I just see e.g. Version = 12.0 for every dev version of the package, correct? And so if I report a bug, I cannot tell which exactly build of the package I used - which makes the communication between testers and devs harder (e.g. the bug was fixed in the later build of the package, which I missed to install etc.)

You cannot see the SHA of the build indeed, you only get the version like this:
image

But with frequent enough version incrementations, it is really easy to communicate with the user. Communication/environment replication problem arise when: The user cannot identify clearly the version he is using and when the software doesn't get frequent version bump while its code base is evolving.
Unfortunately, our current release process has these two issues : ospsuite r package dev have been 12.0.0 for more than a year now, but its code base changed a lot since then. 
It get even worse if you take into acount the Core upgrades and the PKSim core: it is currently very difficult to know which version of {ospsuite} core works with which pksim core, and so, very hard to replicate a user environment.


The `.9000` is not only for documentation, it also allows the user to easily see that he has a dev version installed or not.
And these PR are not just about documentation. When `.9000` is removed, we also update the `NEWS` header, add the version tag to the branch. Additionally there forementioned steps are not really done manually, all this is taken care of by one function call:  `usethis::use_version()`.

The workflow presented [here](https://dev.open-systems-pharmacology.org/r-development-resources/collaboration_guide#releasing-versions) takes less than 2 minutes.

I agree that processes should be automated as much as possible. However, in this case, a 2 minute process that we do max 10 times a year doesn't seems that much urgent to fully automate.

Cf:

Yuri05:
> But with frequent enough version incrementations, it is really easy to communicate with the user. Communication/environment replication problem arise when: The user cannot identify clearly the version he is using and when the software doesn't get frequent version bump while its code base is evolving.

Seems we have to agree on disagree.

>Unfortunately, our current release process has these two issues : ospsuite r package dev have been 12.0.0 for more than a year now, but its code base changed a lot since then.

Only if you install directly from the GitHub codebase. If you install from the artifact in appveyor-build - it has the corresponding build number included (e.g. 12.0.123)

>It get even worse if you take into acount the Core upgrades and the PKSim core: it is currently very difficult to know which version of {ospsuite} core works with which pksim core, and so, very hard to replicate a user environment. 

I think it has nothing to do with the build numbers of the R package. As soon as we include the "PK-Sim minimal" into the R package - this should be solved.

> The `.9000` is not only for documentation, it also allows the user to easily see that he has a dev version installed or not.
> And these PR are not just about documentation. When `.9000` is removed, we also update the `NEWS` header, add the version tag to the branch. Additionally there forementioned steps are not really done manually, all this is taken care of by one function call: `usethis::use_version()`.

OK.



> The workflow presented [here](https://dev.open-systems-pharmacology.org/r-development-resources/collaboration_guide#releasing-versions) takes less than 2 minutes.

Not for me 😆 
By the way, I would replace "should" with "must" in the docu (created an [issue for this](https://github.com/Open-Systems-Pharmacology/developer-docs/issues/39))
grafik

Felixmil:
Remaining todos after 2024/07/03 discussion:

- [ ] Add an autoincrementation of dev version identifier when a PR is done (if is dev version),
- [ ] Include Linux build generation,
- [ ] Update release process documentation: https://github.com/Open-Systems-Pharmacology/developer-docs/issues/39
- [ ] Delete main branch
- [ ] Rename develop to main

Yuri05:
> Delete main branch

I would suggest not to delete the main branch, but rather to rename it (e.g. `archived_main` or so) and then set to locked.

.github/workflows/main-workflow.yaml
```diff
@@ -0,0 +1,43 @@
...
+  pull_request:
+    branches: [main, develop]
+
+
+jobs:
+  bump-dev-version: # only do that when actually merging in main/develop branch
+    if: github.event_name != 'pull_request'
+    uses: ./.github/workflows/bump-dev-version.yaml
+  R-CMD-Check:
+    if: ${{ always() }}
```
Yuri05, line 15: Here and in other steps: why would we need this `if (true) `condition?

Felixmil: All steps where we have `needs: [bump-dev-version]` would not be ran if `bump-dev-version` step is not executed (which is the case when we only push to the branch).

Yuri05: ok, thx for the explanation :)

PavelBal:
I think we should not put any more effort in making this work with rClr and Ubuntu 18 and focus directly on rSharp. So I propose to base the PR on the rSharp branch.

Felixmil:
> I think we should not put any more effort in making this work with rClr and Ubuntu 18 and focus directly on rSharp. So I propose to base the PR on the rSharp branch.

Maybe we can merge this, then create another branch based on rSharp implementation ?

Yuri05:
> Maybe we can merge this, then create another branch based on rSharp implementation for the linux build ?

For me is OK to merge for the moment AS IS.

For rSharp we already have the branch **r-sharp-r6** which creates both Windows and Ubuntu22 builds.
The Ubuntu package is created in Windows build environment (AppVeyor build script unzips the windows package, replaces the Windows versions of C++ libs by their Linux counterparts and archives the package into .tar.gz again). 
As discussed yesterday, it would be better to create the Linux version of the package in the Linux build environment (which would then execute all the tests and package checks there)

Felixmil:
I don't have the administration rights to manage branches. Can you take care of this part @Yuri05  ?

PavelBal (approved):

PR #1397, opened 2024-04-17
- Update R version used by appveyor to 4.3.3

PavelBal:
No idea if it will fix the failing unicode vignette, but lets try

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ install:
 platform: x64
 
 environment:
-  R_VERSION: 4.3.1
+  R_VERSION: 4.3.3
   R_ARCH: x64
   USE_RTOOLS: true
   R_BUILD_ARGS: --no-build-vignettes --no-manual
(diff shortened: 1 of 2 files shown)


Felixmil:
What the hell, it worked ?

Yuri05:
<details><summary>I always knew it...</summary>
 = </font> 
</details>

Felixmil:
More like:
 = </font>

PavelBal:
I give up.

Yuri05:
> I give up.

<details><summary>Never ever give up 😄 </summary>

</details> 

For now I would suggest to rewrite the test avoiding the problematic units.
For later: maybe we need to do the same for ² etc. what we did for µ.
R+Unicode+Windows is a nightmare

appveyor.yml
```diff
@@ -5,6 +5,10 @@ version: '{build}' # comment to use specific version and uncomment the next line
 
 # Download script file from GitHub
 init:
+  # Trying to enforce unicode support
+  - ps: Set-WinSystemLocale en-US.UTF-8
+  - ps: Start-Sleep -s 5; Restart-Computer
+  - ps: Start-Sleep -s 5
```
Yuri05, line 11: this is pretty insane.
Not sure what will happen on AppVeyor 😆

PavelBal: Boom!

Apparently, this is supposed to work. But setting the locale does not.

PR #1398, opened 2024-04-23
Add documentation how to vary RHS params in SimulationBatch

PavelBal:


PavelBal:
Proposal from @rwmcintosh - add a new argument `stateVariableParams` to the function, so `createSimulationBatch(simulation, parametersOrPaths = NULL, stateVariableParams = NULL, moleculesOrPaths = NULL)`.

@msevestre @Yuri05 what do you think?

Yuri05:
> Proposal from @rwmcintosh - add a new argument `stateVariableParams` to the function, so `createSimulationBatch(simulation, parametersOrPaths = NULL, stateVariableParamsOrPaths = NULL, moleculesOrPaths = NULL)`.
> 
> @msevestre @Yuri05 what do you think?

why not

PavelBal:
> Proposal from @rwmcintosh - add a new argument stateVariableParams to the function, so createSimulationBatch(simulation, parametersOrPaths = NULL, stateVariableParamsOrPaths = NULL, moleculesOrPaths = NULL).

I realized this is not that straightforward. `SimulationBatch$addRunValues()` supports adding values for parameters and initial values of molecules, and RHS params should be in the `molecules` list. If we want to extend the `createSimulationBatch` function by the argument for state variable parameters, `addRunValues` should support this too, but this is implemented in .NET. IMO too much overhead for now.

PavelBal:
@Yuri05 @msevestre @Felixmil pls review

PR #1399, opened 2024-04-23
replace unit conversion example in the vignette to avoid the unicode issue on AppVeyor

Yuri05:
we need to get over the problem and finally get working nightly builds!


Felixmil:
Also run pkgdown::build_site() ?

Yuri05:
> Also run pkgdown::build_site() ?

never managed to run it with ospsuite-R locally - always throws errors.

PR #1419, opened 2024-05-29
Temporary fix vignette

Yuri05:


Yuri05:
@PavelBal 1 check on vignettes was still active. Now should be really disabled.

PavelBal (approved):

PR #1443, opened 2024-07-08
Use GitHub App to push version bumps

Felixmil:
TO DO:

- [x] Create an organization GitHub App (with write access to repo content) + private key
- [x] Add App ID and private key as organization variables (`ACTIONS_MANAGER_APPID` and `ACTIONS_MANAGER_TOKEN`)
- [x] Install App for OSP repositories
- [x] Restore review requirement before merging


.github/workflows/main-workflow.yaml
Yuri05, line 1: now that we don't have the **develop** branch anymore - it can be deleted from the trigger conditions

Yuri05, line 1: Question: how do we prevent that the commit executed by **bump-dev-version** prevents another **on push** workflow triggering?

.github/workflows/main-workflow.yaml
```diff
@@ -11,6 +11,8 @@ jobs:
   bump-dev-version: # only do that when actually merging in main/develop branch
     if: github.event_name != 'pull_request'
     uses: ./.github/workflows/bump-dev-version.yaml
+    secrets: inherit
```
Yuri05, line 14: This inherits the secrets from the organization?

Felixmil: Yes

.github/workflows/main-workflow.yaml
Felixmil: By adding `[skip actions]` in the commit message added by the action.

Felixmil: done in 3ab28a8

Yuri05: Nice. Didn't know it works like this.

.github/workflows/main-workflow.yaml
Yuri05: _Environment variables_ from the organization are inherited automatically?

Yuri05 (approved):

.github/workflows/main-workflow.yaml
Felixmil: Yes:
> To reuse variables in multiple workflows, set them at the organization, repository, or environment levels and reference them using the vars context. For more information see "[Variables](https://docs.github.com/en/actions/learn-github-actions/variables)" and "[Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts#vars-context)."

from [here](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations)

PR #1461, opened 2024-07-23
Move loading of PK-Sim libs to package init

### LINKED ISSUE #1433: Load PK-Sim DLL upon package loading
I think once PK-SIM.R can be loaded in Linux, it should be done in `.initPackage` and not as a separate function `initPKSim`.

- Load PKSim.R.dll in initPackage
- Remove function `initPKSim`
- Remove `ospsuiteEnv$isPKSimLoaded`
- Remove everything in `utilities-pksim`

PavelBal:
Fixes #1433


msevestre (commented):
so cool

Felixmil:
Doesn't it slow down package loading ?

PavelBal:
> Doesn't it slow down package loading ?

It does indeed, but we just assume it should always be available?

msevestre:
Well the initOnce can only be called when one really use PKSim stuff no?

R/init-package.R
```diff
@@ -35,4 +35,9 @@
   rSharp::callStatic("OSPSuite.R.Api", "InitializeOnce", apiConfig)
 
   .initializeDimensionAndUnitLists()
+
+  # Load PK-Sim library
+  pksimR <- system.file("lib", "PKSim.R.dll", package = ospsuiteEnv$packageName)
+  rSharp::loadAssembly(pksimR)
+  rSharp::callStatic("PKSim.R.Api", "InitializeOnce")
```
msevestre, line 42: I am wondering if this should be done like it was before. Whast's the gain of ALWAYS loading PKSim?

PavelBal: Okay I can revert.

Felixmil:
I'm in favor of loading PKSIM only when/if used.

PavelBal:
Yepp, closing.

PR #1469, opened 2024-07-25
update README.Rmd

Yuri05:


PavelBal (dismissed):

Felixmil (changes requested):
Needs to be knitted into `.md` file

I would rename "Main-Workflow" with "build", it is more generic

Felixmil:
Merge #1468 first so that we can run all actions here

Felixmil (approved):

PR #1497, opened 2024-11-22
Implement pkAnalysisTODataFrame() without saving to csv

rengelke:
Implemented `pkAnalysisToDataFrame` without saving to csv-file.
Given the exposed .NET methods, data extraction required cell-by-cell processing, resulting in many .NET calls. This repeated crossing of the R/.NET boundary seems to negatively impact performance.

@PavelBal @rwmcintosh : Should we keep this implementation or add a dedicated C# method for bulk data retrieval?

``` r
devtools::load_all(".")

sim <- loadTestSimulation("S1")
clearOutputs(sim)
outputs <- "Organism|VenousBlood|*|*"
addOutputs(outputs, sim)
results <- runSimulations(sim)[[1]]
pkAnalyses <- calculatePKAnalyses(results)


library(microbenchmark)
benchmark_results <- microbenchmark(
  res1 = pkAnalysesToDataFrame(pkAnalyses),  # existing method
  res2 = pkAnalysesToDataFrame_(pkAnalyses), # w/o saving to csv
  times = 50L
)

benchmark_results
#> Unit: milliseconds
#>  expr     min       lq      mean   median       uq      max neval cld
#>  res1 28.0700  30.5758  38.41138  31.6742  35.0210 134.4465    50  a 
#>  res2 84.8444 106.6241 246.30354 135.9451 322.5596 747.7574    50   b
```


PavelBal:
add a dedicated C# method for bulk data retrieval?

PavelBal:
@rwmcintosh We agreed on implementing a method that converts the PK analysis data dable to an R data frame in core, right?

rwmcintosh:
I don't remember how we left it. I will create an issue for it in Core.

rwmcintosh:
> I don't remember how we left it. I will create an issue for it in Core.

Now I remember. We can't do it in Core because Core has no idea about R data structures. We could do an implementation in rSharp that takes 2d DataTables and converts them into dataframes with strings for all values.

The main issue with doing any of this is that OSPSuite.Core does not know anything about R specifically and Sharp does not know anything about the domain itself. The only place where they meet is in OSPSuite-R, which does not have its own .NET layer.

PR #1499, opened 2025-01-08
Make 64bits R version check compatible with arm architecture

Felixmil:
# Description
- Adds native library loading logic on package load (rename files depending on host's os)
- Adds github action to build SQlite natives for macos

# To Do
- [ ] Fix running simulations from snapshots
- [x] Integrate SQlite building with libs injection workflow.


rwmcintosh:
The problem with implementing an ARM version is that System.Data.SQLite is delivered with only native binaries for x64. We cannot simply compile our own it seems because we get this `EntryPointNotFound` exception.

The advice from StackOverflow is to compile our own System.Data.SQLite, but I think this is going too far.

We might be able to offer compatibility for Mac if you use the x64 version of R instead of ARM64. That seems to be workable.

Felixmil:
@rwmcintosh  What about https://www.nuget.org/packages/Microsoft.Data.Sqlite/ ?

rwmcintosh:
> @rwmcintosh What about https://www.nuget.org/packages/Microsoft.Data.Sqlite/ ?

Yeah, that would probably work. Replacing the package in PK-Sim will require an investigation.

Felixmil:
Closed in favor of cleaner PR #1621

PR #1501, opened 2025-01-29
Implement `getMolWeightFor()` to work with `Quantity`

### LINKED ISSUE #1500: Implement `getMolWeightFor()` for quantities
```
#' @param quantity A `Quantity` object.
#' @return Numeric value of the molecular weight in `kg/µmol`.
getMolWeightFor <- function(quantity) {...}
```

rengelke:
`getMolWeightFor()` retrieves the molecular weight of a quantity's (parent) molecule, returning it in the specified unit. 
Returns NA if no molecule is found, or throws an error if stopIfNotFound = TRUE.

Utilizes helper `.getParentContainerByType` which allows to get the parent container of a quantity by type.

Closes #1500


PavelBal:
@Felixmil @rwmcintosh I forgot how it works now with the core files - they are updated automatically? Should they be part of the PR?

Felixmil:
> @Felixmil @rwmcintosh I forgot how it works now with the core files - they are updated automatically? Should they be part of the PR?

Yes, they are retrieved from latest pksim build and commit to the PR's branch by a github action:
image

Felixmil:
> Just learned that there will be a problem pulling the core files when the package is loaded

Yes, because they are in use. Restarting the R session is necessary. We should also update the `update_core_files.R` script to restart the session before running it.

PavelBal (approved):

rengelke:
> > Just learned that there will be a problem pulling the core files when the package is loaded
> 
> Yes, because they are in use. Restarting the R session is necessary. We should also update the `update_core_files.R` script to restart the session before running it.

Sure, makes sense to restart, but one might not consider it during the first time. 

> > Just learned that there will be a problem pulling the core files when the package is loaded
> 
> Yes, because they are in use. Restarting the R session is necessary. We should also update the `update_core_files.R` script to restart the session before running it.

Yes, makes sense to restart before pulling the core files while working on the PR. But Do I miss something - how is changing `update_core_files.R` going to help as this is used in the context of the GitHub action.

PR #1534, opened 2025-03-21
Do not export `SimulationBatchRunValues` and `SimulationBatchOptions`

### LINKED ISSUE #1526: Make `SimulationBatchRunValues` internal class?
The class `SimulationBatchRunValues` doesn't seem to be of any use to the end user, I even do not know how to create it.

I think it should not be exported.

### LINKED ISSUE #1527: Make `SimulationBatchOptions` internal
As far as I can see it, it cannot be used by the user directly anywhere. Would make it internal.

PavelBal:
Closes #1527
Closes #1526


PavelBal:
@Yuri05 @msevestre  I would like to have your "okay" on this before merging.

msevestre:
How do you create them outside of the package if you don’t export them ? What’s the reasoning?

msevestre:
Oh they are not used outside of the package it seems ? Then fine with me

PavelBal:
`SimulationBatchRunValues` is created in the `addRunValues` function of `SimulationBatch`:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/af121696c6cb7fd5194b3ae80bcea087eebc9c9b/R/simulation-batch.R#L83

I could not find any other use of it outside of this function.

`SimulationBatchOptions` is used:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/af121696c6cb7fd5194b3ae80bcea087eebc9c9b/R/simulation-batch.R#L95
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/af121696c6cb7fd5194b3ae80bcea087eebc9c9b/R/simulation-batch.R#L108


Also I could not find any other use of it, and it can neither be created by the user nor directly passed anywhere. In this case, I prefer not to export these classes so the user is not confused.

msevestre (approved):

PR #1544, opened 2025-04-11
Add legendLabels Field to DefaultPlotConfiguration

### LINKED ISSUE #1543: Add `legendLabels` Field to `DefaultPlotConfiguration`
To support legend customization in sensitivity analysis related plots in `esqlabsR`

rengelke:
Closes #1543


PavelBal:
Yes but if this is exposed in `ospsuite-r`, it must also be documented somehow.

rengelke:
> Yes but if this is exposed in `ospsuite-r`, it must also be documented somehow.

Basically it's just an additional field in the R6 class. If we want to demonstrate it's usage, it would mean to add functionality in tlf-based functions, right?

Felixmil:
What element do you want to target with `legendLabel` ? Doesn't `legendTitle` already covers what you need ?

rengelke:
> Which element do you want to target with `legendLabel` ? Doesn't `legendTitle` already covers what you need ?

Not exactly, the legendTitle targets the title of the legend while I would like to rename the individual entries in the legend itself.

Felixmil:
If we want to allow that for user, we would be better off rewriting ggplot2... To me it is a big no no. This is not how ggplot2 works.
If you want legend entries to have a different name, user should just rename the **input** data.

rengelke:
> If we want to allow that for user, we would be better off rewriting ggplot2... To me it is a big no no. This is not how ggplot2 works. If user wants legend entries to have a different name, they should just rename them in input data.

That's true in general, but not always feasible. When input data is internally generated using parameter paths like in sensitivity calculation. There’s no clean entry point to rename the parameters before plotting.

Felixmil:
Not even as an argument in the plot generating function in ESQlabsR ? Then manipulate data or plot object before generating/returning it.

Felixmil (changes requested):
Should be done in ESQlabsR instead of here.

rengelke:
> Not even as an argument in the plot generating function in ESQlabsR ? Then manipulate data or plot object before generating/returning it.

Sure, this is possible. Then we have the case that configurations are spread between DefaultPlotConfiguration and function signature

Felixmil:
Yes but since it is a specific alteration, I think it's ok. You could create a child class.

rengelke:
> create a child cla

Child class is probably the best approach

rengelke:
Adding `legendLabels` to `DefaultPlotConfiguration` isn’t appropriate without corresponding functionality within {ospsuite} itself.

PR #1546, opened 2025-04-14
Add option to show separate legend entries per dataset

### LINKED ISSUE #1548: Plot framework - allow separate legend entries for grouped entries

rengelke:
Closes #1548


rengelke:
The current code in `.plotTimeProfile` removes the guides related to names:
```
# linetype: simulated data set names
# shape: observed data set names
profilePlot <- profilePlot + ggplot2::guides(linetype = "none", shape = "none")
```
This step could be controlled as suggested. 

However when both observed and simulated datasets >1, we expect to see legends for:
- color (group)
- shape (observed dataset names)
- linetype (simulated dataset names)
which is not the case.

The way `shape` and `linetype` legends are shown (or suppressed) is not consistent:
observed > 1
plot1
simulated > 1
plot2
observed > 1 and simulated > 1
plot3

tests/testthat/_snaps/plot-individual-time-profile/multiple-obs-separate-legend.svg
Yuri05, line 1: Somehow this legend does not make sense anymore?
I would expect that the legend entry "Aciclovir observed" disappears and 2 other legend entries are displayed with the correct color (and not only with the shape match).

rengelke: Agree, the current legend isn't ideal. Hiding the group-level color legend when showing all dataset names would require additional changes to how `tlf::plotTimeProfile()` handles legends, or adding logic to modify the generated plot. I tested resetting the guides using `ggplot2::guides(...)`, but the aesthetics were getting mixed up: shapes and lines didn’t render as expected.

Given that this is a temporary solution, as @PavelBal indicated, I kept it minimal. It is mentioned as experimental in the documentation.

NEWS.md
```diff
@@ -8,6 +8,9 @@ They should not be used directly.
 
 ## Minor improvements and bug fixes
 
+- - Added `showLegendPerDataset` parameter to `plotIndividualTimeProfile()` and
```
PavelBal, line 11:
```suggestion
- Added `showLegendPerDataset` parameter to `plotIndividualTimeProfile()` and 
```

PavelBal (approved):

PavelBal (approved):

PR #1552, opened 2025-04-29
Set simulation name

### LINKED ISSUE #1245: Allow changing of the `name` property of `Simulation` object
When creating multiple simulations from the same pkml file and saving them, it would be useful to be able to change the `Name` property, so when loading the simulation in MoBi it will have a name defined in R.

Is this property read-only in .NET and do you see any problems with changing it from R @msevestre  @Yuri05

PavelBal:
Fixes #1245

- Added a new property `illegalCharacters` that can be retrieved with `getOSPSuiteSetting("illegalCharacters")` to get the list characters that are not allowed to be used in simulation names.
- The `name` property of a `Simulation` can now be changed (#1245)


Yuri05 (approved):

PavelBal:
Have to figure out why the build is failing... we had something similar already.

R/zzz.R
```diff
@@ -14,5 +14,6 @@
   }
 
   .initPackage()
+  ospsuiteEnv$illegalCharacters <- .getIllegalCharacters()
```
PavelBal, line 17: @Yuri05  could you re-approve. I had to move the call to .NET to here because otherwise the checks didn't pass (I still don't understand how R handles some things...)

R/simulation.R
```diff
@@ -40,6 +40,18 @@ Simulation <- R6::R6Class(
...
     sourceFile = function(value) {
       private$.readOnlyProperty("sourceFile", value, private$.sourceFile)
+    },
+    #' @field name Name of the simulation
+    name = function(value) {
+      if (missing(value)) {
+        return(self$get("Name"))
+      } else {
+        # Check for illegal characters
+        if (any(stringr::str_detect(value, paste0("[", getOSPSuiteSetting("illegalCharacters"), "]")))) {
```
Yuri05, line 50: Why also adding `[` and `]` ?

R/zzz.R
Yuri05: Approved.
R =

R/simulation.R
PavelBal: This way it is considered as a regular expression with the symbols inside `[]`. From documentation:

`str_detect(fruit, "[aeiou]")`

will look for the letters `aeiou` inside the string `fruit`.

Yuri05 (approved):

Yuri05:
> Why are the paths for population simulation constructed differently?

@PavelBal If you execute this code for a population simulation **without changing the simulation name** - do you still get the simulation name in the output paths?

PavelBal:
@Yuri05 Interesting, no I don't.  So changing the name somehow prepends the **old** simulation name to the population outputs, but not individual?

Yuri05:
> @Yuri05 Interesting, no I don't. So changing the name somehow prepends the **old** simulation name to the population outputs, but not individual?

Hmm. No real idea, why. Needs to be debugged.

Yuri05 (changes requested):
Additionally to the illegal chars in the new name it must be checked, that:
* The new name differs from the old name
* Is not one of the forbidden names for the given simulation. This includes 
  * the names of all direct child containers of the top container of the simulation
  * `MoleculeProperties`

E.g. here all forbidden names for this simulation:
grafik
grafik

Also the new named should be trimmed (remove leading and trailing blanks)

R/simulation.R
```diff
@@ -40,6 +40,33 @@ Simulation <- R6::R6Class(
...
       private$.readOnlyProperty("sourceFile", value, private$.sourceFile)
+    },
+    #' @field name Name of the simulation
+    name = function(value) {
+      if (missing(value)) {
+        return(self$get("Name"))
+      } else {
+        validateIsString(value)
+        # Trim the name of ws
+        value <- trimws(value, which = "both")
```
PavelBal, line 51: @Yuri05 Trimming the new name

R/simulation.R
```diff
@@ -40,6 +40,33 @@ Simulation <- R6::R6Class(
...
+        # Check if the name actually changes
+        if (self$name == value) {
```
PavelBal, line 53: @Yuri05 Do nothing if the name does not change.

R/simulation.R
```diff
@@ -40,6 +40,33 @@ Simulation <- R6::R6Class(
...
+          return(invisible(self))
+        }
+
+        # Check for forbidden names
+        forbiddenNames <- .getIllegalSimulationNames(self)
```
PavelBal, line 58: Checking for forbidden names

R/utilities.R
```diff
@@ -97,3 +97,33 @@ clearMemory <- function(clearSimulationsCache = FALSE) {
...
+#' Get illegal simulation names
+#'
+#' Illegal names are the names of all direct child containers of the top container of the simulation,
+#' plus the reserved word `MoleculeProperties`.
+#'
+#' @param sim A `Simulation` object.
+#'
+#' @returns A vector of illegal names.
+#' @noRd
+.getIllegalSimulationNames <- function(sim) {
```
PavelBal, line 122: @Yuri05 This is how the forbidden names for a simulation are retrieved.

Yuri05 (approved):

Yuri05:
@PavelBal Looks good for me 👍

PR #1554, opened 2025-05-20
Mobi r

PavelBal:
Suggestion of the API to work with MoBi projects from R.

Still in progress, but I already upload and you can start reviewing.

### DIFF
--- a/R/MoBiProject.R
+++ b/R/MoBiProject.R
@@ -34,7 +34,7 @@ MoBiProject <- R6::R6Class(
       if (missing(value)) {
         model <- self$get("Individuals")
       } else {
-        private$.throwPropertyIsReadonly("Indivduals")
+        private$.throwPropertyIsReadonly("Individuals")
       }
     },
     #' @field ExpressionProfiles Names of the expression profiles that are present in the project
--- a/R/MoBiProject.R
+++ b/R/MoBiProject.R
@@ -1,0 +1,163 @@
+#' @title MoBi Project
+#' @docType class
+#' @description  A MoBi project, containing modules, building blocks, simulations, etc.
+#' @format NULL
+MoBiProject <- R6::R6Class(
+  "MoBiProject",
+  cloneable = FALSE,
+  inherit = ObjectBase,
+  active = list(
+    #' @field sourceFile Path to the file the project was loaded from (read-only)
+    sourceFile = function(value) {
+      private$.readOnlyProperty("sourceFile", value, private$.sourceFile)
+    },
+    #' @field simulationNames Names of the simulations that are present in the project
+    simulationNames = function(value) {
+      if (missing(value)) {
+        model <- self$get("SimulationNames")
+      } else {
+        private$.throwPropertyIsReadonly("simulationNames")
+      }
+    },
+    #' @field parameterIdentificationNames Names of the parameter identifications that are present in the project
+    #'
+    #' PRIO 2
+    parameterIdentificationNames = function(value) {
+      if (missing(value)) {
+        model <- self$get("ParameterIdentificationNames")
+      } else {
+        private$.throwPropertyIsReadonly("parameterIdentificationNames")
+      }
+    },
+    #' @field Individuals Names of the individuals that are present in the project
+    Individuals = function(value) {
+      if (missing(value)) {
+        model <- self$get("Individuals")
+      } else {
+        private$.throwPropertyIsReadonly("Indivduals")
+      }
+    },
+    #' @field ExpressionProfiles Names of the expression profiles that are present in the project
+    ExpressionProfiles = function(value) {
+      if (missing(value)) {
+        model <- self$get("ExpressionProfiles")
+      } else {
+        private$.throwPropertyIsReadonly("ExpressionProfiles")
+      }
+    }
+  ),
+  public = list(
+    #' @description
+    #' Initialize a new instance of the class
+    #'
+    #' Should not be directly used. Instead, use function `loadMoBiProject()`
+    #' to load a project.
+    #' @param netObject Reference to `NetObject` .NET MoBi-project object
+    #' @param sourceFile (Optional) File used to load the project
+    #' @returns A new `MoBiProject` object.
+    initialize = function(netObject, sourceFile = NULL) {
+      super$initialize(netObject)
+      private$.sourceFile <- sourceFile
+    },
+
+    #' @description
+    #' Load a simulation from the project
+    #'
+    #' @param simulationName Name of the simulation.
+    #' @returns A `Simulation` object, if the simulation with the given name is present in the
+    #' project. `NULL` if no such simulation is available.
+    getSimulation = function(simulationName) {
+    },
+
+    #' @description
+    #' Get observed data present in the project.
+    #'
+    #' @param dataSetNames Optional. List of names of observed data sets to retrieve
+    #' from project. If `NULL`, all data sets are returned. If a specified data set
+    #' is not found, the name is ignored.
+    #'
+    #' @returns A named list of `DataSet` objects. `NULL` if the project does not contain
+    #' any observed data or no specified data sets are found.
+    getObservedData = function(dataSetNames = NULL){
+    },
+
+    #' @description
+    #' Get modules present in the project.
+    #'
+    #' @returns A named list of `MoBiModule` objects. `NULL` if the project does not contain
+    #' any module.
+    getModules = function(){
+    },
+
+    #' @description
+    #' Get a specified individual from the project
+    #' @param name Name of the individual
+    #' @param stopIfNotFound If `TRUE` (default), an error is thrown if the specified
+    #' individual is not present in the project.
+    #'
+    #' @returns An object of the type `BuildingBlock`. `NULL` if the project does not contain
+    #' such an individual and `stopIfNotFound = FALSE`.
+    getIndividual = function(name, stopIfNotFound = TRUE){
+    },
+    #' @description
+    #' Get specified expression profiles from the project.
+    #' @param names List of names of the expression profiles to retrieve.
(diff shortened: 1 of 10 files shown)


vignettes/mobi-projects.Rmd
```diff
@@ -0,0 +1,266 @@
...
+## Creating simulations
+
+The user can create new simulation from the modules available in the project. In this example,
+we create a simulation for the rat species from the modules "Thyroid_QST", "Rat physiology", "Thyroid", "Pituitary", "Endogenous_TH", "TH_activeTransports", and "TH_plasma_binding". We will
+select the expression profiles "UDPGT1|Human|Healthy", "DIO1|Human|Healthy", "DIO3|Human|Healthy",
+"UDPGT2|Human|Healthy", and specify the parameter values building block "Rat" from the 
+module "Thyroid":
+
+```{r createSimulationFromProject}
+sim1 <- myProject$createSimulation(simulationName = "Thyroid_QST_rat",
```
rwmcintosh, line 122: @benjaperez1983 and I implemented an integration test this way as well and as a test it works.

With this workflow, you can create a simulation only if the project already has all the modules, and you cannot use any other modules from another source.

However, I think a better idea might be to have the createSimulation work with `Modules` and not have a built-in dependency on the project....

The workflow would be to get the modules and bb's from the project (or file system, or github) then create the simulation from the modules and bb's independently of the project.

rwmcintosh: ok, reading the comments below, I see this is planned, but I think doing it by supplying the strings is going to be a sort of 'limited' version of the code below. I don't see a need for it.

rwmcintosh: Sure, you can also hide whatever abstractions in R code as well.

R/MoBiProject.R
```diff
@@ -0,0 +1,163 @@
...
+    simulationNames = function(value) {
+      if (missing(value)) {
+        model <- self$get("SimulationNames")
+      } else {
+        private$.throwPropertyIsReadonly("simulationNames")
+      }
+    },
+    #' @field parameterIdentificationNames Names of the parameter identifications that are present in the project
+    #'
+    #' PRIO 2
```
Yuri05, line 24: delete the comment `PRIO 2`? or what does it mean?

R/MoBiProject.R
```diff
@@ -0,0 +1,163 @@
...
+    parameterIdentificationNames = function(value) {
+      if (missing(value)) {
+        model <- self$get("ParameterIdentificationNames")
+      } else {
+        private$.throwPropertyIsReadonly("parameterIdentificationNames")
+      }
+    },
+    #' @field Individuals Names of the individuals that are present in the project
+    Individuals = function(value) {
```
Yuri05, line 33: Consistency: `Individuals `and `ExpressionProfiles `start with the upper case letter, while other fields start with the lower case

R/MoBiProject.R
```diff
@@ -0,0 +1,163 @@
...
+    getIndividual = function(name, stopIfNotFound = TRUE){
+    },
+    #' @description
+    #' Get specified expression profiles from the project.
+    #' @param names List of names of the expression profiles to retrieve.
+    #' @param stopIfNotFound If `TRUE` (default), an error is thrown if any of the specified
+    #' expression profiles is not present in the project.
+    #' @returns A named list of objects of the type `BuildingBlock`. `NULL` for each
+    #' specified expression profile that is not present in the project  if `stopIfNotFound = FALSE`.
+    getExpressionProfiles = function(names, stopInfNotFound = TRUE){
```
Yuri05, line 109: typo `stopInfNotFound `

vignettes/mobi-projects.Rmd
```diff
@@ -0,0 +1,266 @@
...
+                                            modules = c(modules[["Thyroid_QST"]],
+                                                        thyroidModule,
+                                                        modules[["Pituitary"]],
+                                                        modules[["Endogenous_TH"]],
+                                                        modules[["TH_activeTransports"]],
+                                                        modules[["TH_plasma_binding"]]),
+                                          
+                                            individual = myIndividual,
+                                            expressionProfiles = expressionProfiles,
+                                            selectedParameterValues = list("Thyroid" = "Human")
```
Yuri05, line 259: indentation

R/utilities-building-block.R
```diff
@@ -0,0 +1,177 @@
...
+#' @param formula A list of `Formula` objects that will be set for the quantities.
+#' Mixing of `formula` and `quantityValues` is not allowed. If `formula` is provided,
+#' `quantityValues` should be `NULL`. The length of this list should be equal to the length of `quantityPaths`.
+#'
+#' @returns The updated `icBuildingBlock` object.
+#' TBD: no return? To be consistent (or not confusing) with the extend functions?
+#' @export
+#'
+#' @examples
+setInitialConditions <- function(icBuildingBlock,
```
Yuri05, line 29: would give the full name to the argument: `initialConditionsBuildingBlock`

R/MoBiProject.R
PavelBal: That this feature is considered a second priority :) TBD if it should/can be implemented in the first iteration.

R/utilities-building-block.R
```diff
@@ -0,0 +1,177 @@
...
+                                 quantityPaths,
+                                 quantityValues,
+                                 scaleDivisors = 1,
+                                 isPresent = TRUE,
+                                 negativeValuesAllowed = FALSE,
+                                 formula = NULL){
```
Yuri05, line 35: rename `formula `to `formulas` (plural)?

R/utilities-building-block.R
```diff
@@ -0,0 +1,177 @@
+#' Set or add initial conditions to an existing Intitial Conditions building block.
+#'
+#' This functions allows adding or modifying initial condition entries.
+#'
+#' @param icBuildingBlock A `BuildingBlock` object of type `Initial Conditions`.
+#' The entries will be added to or set in this building block.
+#' @param quantityPaths A list of full paths of the quantities (usually molecules). Should contain
+#' all path elements and the molecule name, separated by `|`.
+#' @param quantityValues A list of values for the quantities. The length of this list should be equal to the length of `quantityPaths`.
```
Yuri05, line 9: according to the comment below, `quantityValues `can be null if the `formulas `argument is not null

R/utilities-building-block.R
```diff
@@ -0,0 +1,177 @@
...
+#' with these names will be added to the `icBuildingBlock`.
+#'
+#' @returns Paths of entries added to the building block.
+#' @export
+#'
+#' @examples
+extendInitialConditions <- function(icBuildingBlock,
+                                    spatialStructureBB,
+                                    moleculesBB,
+                                    moleculeNames = NULL){
```
Yuri05, line 74: * should we also provide list of container paths (NULL per default) to extend only for selected containers?
* it is not clear from the comment what will happen if an initial condition for the combination {container, molecule} alreade exists in the IC-BB (overwritten, kept AS IS or error)

R/utilities-building-block.R
```diff
@@ -0,0 +1,177 @@
...
+#' @param quantityValues A list of values for the quantities. The length of this list should be equal to the length of `quantityPaths`.
+#'
+#' @export
+#'
+#' @examples
+setParameterValues <- function(pvBuildingBlock,
+                                 quantityPaths,
+                               quantityValues,
+                               dimensions = ospDimensions$Dimensionless,
+                                 formula = NULL){
```
Yuri05, line 103: * indentation
* same comments/suggestions as for `setInitialConditions`

R/utilities-building-block.R
```diff
@@ -0,0 +1,177 @@
...
+#' with these names will be added to the `pvBuildingBlock`.
+#'
+#' @returns Path of entries added to the building block.
+#' @export
+#'
+#' @examples
+addLocalMoleculeParameters <- function(pvBuildingBlock,
+                                    spatialStructureBB,
+                                    moleculesBB,
+                                    moleculeNames = NULL){
```
Yuri05, line 142: same question/suggestion as for the initial conditions (optional list of container start paths)

R/utilities-individual.R
```diff
@@ -1,3 +1,38 @@
...
+                                          weightUnit = "kg",
+                                          height = NULL,
+                                          heightUnit = "cm",
+                                          age = NULL,
+                                          ageUnit = "year(s)",
+                                          gestationalAge = 40,
+                                          gestationalAgeUnit = "week(s)",
+                                          moleculeOntogenies = NULL,
+                                          seed = NULL){
+}
```
Yuri05, line 34: missing arguments:
* disease state
* list of disease state parameters

R/utilities-mobi-project.R
```diff
@@ -0,0 +1,20 @@
...
+#' @returns An object of the `MoBiProject` type.
+#' @export
+#'
+#' @examples
+#' projectPath <- system.file("extdata", "simple.mbp3", package = "ospsuite")
+#'
+#' myProject <- loadMoBiProject(projectPath)
+loadMoBiProject <- function(filePath){
+  # .NET task that handles loading of a MoBi project
+  netTask <- .getNetTask("SomeMoBiTask")
```
Yuri05, line 14: I assume "SomeMoBiTask" is a placeholder?

R/utilities-building-block.R
PavelBal: > it is not clear from the comment what will happen if an initial condition for the combination {container, molecule} already exists in the IC-BB (overwritten, kept AS IS or error)

Same behavior as in MoBi -  I assume kept AS IS (== ignored), @rwmcintosh can you confirm?

PavelBal:
Closing in favor of #1563

PR #1563, opened 2025-07-18
Mobi r configuration

PavelBal:


R/MoBiProject.R
```diff
@@ -0,0 +1,170 @@
...
+    #' where the name should be the name of the module and the value the name of the IC BB.
+    #' By setting the value to `NULL`, no IC BB from the specified module will be applied.
+    #' @param selectedParameterValues By default, the first Parameter Values
+    #' (PV) building block (BB) of each module will be selected. If a module has multiple
+    #' PV BBs, it is possible to specify which PV BB to apply by providing a named list,
+    #' where the name should be the name of the module and the value the name of the PV BB.
+    #' By setting the value to `NULL`, no PV BB from the specified module will be applied.
+    #'
+    #' @returns A `SimulationConfiguration` object.
+     createSimulationConfiguration = function(modulesNames, individualName = NULL, expressionProfilesNames = NULL, selectedInitialConditions = NULL, selectedParameterValues = NULL){
```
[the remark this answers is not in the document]
rwmcintosh: So, is the plan to have the moduleNames order and selected lists order related? I would say the better way to handle this is that we mirror the .NET construction. You need a set of `ModuleConfigurations` to create a `SimulationConfiguration` The `ModuleConfiguration` indicates which, if any, PV and IC are selected.

rwmcintosh, line 130: So, is the plan to have the moduleNames order and selected lists order related? I would say the better way to handle this is that we mirror the .NET construction. You need a set of `ModuleConfigurations` to create a `SimulationConfiguration` The `ModuleConfiguration` indicates which, if any, PV and IC are selected.

R/MoBiProject.R
```diff
@@ -0,0 +1,170 @@
...
+       modules <- self$getModules(moduleNames)
+       individual <- self$getIndividual(individualName, stopIfNotFound = FALSE)
+       expressionProfiles <- self$getExpressionProfiles(expressionProfilesNames, stopIfNotFound = FALSE)
+
+       configuration <- createSimulationConfiguration(modules = modules,
```
rwmcintosh, line 135: Here you would create a set of `ModuleConfigurations` that group the module, and selected building blocks, then create the `simulationConfiguration`.

R/MoBiProject.R
```diff
@@ -0,0 +1,170 @@
...
+      private$.sourceFile <- sourceFile
+    },
+
+    #' @description
+    #' Load a simulation from the project
+    #'
+    #' @param simulationName Name of the simulation.
+    #' @returns A `Simulation` object, if the simulation with the given name is present in the
+    #' project. `NULL` if no such simulation is available.
+    getSimulation = function(simulationName) {
```
PavelBal, line 67: @rwmcintosh @benjaperez1983 this is not supported yet?

R/MoBiProject.R
```diff
@@ -0,0 +1,170 @@
...
+
+    #' @description
+    #' Create a simulation configuration.
+    #'
+    #'
+    #' @param modulesNames A list of the modules from which to create in simulation.
+    #' All defined modules must be present in the project. The order of module names defines the order in which the modules will be combined to a simulation!
+    #' @param individualName Optional, name of the individual.
+    #' @param expressionProfilesNames Optional, list of expression profiles to apply to the simulation.
+    #' @param selectedInitialConditions By default, the first Initial Conditions
```
PavelBal, line 118: @rwmcintosh @benjaperez1983 @msevestre 

Is it somehow possible to implement it this way? So that there is the default behavior, that the first available BB is selected, by also being able to select NO IC BB?

PavelBal: And what is the current behavior?

rwmcintosh: ah, by default we will pre-select an IC or PV if there is only one. That's Presentation logic, not Core logic.

R/MoBiProject.R
rwmcintosh: no, that looks like we have `projectTask.AllSimulations(MoBiProject moBiProject)` but not a single by name.
The API here will need to supply the project from which we retrieve it.
```
getSimulation = function(project, simulationName)
```

inst/lib/OSPSuite.TeXReporting.dll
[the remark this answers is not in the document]
benjaperez1983: I`ve created [this task](https://github.com/Open-Systems-Pharmacology/MoBi/issues/2172) for this.
Feel free to add your comments on it. @PavelBal

PR #1575, opened 2025-08-07
Load mobi r

PavelBal:


R/get-net-task.R
```diff
@@ -20,9 +32,14 @@
 #' @return returns an instance of of the specified `.NET` task.
 #'
 #' @keywords internal
-.getNetTaskFromCache <- function(taskName) {
+.getNetTaskFromCache <- function(taskName, isMoBiR = FALSE) {
```
PavelBal, line 35: @rwmcintosh See, here, we are only calling to .NET if the task has not been created, yet. But `.getNetTask()` always calls to `OSPSuite.R.Api`, so I had to create a separate function to call to MoBi.R. Not nice.

R/init-package.R
```diff
@@ -32,7 +32,11 @@
   apiConfig$dimensionFilePath <- filePathFor("OSPSuite.Dimensions.xml")
   apiConfig$pkParametersFilePath <- filePathFor("OSPSuite.PKParameters.xml")
 
-  rSharp::callStatic("OSPSuite.R.Api", "InitializeOnce", apiConfig)
```
PavelBal, line 35: According to @rwmcintosh we don't have to call `InitializeOnce`  of `OSPSuite.R.Api` because it is done in `rSharp::callStatic("MoBi.R.Api", "InitializeOnce", apiConfig)`

R/get-net-task.R
rwmcintosh: Well, yeah, but this needs to be somehow addressed. It can be hidden by a .NET API in this project, or in R. Someone has to know where to find this method.

R/get-net-task.R
```diff
@@ -10,6 +10,18 @@
...
+#' @title .getMoBiTask
+#' @description Get an instance of the specified `.NET` Task in MoBi.R
+#'
+#' @param taskName The name of the task to retrieve (**without** `Get` prefix).
+#'
+#' @return returns a new instance of of the specified `.NET` task.
+#'
+#' @keywords internal
+.getMoBiTask <- function(taskName) {
+  rSharp::callStatic("MoBi.R.Api", paste0("Get", taskName))
```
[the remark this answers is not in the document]
PavelBal: I think everything should go through ospsuite.r API, no?

R/utilities-mobi-project.R
```diff
@@ -6,14 +6,20 @@
 #' @export
 #'
 #' @examples
+#' \dontrun{
 #' projectPath <- system.file("extdata", "simple.mbp3", package = "ospsuite")
 #'
 #' myProject <- loadMoBiProject(projectPath)
+#' }
 loadMoBiProject <- function(filePath) {
+  if (!file.exists(filePath)) {
```
PavelBal, line 15: Loading a MoBi project

PavelBal:
@Yuri05 @msevestre 
First step. Loading the MoBi.R DLLs and loading a MoBi project.

R/get-net-task.R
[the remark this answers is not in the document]
rwmcintosh: I think it should be abstracted by R. OSPsuite-R can have the knowledge of what services come from what namespaces.

R/get-net-task.R
msevestre: It feels to me that this should be TWO methods not one
.getMoBiTaskFromCache 
.getCoreTaskFromCache

this isMoBiR as aparameter is not ergonomic at all

PR #1607, opened 2025-09-17
Main update

PavelBal:
- Format all files with Air (new formatter in Positron
- Rebuilt documentation
- Update all snapshots with ggplot4 (no visible changes)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: ospsuite
 Title: R package to manipulate OSPSuite Models
-Version: 12.3.1.9001
+Version: 12.3.1.9002
 Authors@R: 
     c(person("Open-Systems-Pharmacology Community", role = c("cph", "fnd")),
       person("Michael", "Sevestre", role = c("aut", "cre"), email = "michael@design2code.ca"),
--- a/tests/testthat/test-utilities-population.R
+++ b/tests/testthat/test-utilities-population.R
@@ -30,7 +30,7 @@ test_that("It returns the random seed used if not specified", {
     numberOfIndividuals = 10,
   )
   human_values <- createPopulation(populationCharacteristics = human)
-  expect_gt(human_values$seed, 0)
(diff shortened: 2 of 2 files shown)


Felixmil:
GHA builds show exactly what happens when we cap versions:
TLF asks for gglot < 4 but ospsuite (which also depends on ggplot2 directly) will install latest. Result: they are incompatible.

Yuri05:
> GHA builds show exactly what happens when we cap versions: TLF asks for gglot < 4 but ospsuite (which also depends on ggplot2 directly) will install latest. Result: they are incompatible.

Which I don't fully understand: 
1. the order of installation should usually be: `tlf` before `ospsuite`, right?
2. `tlf` requires `ggplot2<4` and thus install the latest 3.x version.
3. `ospsuite` has no constraints on the version of ggplot2 - thus should just consume the already installed version of ggplot2 instead of installing the newest one?

Felixmil:
A

> > GHA builds show exactly what happens when we cap versions: TLF asks for gglot < 4 but ospsuite (which also depends on ggplot2 directly) will install latest. Result: they are incompatible.
> 
> Which I don't fully understand:
> 
> 1. the order of installation should usually be: `tlf` before `ospsuite`, right?
> 2. `tlf` requires `ggplot2<4` and thus installs the latest 3.x version.
> 3. `ospsuite` has no constraints on the version of ggplot2 - thus should just consume the **already installed version of ggplot2 instead of installing the newest one**?

Capped version does not make R install the latest version compliant with the restriction (3.5.2). It always installs the latest version (so here 4.0.0) then compare installed version to required one on load.
Here, pak throws an early error because it know in advance that the requirement for tlf will not be met.

Yuri05:
> They don’t auto-install the right version, but they protect against using versions we know won’t work. The main point is to provide users clear feedback about compatibilities instead of cryptic errors or wrong results.

OK, in this context definition of the _max version_ does not make any sense indeed. I always assumed that the version restrictions are applied not only when _starting_ the package, but also when _installing_ the package.

PR #1610, opened 2025-09-17
- Checking for seed type "integer" instead of >0

### LINKED ISSUE #1609: Test for population seed may fail because of negative seed
The test "It returns the random seed used if not specified" checks if the seed value is greater 0:

```
test_that("It returns the random seed used if not specified", {
  human <- createPopulationCharacteristics(
    species = Species$Human,
    population = HumanPopulation$BlackAmerican_NHANES_1997,
    numberOfIndividuals = 10,
  )
  human_values <- createPopulation(populationCharacteristics = human)
  expect_gt(human_values$seed, 0)
})
```

However, .NET can generate negative seeds sometimes. In this case, the test would fail.

PavelBal:
Fixes #1609


msevestre (approved):

msevestre:
How can the seed be negative ? Isn’t it a time since sthg ?

Felixmil:
If time since last system start doesn't fit into an Int32.

Felixmil:
> that's never the case lol. the type is Int32
[Int32](https://learn.microsoft.com/en-us/dotnet/api/system.int32?view=net-9.0)
A 32-bit signed integer containing the amount of time in milliseconds that has passed since the last time the computer was started.

2,147,483,647 ms ≈ 24.86 days

So they use negative to extend time coverage to 49.8 days and then it cycles.

Felixmil:
that's never the case lol. the type is Int32

[Int32](https://learn.microsoft.com/en-us/dotnet/api/system.int32?view=net-9.0)
A 32-bit signed integer containing the amount of time in milliseconds that has passed since the last time the computer was started.

Maximum number in Int32 is 2,147,483,647

2,147,483,647 ms ≈ 24.86 days

So they use negative to double time coverage to 49.8 days and then it cycles.

PR #1633, opened 2025-10-20
Fix SQLite stack overflow on macOS ARM64 + Read-only installation support

Felixmil:
# Fix SQLite stack overflow on macOS ARM64 + Read-only installation support

## Problem

1. **macOS ARM64 crash**: `runSimulationsFromSnapshot()` crashes with stack overflow due to complex SQL VIEWs exceeding .NET P/Invoke stack limits
2. **Read-only installations**: Package fails when installed in read-only locations (system libraries, Docker, CI)

### 1. SQLite Stack Overflow Fix
- **Early fix**: Convert problematic VIEWs to TABLEs during package initialization (before .NET code runs)
- **RSQLite-based**: Use `RSQLite::dbReadTable()` to avoid .NET P/Invoke stack issues
- **Smart caching**: Marker file (`.ospsuite_macos_fixed`) prevents re-checking
- **Auto-refresh**: Re-apply when database is updated

### 2. Read-only Installation Support
- **Fallback mechanism**: Use `tools::R_user_dir("ospsuite", which = "cache")` when `inst/lib` is read-only
- **Automatic copying**: Copy all files to writable cache directory
- **Cache management**: Marker file tracks freshness, auto-refresh on source updates
- **Transparent**: Works automatically without user intervention

## Key Changes

- **`R/utilities-database.R`**: Database fix logic with marker file management
- **`R/init-package.R`**: Read-only fallback + early database fix
- **`R/utilities-snapshots.R`**: Remove duplicate fix call
- **`DESCRIPTION`**: Add `RSQLite` and `withr` to `Suggests`
- **`README.Rmd`**: macOS development setup instructions
- **Tests**: 46 new readonly-fallback tests + CI-compatible helpers

### DIFF
--- a/tests/testthat/helper-for-tests.R
+++ b/tests/testthat/helper-for-tests.R
@@ -32,46 +32,18 @@ executeWithTestFile <- function(actionWithFile) {
 
 # Helper function to copy package to temp location
 .copyPackageToTemp <- function() {
-  # Find package root - works in multiple contexts (interactive, testthat, CI)
-  # Start from test path and search upward for DESCRIPTION
-  testDir <- testthat::test_path()
-
-  # Try different strategies to find the package root
+  # Find the package root directory (where DESCRIPTION file is located)
   pkgRoot <- NULL
 
-  # Strategy 1: Go up two levels from tests/testthat (most common case)
-  candidate <- normalizePath(file.path(testDir, "../.."))
-  if (file.exists(file.path(candidate, "DESCRIPTION"))) {
-    pkgRoot <- candidate
+  # Strategy 1: Check current working directory first
+  if (file.exists("DESCRIPTION")) {
+    pkgRoot <- getwd()
   }
 
-  # Strategy 2: In R CMD check, the package is installed in .Rcheck/
-  # Look for the installed package directory
-  if (is.null(pkgRoot) && grepl("\\.Rcheck", testDir)) {
-    # Extract package name from the path and find installed location
-    # In .Rcheck, tests run from: pkg.Rcheck/tests/testthat/
-    # The installed package is in: pkg.Rcheck/00_pkg_src/pkg/
-    rcheckDir <- sub("(/tests/testthat.*|/tests.*)", "", testDir)
-
-    # Try 00_pkg_src subdirectory (used during check)
-    srcDir <- file.path(rcheckDir, "00_pkg_src")
-    if (dir.exists(srcDir)) {
-      # Find the package directory inside 00_pkg_src
-      pkgDirs <- list.dirs(srcDir, recursive = FALSE, full.names = TRUE)
-      for (dir in pkgDirs) {
-        if (file.exists(file.path(dir, "DESCRIPTION"))) {
-          pkgRoot <- dir
-          break
-        }
-      }
-    }
-  }
-
-  # Strategy 3: Search upward from current directory
+  # Strategy 2: Search upward from current directory
   if (is.null(pkgRoot)) {
-    searchDir <- testDir
+    searchDir <- getwd()
     for (i in 1:10) {
-      # Limit search depth
       if (file.exists(file.path(searchDir, "DESCRIPTION"))) {
         pkgRoot <- searchDir
         break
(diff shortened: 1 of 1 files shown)


R/init-package.R
```diff
@@ -91,6 +91,9 @@
     }
   }
 
+  # Fix macOS SQLite database if needed (first load only)
+  .fixMacOSDatabaseIfNeeded(filePathFor("PKSimDB.sqlite"))
+
```
Yuri05: I don't quite understand the workflow.
The conversion should take place only once after the package is installed.
But this will be called every time when the library is loaded in R session... or not?

R/utilities-database.R
```diff
@@ -0,0 +1,163 @@
...
+  
+  # Check if we're on ARM64 (Apple Silicon) architecture
+  if (Sys.info()[["machine"]] != "arm64") {
+    return(FALSE)
+  }
+
+  # Check for marker file to see if fix was already applied
+  markerFile <- paste0(dbPath, ".macos-fixed")
+  backupPath <- paste0(dbPath, ".original")
+
```
Yuri05: Please take into account that the package installation path can be **read-only** and database patching will not be possible

Yuri05: I would extract this patching Code into a separate **patchOnMacWhatsoever.R** which can be called both from GHA and locally when developing on MAC

rwmcintosh: Yes, Windows/Linux would have the original database. Mac needs these views converted to tables.

Felixmil: Actually I realise now that there is already many renaming operations on lib files depending on OS which would already break in non writeable installation directory (centrally managed package installation with user restrictions)

.gitignore
```diff
@@ -59,3 +59,23 @@ inst/lib/libOSPSuite.FuncParserNative.dylib
...
+#   git update-index --skip-worktree inst/lib/PKSimDB.sqlite
+#
+# The fix is automatically re-applied when the database file's modification date
+# is newer than the marker file (indicating a database update from CI).
+#
+# To temporarily check database status:
+#   git update-index --no-skip-worktree inst/lib/PKSimDB.sqlite
+#   git status inst/lib/PKSimDB.sqlite
+#   git update-index --skip-worktree inst/lib/PKSimDB.sqlite
+
```
Yuri05: I think this part belongs into README or into the dev-docu.

Yuri05:
> > I have so many concerns that I don't know where to start. I would like to have the response from ALL reviewers before merging. So I converted the PR to draft so far.
> 
> @Yuri05—Felix and I arrived at this approach after he found that System.Data.SQLite causes a stack overflow exception on Mac ARM.
> 
> Creating this table during load is the **_least objectionable_** way to add Mac ARM support to OSPSuite-R because it only affects Mac ARM users who would otherwise not be able to use it at all.
> 
> Maybe there's something less intrusive that we didn't consider, or we should not include ARM support. I think it's a discussion worth having here.

I am more concerned about this copying of the package libs to a temporary location, which affects **ALL** OS

Yuri05:
> 1. Why does it crash on Mac, but not on Windows or Linux?
> 
>     2. Is't it a hint that the DB is set up not in the optimal way? Should this be investigated? @Yuri05 what do you think?
> 
> 
> Agree that not too many resources should be occupied by the Mac problem, but maybe this is a general problem that has been uncovered?

While there is always space for improvements, I don't think that we have a database problem here. 
As far as I understood Felix and Rob, the issue on MAC is that a **view** which has a **join** on at least 3 tables and a **union**  crashes on MAC with some kind of stack overflow and must be converted to a table first.

For me, it's clearly a bug in sqlite MAC implementation (why only on MAC - no idea).
Have you ever asked yourself why the sqlite library maintainers provide packages for Windows/Linux - but not for MAC? Maybe because something on MAC is different and was not easy to solve.

Like look at the example below (one of the failing views on MAC: `ContainerParameters_Species`): it's the simplest, very basic SQL (not even sqlite-specific or so).

<details><summary><b>Example</b></summary>

```
SELECT DISTINCT 
                [cpv].[container_id], 
                [cpv].[container_type], 
                [cpv].[container_name], 
                [cpv].[parameter_name], 
                [species]
FROM   [tab_container_parameter_values] AS [cpv],
       [tab_container_parameters] AS [cp]
WHERE  [cpv].[container_id] = [cp].[container_id]
         AND [cpv].[parameter_name] = [cp].[parameter_name]
         AND [building_block_type] = "INDIVIDUAL"
UNION
SELECT DISTINCT 
                [cpr].[container_id], 
                [cpr].[container_type], 
                [cpr].[container_name], 
                [cpr].[parameter_name], 
                [species]
FROM   [tab_container_parameter_rates] AS [cpr],
       [tab_species_calculation_methods] AS [scm],
       [tab_container_parameters] AS [cp]
WHERE  [cpr].[calculation_method] = [scm].[calculation_method]
         AND [cpr].[container_id] = [cp].[container_id]
         AND [cpr].[parameter_name] = [cp].[parameter_name]
         AND [building_block_type] = "INDIVIDUAL"
```
</details> 

Maybe the case like this can be solved by splitting a union view into several sub-views, where the final view is just
```
select * from SUB_VIEW1
union
select * from SUB_VIEW2
```
But should we spend time on this and make the database (even) more complicated just to overcome the implementation bugs in sqlite-MAC-engine? I don't think so.

@Felixmil Can you try to replace the view definition of `ContainerParameters_Species` with the one below and check if you can read it on MAC then without converting into a table?

<details><summary><b>Modified view</b></summary>

```
SELECT 
       [cpv].[container_id], 
       [cpv].[container_type], 
       [cpv].[container_name], 
       [cpv].[parameter_name], 
       [cpv].[species]
FROM   [tab_container_parameter_values] AS [cpv]
       JOIN [tab_container_parameters] AS [cp] ON [cpv].[container_id] = [cp].[container_id]
            AND [cpv].[parameter_name] = [cp].[parameter_name]
WHERE  [cp].[building_block_type] = 'INDIVIDUAL'
UNION
SELECT 
       [cpr].[container_id], 
       [cpr].[container_type], 
       [cpr].[container_name], 
       [cpr].[parameter_name], 
       [scm].[species]
FROM   [tab_container_parameter_rates] AS [cpr]
       JOIN [tab_species_calculation_methods] AS [scm] ON [cpr].[calculation_method] = [scm].[calculation_method]
       JOIN [tab_container_parameters] AS [cp] ON [cpr].[container_id] = [cp].[container_id]
            AND [cpr].[parameter_name] = [cp].[parameter_name]
WHERE  [cp].[building_block_type] = 'INDIVIDUAL'
```
</details>

PR #1637, opened 2025-10-22
Enable MacOS support in v13

Felixmil:


Felixmil:
@rwmcintosh probably a nice opportunity to update the list of exported artifacts like you did [here](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3306).

rwmcintosh:
Updated everything but I can't run tests. They are failing, but not for reasons related to the new assemblies

R/init-package.R
```diff
@@ -13,19 +13,6 @@
     file.path(libDir, name)
   }
 
-  if (!file.exists(libPathFor("System.Data.SQLite.dll"))) {
```
Felixmil: Not necessary since the new SQlite backend is interoperable across OSs

Felixmil:
> Updated everything but I can't run tests. They are failing, but not for reasons related to the new assemblies

@rwmcintosh Tests now pass on all OSs

Felixmil:
No idea, I also have format on save and used `format air .` to format the whole project folder.

PR #1649, opened 2025-11-13
Use `renv`

Felixmil:
- Sets `renv` with latest dependency versions.
- Add a nighlty GitHub action that: update packages, Run R CMD check, if it passes, update lockfile and commits it to the repo.

Can pass only once https://github.com/Open-Systems-Pharmacology/Workflows/pull/151 is accepted

### DIFF
--- a/renv/settings.json
+++ b/renv/settings.json
@@ -10,7 +10,7 @@
   "ppm.enabled": null,
   "ppm.ignored.urls": [],
   "r.version": null,
-  "snapshot.type": "explicit",
+  "snapshot.type": "implicit",
   "use.cache": true,
   "vcs.ignore.cellar": true,
   "vcs.ignore.library": true,
--- a/renv.lock
+++ b/renv.lock
@@ -1,6 +1,6 @@
 {
   "R": {
-    "Version": "4.5.1",
+    "Version": "4.5.2",
     "Repositories": [
       {
         "Name": "CRAN",


renv/settings.json
```diff
@@ -0,0 +1,20 @@
...
+  "package.dependency.fields": [
+    "Imports",
+    "Depends",
+    "LinkingTo",
+    "Suggests"
+  ],
+  "ppm.enabled": null,
+  "ppm.ignored.urls": [],
+  "r.version": null,
+  "snapshot.type": "explicit",
```
Felixmil: Will track only packages explicitely used in R code and mentionned in `DESCRIPTION`

.github/workflows/update-renv-nightly.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Set up R
+        uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+
+      - name: Install renv
+        run: Rscript -e "install.packages(c('renv','devtools'))"
+
+      - name: Install packages from renv.lock
+        run: Rscript -e "renv::install(prompt = F)"
```
Felixmil: This will install all packages from lockfile, but does not respect the versions from lockfile, instead installs the latest versions available.

.github/workflows/update-renv-nightly.yaml
```diff
@@ -0,0 +1,76 @@
...
+        uses: r-lib/actions/setup-r@v2
+
+      - name: Install renv
+        run: Rscript -e "install.packages('renv')"
+
+      - name: Update packages listed renv.lock
+        run: Rscript -e "renv::update(prompt = F)"
+
+      - name: Update renv.lock
+        run: Rscript -e "renv::snapshot(prompt = FALSE, dev = TRUE)"
```
Felixmil: `dev = TRUE` is required to snapshot `Suggests` dependencies

.github/workflows/update-renv-nightly.yaml
```diff
@@ -0,0 +1,76 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up R
+        uses: r-lib/actions/setup-r@v2
+
+      - name: Install renv
+        run: Rscript -e "install.packages('renv')"
+
+      - name: Update packages listed renv.lock
+        run: Rscript -e "renv::install(prompt = F)"
```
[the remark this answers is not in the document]
Felixmil: `install` will install latest available version of packages specified in DESCRIPTION.
`update` just updates currently installed packages.

Felixmil:
example run of nightly workflow: https://github.com/Felixmil/OSPSuite-R/actions/runs/19387146490

renv/activate.R
Yuri05: @Felixmil Could you please explain how this file was generated and why it is required in the first place?
(asking this also because I'm already thinking about further repositories, such as the Reporting Engine, where we will implement the renv concept later, as soon as it is running smoothly here)

This is the only part of the PR which I cannot judge - the rest of the PR looks good for me.

PR #1656, opened 2025-11-18
Remove all explicit mentions to {ospsuite} within package

Felixmil:
To fix the issue with the nightly renv update, which caused the error:

```r
Error: Error: package 'ospsuite' is not available
```

I’ve learned a few things:

1. When using `implicit` dependencies discovery in renv, it searches the entire codebase for any mention of a dependency. Including `ospsuite` because it is mentionned many times (vignettes, readme, some test files, and even in the code directly 🤯 )
2. The `ignored.packages` setting doesn’t seem to be respected by `renv::install()` during nightly checks.
3. Our configuration prevents `renv::install()` from working even locally because it’s looking for `{ospsuite}`.

In our case, `{ospsuite}` was mentioned in many places, most of which were incorrect (using `library()` or `ospsuite::…` in the codebase). This PR addresses that issue.

However, in other cases, such as vignettes and a native testthat file, `{ospsuite}` still needs to be present in the form of `library(ospsuite)`, which triggers `renv` dependency discovery. This leads `renv` with `implicit` mode to search for `{ospsuite}` as a dependency.

To resolve this, I removed all unnecessary direct references and added a `.renvignore` file that explicitly avoids discovering dependencies in specified folders and files.

This PR is simply to share my knowledge. I believe it would be better to use `explicit` dependency discovery. According to my tests, switching to this mode solves all issues described here.
Alternatively, we could change the nightly workflow to go through the steps of `renv::restore()` followed by `renv::update()` instead of `renv::install()`. However, if we choose the latter option, the problem will persist, and `renv::install()` will remain unusable, even locally.
Or, of course, we keep the changes from this PR. But every R packages that use this `renv` setup will require a similar review, configuration and maintenance to make it work.

Working run: https://github.com/Felixmil/OSPSuite-R/actions/runs/19483707662/job/55761097250

### DIFF
--- a/renv.lock
+++ b/renv.lock
@@ -54,7 +54,8 @@
       "Description": "Provides color schemes for maps (and other graphics) designed by Cynthia Brewer as described at http://colorbrewer2.org.",
       "License": "Apache License 2.0",
       "NeedsCompilation": "no",
-      "Repository": "CRAN"
+      "Repository": "RSPM",
+      "Encoding": "UTF-8"
     },
     "Rcpp": {
       "Package": "Rcpp",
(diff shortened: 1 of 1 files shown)


Felixmil:
FYI @Yuri05

PR #1666, opened 2025-11-26
update assemblies

rwmcintosh:
CheckForNegativeValues property setting needs to be removed since this property is exposed another way.

### DIFF
--- a/R/simulation-run-options.R
+++ b/R/simulation-run-options.R
@@ -22,8 +22,12 @@ SimulationRunOptions <- R6::R6Class(
       if (!is.null(checkForNegativeValues)) {
         lifecycle::deprecate_warn(
           when = "13.0.0",
-          what = I("ospsuite::SimulationRunOptions(checkForNegativeValues)"),
-          with = I("simulation$solver$checkForNegativeValues")
+          what = "ospsuite::SimulationRunOptions(checkForNegativeValues)",
+          with = I("simulation$solver$checkForNegativeValues"),
+          details = "This argument is maintained in SimulationRunOptions for backward compatibility but has no effect.",
+          always = TRUE,
+          env = rlang::caller_env(),
+          user_env = rlang::caller_env(2)
         )
       }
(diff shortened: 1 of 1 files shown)


rwmcintosh:
@PavelBal @Felixmil You should deal with CheckForNegativeValues in this branch.

Yuri05 (approved):

rwmcintosh:
> Let's do another update for PK-Sim deps after [Open-Systems-Pharmacology/PK-Sim#3347](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3347)

Updated

Felixmil:
@Yuri05 I introduced a deprecation warning instead of making this a breaking change. If that's acceptable for you, I will merge.

R/simulation-run-options.R
```diff
@@ -35,9 +41,16 @@ SimulationRunOptions <- R6::R6Class(
...
       ospsuite.utils::ospPrintItems(list(
         "numberOfCores" = self$numberOfCores,
-        "checkForNegativeValues" = self$checkForNegativeValues,
         "showProgress" = self$showProgress
       ))
+    },
+    #' @description
+    #' Get the checkForNegativeValues property (deprecated)
+    #' @return The value of checkForNegativeValues from the solver
+    .getCheckForNegativeValues = function() {
```
PavelBal: How does this insure backwards compatibility? The field `checkForNegativeValues` is gone, so old code using it will still break. Or what am I missing?

Yuri05:
> > Fine for me if it's clear from the deprecation warning that the old property is not only deprecated but in fact has no effect on the simulation.
> 
> This is the warning:
> 
> ```r
> Warning message:
> The `checkForNegativeValues` argument of `SimulationRunOptions()` is deprecated as of ospsuite 13.0.0.
> ℹ Please use simulation$solver$checkForNegativeValues instead.
> ℹ This argument is maintained in SimulationRunOptions for backward compatibility but has no effect.
> ```
> 
> Acceptable ?

👍

PR #1685, opened 2026-01-15
R version 4 4

PavelBal:


Yuri05:
@PavelBal Do we actually need to extend the **Installation** section in **Readme** by providing instructions how to install using **renv.lock** which is now distributed with every release?

PR #1846, opened 2026-04-08
Convert released-deps check to standalone nightly workflow

### LINKED ISSUE #1664: Discrepancy between tested environment and installed environments (user)
On CI and locally, the tested environment is defined by `renv.lock`, which contains the latest development versions of OSP R packages.

However, when a user installs a package, the `Remotes` fields (containing `@*release`) are evaluated, resulting in the installation of the latest **released** versions.

Consequently, the installed dependency tree differs from the tested one: we might produce unusable releases.

I propose a few solutions:

1. Fix the `Remotes` fields with `@*release` and use the same for our `renv.lock`. This ensures our development environment aligns with the installed environment but reduces flexibility, as we may need a feature before it’s officially released. Alternatively, we could temporarily target development versions, but this might lead to similar issues.

2. Add another CI job "test against releases" that doesn’t use `renv`. This job resolves the dependency tree against released versions only and runs checks. If the `renv` workflow passes while the other doesn’t, we know we need to make a release on a dependency before releasing here.

3. Define two `renv` profiles:
  - In the development profile, use your current setup (dev remotes, branches, etc.).
  - In the release profile, snapshot a library where dependencies originate from CRAN or GitHub tags that match `@*release`.
  Then, run the CI job using both profiles.

Do you have any other ideas?

Felixmil:
closes #1664 

Instead of running on every PR via main-workflow, run as a scheduled nightly job (3 AM UTC) that pins Remotes to @*release and tests against released OSP dependencies. Adds diagnostic warnings on failure.

## Summary by CodeRabbit

* **Chores**
  * Added an automated daily testing workflow to validate the package against released dependency versions on Windows, Ubuntu, and macOS.
  * Supports manual triggering and runs checks using the current R release.
  * Treats only errors as failures and emits warnings when mismatches against released dependencies are detected.


Yuri05 (dismissed):

Yuri05 (approved):

PavelBal:
@Felixmil should the branch be deleted (and https://github.com/Open-Systems-Pharmacology/OSPSuite-R/tree/fix-issue-1664) ?

PR #1859, opened 2026-04-09
Consolidate tests

PavelBal:
## Summary by CodeRabbit

* **New Features**
  * Individual-building-block creation now requires a name parameter.
  * New shared test fixtures and helper accessor functions for prebuilt data and plot configurations.

* **Bug Fixes**
  * Improved ignore patterns to exclude additional test artifacts and corrected dev-environment ignore entry.

* **Chores**
  * Bumped package dependency versions and enabled streamlined parallel test execution.
  * Standardized test setup/teardown, reorganized snapshots, and consolidated test helpers.


tests/testthat/setup.R
```diff
@@ -1,3 +1,6 @@
 # This file is run once before all tests
 # Set watermark option globally for all tests
 options(ospsuite.plots.watermarkEnabled = TRUE)
+
+# Create data combined objects for testing and save them as RDS files. This is done to avoid the need to create these objects from scratch in each test, which can be time-consuming.
+source(testthat::test_path("../data/create_and_save_data_combined_objects.R"))
```
PavelBal: @Felixmil @rengelke @KatrinCoboeken  I think these "DataCombined" objects must be re-created every time we run the tests. This way we can be sure that we are not testing against some outdated DC structure.

PavelBal:
@Felixmil @rengelke @sorinvoicu @nautilus69 

I added 

```
source(testthat::test_path("../data/create_and_save_data_combined_objects.R"))
``` 

to testthat.R so the DataCombined files are generated before the tests are executed. Now I get this error on GitHub:

```
── Test failures ───────────────────────────────────────────────── testthat ────

> # This file is part of the standard setup for testthat.
> # It is recommended that you do not modify it.
> #
> # Where should you do additional test configuration?
> # Learn more about the roles of various files in:
> # * https://r-pkgs.org/testing-design.html#sec-tests-files-overview
> # * https://testthat.r-lib.org/articles/special-files.html
> 
> library(testthat)
> library(ospsuite)
> 
> numCores <- parallel::detectCores()
> numCores <- if (is.na(numCores)) 1L else numCores
> Sys.setenv(TESTTHAT_PARALLEL = "true")
> Sys.setenv(TESTTHAT_CPUS = as.character(numCores))
> 
> # Create data combined objects for testing and save them as RDS files. This is done to avoid the need to create these objects from scratch in each test, which can be time-consuming.
> source(testthat::test_path("../data/create_and_save_data_combined_objects.R"))
Error in `testthat::test_path()`:
! Can't find 'tests/testthat'.
Backtrace:
    ▆
 1. ├─base::source(testthat::test_path("../data/create_and_save_data_combined_objects.R"))
 2. └─testthat::test_path("../data/create_and_save_data_combined_objects.R")
 3.   └─cli::cli_abort("Can't find {.path {base}}.")
 4.     └─rlang::abort(...)
Execution halted

```

What am I doing wrong?

Felixmil:
> What am I doing wrong?

I think this is because testthat's working directory is not yet defined before test_check() runs.

This may work: `source(file.path("tests", "data", "create_and_save_data_combined_objects.R"))`

Felixmil:
I read the testthat documentation again and the testthat docs and R Packages book recommend keeping testthat.R unmodified (just library() calls + test_check()).                                                                      
                                                                                                                                                                                                      
The fix is to move all setup logic (parallel config, data object creation) into tests/testthat/setup.R, which is the dedicated mechanism for one-time pre-test setup. Inside setup.R, test_path() works correctly because testthat's context is active.

Also, we already have some code there so we just better use it.

PavelBal:
> Each worker begins by loading testthat and the package being tested. It then runs any setup files (so if you have existing setup files you’ll need to make sure they work when executed in parallel).

Does this mean that setup will be run on each runner? Then, setting the parallel options would make no effect as they must be set BEFORE the runners are created..

PavelBal:
https://share.google/aimode/EJ9zpyARzq1dpDQQ5

So either testthat.R or in the config?

Felixmil:
> https://share.google/aimode/EJ9zpyARzq1dpDQQ5
> 
> So either testthat.R or in the config?

Don't have access but you're right that setup.R runs inside each worker after they're spawned, so setting TESTTHAT_CPUS there would be too late. 

The parallel toggle is now handled via `Config/testthat/parallel: true` in DESCRIPTION (as recommended by testthat). 
The CPU count is set dynamically in testthat.R before test_check(), which is the only place that runs in the main process before workers are created.

tests/testthat/teardown.R
```diff
@@ -1,3 +1,11 @@
 # This file is run once after all tests
 # Clean up watermark option
 options(ospsuite.plots.watermarkEnabled = NULL)
+
+# Remove files created in "create_and_save_data_combined_objects.R"
```
PavelBal: @Felixmil I wonder if this can work in parallel...

tests/testthat/test-1-plot-observed-vs-simulated.R
PavelBal, line 1: According to documentation, executing heavy tests at the beginning might speed up the testing significantly. So I ahve renamed all heavy tests by adding a "test-1-" prefix.

PavelBal:
@coderabbit full review

PavelBal:
@Felixmil Now I am thinking... Does it make sense at all to write the DCs into files and read them again in the same session? I think it is better to define them as test-global variables (like we are already loading simulaitons in helper-for-tests) and use them directly in the respective tests...

Right now we could still have a case where one worker just finished its job and removes the rds, while the over worker wants to read them.

Felixmil:
The risk: datacombined objects could get mutated during tests. But we would notice easily if it happens.

Will look into a way to make this foolproof.

PavelBal:
@Felixmil I am on it, let me do it

Felixmil:
I would do something like this:

## helper-for-tests.R   
```r                                                                                                                                                                     
  .oneObsDC <- new DataCombined()                                                                                                                                                                     
  .oneObsDC$addDataSets(...)                                                                                                                                                                          
                                                                                                                                                                                                      
  # Accessor returns a fresh clone each call                                                                                                                                                          
  oneObsDC <- function() .oneObsDC$clone(deep = TRUE)
```                                                                                                                                                                                                      
                             
## test-*.R                                                                                                                                                                                          
```r                             
  test_that("some plot looks correct") {
      plotIndividualTimeProfile(oneObsDC())   # clone created, used, discarded
  }                                                                                                                                                                                                   
   
  test_that("another test") {                                                                                                                                                                         
      dc <- oneObsDC()         # different clone
      dc$addDataSets(extra)    # mutating dc does NOT affect .oneObsDC                                                                                                                                
      plotIndividualTimeProfile(dc)                                                                                                                                                                   
  }
```

PavelBal:
> I would do something like this:
> 
> ## helper-for-tests.R
> ```r
>   .oneObsDC <- new DataCombined()                                                                                                                                                                     
>   .oneObsDC$addDataSets(...)                                                                                                                                                                          
>                                                                                                                                                                                                       
>   # Accessor returns a fresh clone each call                                                                                                                                                          
>   oneObsDC <- function() .oneObsDC$clone(deep = TRUE)
> ```
> 
> ## test-*.R
> ```r
>   test_that("some plot looks correct") {
>       plotIndividualTimeProfile(oneObsDC())   # clone created, used, discarded
>   }                                                                                                                                                                                                   
>    
>   test_that("another test") {                                                                                                                                                                         
>       dc <- oneObsDC()         # different clone
>       dc$addDataSets(extra)    # mutating dc does NOT affect .oneObsDC                                                                                                                                
>       plotIndividualTimeProfile(dc)                                                                                                                                                                   
>   }
> ```

Let's do it in a separate PR later (if needed... now it works and I would just leave it as is). I need this PR merged to continue with MoBiR.

Felixmil:
> Let's do it in a separate PR later (if needed... now it works and I would just leave it as is). I need this PR merged to continue with MoBiR.

Here: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1869

Felixmil (approved):

PR #1903, opened 2026-04-23
Update snapshots following ospsuite.plots update (showtext removal)

Felixmil:
There are some noticeable differences so I would like a round of review to know if anything should be changed on the ospsuite.plots side. (or here).

Snapshots generated using ospsuite.plots 1.1.0.9008.

## Summary by CodeRabbit

* **Chores**
  * Updated internal package management metadata.


PavelBal (dismissed):
OMG, legend entries are finally not messed up.

KatrinCoboeken (dismissed):
Looks great

pchelle (dismissed):
It looks so much better ! 
The sizes of the symbols are now consistent (I also saw that the triangles are correctly centered as well)

tests/testthat/_snaps/plot-with-ospsuite-plots-residuals/defaults-as-qq-plot.svg
pchelle, line 1: Since all the symbols are now consistent, there is actually only one plot whose symbols feels not the same size.
Which is the qq plot below, I have no idea why.

Yuri05:
The only difference I see: the open symbols in the legend by name are now filled.
If it's ok for everyone - OK for me too :)
image

tests/testthat/_snaps/plot-with-ospsuite-plots-residuals/defaults-as-qq-plot.svg
Felixmil: Good catch ! I think this is because it uses `ggplot2::stat_qq` and not our new `geom_point_osp`. I will create a new `stat_qq_osp` in ospsuite.plots

PavelBal:
> If it's ok for everyone - OK for me too :)

Yes, this looks much better and is consistent with the symbols shown in the actual plot.

Felixmil:
Let's wait for https://github.com/Open-Systems-Pharmacology/OSPSuite.Plots/pull/117#event-24801670165 to be merged so QQplots also benefits from the new geom_point_osp.

tests/testthat/_snaps/plot-with-ospsuite-plots-residuals/defaults-as-qq-plot.svg
Felixmil: Fixed when using ospsuite.plots 1.1.0.9009 (cf. 0aaaed5)

pchelle (approved):

pchelle:
@Felixmil 
Is it because of renv that windows failed to install: "S7", "cowplot", "ggh4x", "ggnewscale", "ggplot2", "ggtext", "ospsuite.plots", "patchwork", "tlf"

Felixmil:
Yes because in renv.lock, S7 targets CRAN and not PPM. However, 0.2.2 version of was released on CRAN, making 0.2.1-1 unavailable ?
macOS and Linux pass because they use Posit Package Manager which keeps older versions.

I will merge and trigger a renv.lock update on main.

PR #1912, opened 2026-04-24
Remove `ospsuite.plots::setDefaults()` from tests

PavelBal:
Looks like setting `ospsuite.plots::setDefaults()` during package load does not work? I removed explicit call to this in tests, and all plots have changed. I would expect that if `ospsuite.plots::setDefaults()` called during initialization, this should should not be required.


PavelBal:
Or it works, but only partially?

PavelBal:
No way! It was `showtext` after all.

rengelke:
Redundant if completing #1933 first

PavelBal:
I wanted to do it step-by-step, to evaluate how the changes affect the snapshots.

rengelke (approved):

PR #1921, opened 2026-04-28
Allow-list inst/lib and drop 32 unused .NET assemblies

### LINKED ISSUE #1587: Review supplied libs
Review if all dependencies are really required. RIght now with MoBi.R we are shipping 114 dlls.

rwmcintosh:
## Summary
- Replaces the implicit "ship everything `dotnet build` produces" PostBuild copy in `DependencyManager.csproj` with an explicit `<AllowedAssembly>` allow-list plus a sweep that deletes any inst/lib `*.dll` not on the list. Future upstream transitive deps surface as a test-time `FileNotFoundException` instead of silent bloat.
- Drops 32 confirmed-unused DLLs from `inst/lib`. Phase A captured `AppDomain.CurrentDomain.GetAssemblies()` after the testthat suite on **Windows, macOS arm64, and Ubuntu Linux** — all three platforms agreed identically on the never-loaded set, so the same 32 are removable everywhere.
- `inst/lib` shrinks from 102 → 69 managed DLLs (~5.5 MB freed). Notable removals: full Serilog family, several `System.Security.Cryptography.*`, `MathNet.Numerics`, `Dapper`, `System.Data.SQLite`, `PKSim.Assets.Images`, plus two runtime-shadowed entries (`System.Collections.Immutable`, `System.Reflection.Metadata`) that the .NET 8 host resolves from its own runtime anyway.
- Removals were made one-batch-at-a-time with a full `devtools::test()` pass between each (`FAIL 0 | PASS 2011` on Windows for every batch), so any single batch can be reverted in isolation if a regression surfaces.

## Test plan
- [x] Pull on each platform and run:
  - `dotnet build shared/DependencyManager/DependencyManager.sln --configuration Release` — verify the build log emits `Allow-list sweep: deleting <name>.dll` for ~30 names, and inst/lib lands at 69 `*.dll`.
  - `Rscript -e 'devtools::test()'` — should be `FAIL 0 | PASS 2011`. Note: at the moment two `plotIndividualTimeProfile` SVG snapshot tests fail on this branch, but they fail identically at clean `v13`/HEAD with no audit changes — pre-existing drift unrelated to this PR. Run `testthat::snapshot_review("1-plot-individual-time-profile/")` separately to triage.
- [x] Verify on Windows.
- [x] Verify on macOS arm64.
- [x] Verify on Ubuntu Linux.
- [x] Render vignettes via `devtools::build_vignettes()` on at least one platform — vignettes exercise documented user workflows the unit tests miss; if any vignette fails with `FileNotFoundException`/`TypeLoadException`, the named assembly is the line to re-add to `<AllowedAssembly>`.
- [x] Review the allow-list in `shared/DependencyManager/src/DependencyManager.csproj` and confirm each entry corresponds to a DLL we want to keep shipping.

Closes #1587.

## Summary by CodeRabbit

* **Chores**
  * Updated package dependency to the latest version
  * Enhanced build process with improved assembly validation to ensure only approved libraries are shipped


shared/DependencyManager/src/DependencyManager.csproj
rwmcintosh, line 1: This is a whitelist style update... new assemblies that get included upstream will be removed unless they appear as AllowedAssembly. That prevents creep.

PavelBal (approved):

Yuri05:
Questions:

1) What is the procedure for the future when adding new dependencies to PK-Sim/MoBi? Wait until the loading of ospsuite-R packages crashes due to a missing dependency (is it always clear which one is missing?) and then start adding missing libs one by one until everything runs again?

> * Removals were made one-batch-at-a-time with a full `devtools::test()` pass between each (`FAIL 0 | PASS 2011` on Windows for every batch), so any single batch can be reverted in isolation if a regression surfaces.

2) Such a removal is in principle a multipass approach: try to remove every DLL one by one; if at least 1 DLL was removed - repeat from the begin (could happen that some DLL could not be removed before due to being indirect dependency, but can be removed now). Was it done like that? Or just one "run through all DLLs" iteration?

rwmcintosh:
> Questions:
> 
> 1. What is the procedure for the future when adding new dependencies to PK-Sim/MoBi? Wait until the loading of ospsuite-R packages crashes due to a missing dependency (is it always clear which one is missing?) and then start adding missing libs one by one until everything runs again?

If you don't use the assembly in any pathway of the existing tests then it won't crash. If the assembly is needed it will crash on the OSPSuite-R test phase. The exception message says which assembly cannot be loaded. You have to allow them through per-item so as to not silently bloat the number that are delivered.

> 
> > * Removals were made one-batch-at-a-time with a full `devtools::test()` pass between each (`FAIL 0 | PASS 2011` on Windows for every batch), so any single batch can be reverted in isolation if a regression surfaces.
> 
> 2. Such a removal is in principle a multipass approach: try to remove every DLL one by one; if at least 1 DLL was removed - repeat from the begin (could happen that some DLL could not be removed before due to being indirect dependency, but can be removed now). Was it done like that? Or just one "run through all DLLs" iteration?

Actually, Claude ran through all the tests and then called [`AppDomain.GetAssemblies`](https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.getassemblies?view=net-10.0) and compared against the inst/lib folder. This 'audit' was done for each platform and the list of non-used assemblies was the same for all platforms. Then we removed them in batches and re-ran the tests after each batch.

So, we would not need to recursively test because GetAssemblies is telling us all the assemblies that were loaded.

Yuri05:
> > Questions:
> > 
> > 1. What is the procedure for the future when adding new dependencies to PK-Sim/MoBi? Wait until the loading of ospsuite-R packages crashes due to a missing dependency (is it always clear which one is missing?) and then start adding missing libs one by one until everything runs again?
> 
> If you don't use the assembly in any pathway of the existing tests then it won't crash. If the assembly is needed it will crash on the OSPSuite-R test phase. The exception message says which assembly cannot be loaded. You have to allow them through per-item so as to not silently bloat the number that are delivered.
> 
> > > * Removals were made one-batch-at-a-time with a full `devtools::test()` pass between each (`FAIL 0 | PASS 2011` on Windows for every batch), so any single batch can be reverted in isolation if a regression surfaces.
> > 
> > 
> > 
> > 2. Such a removal is in principle a multipass approach: try to remove every DLL one by one; if at least 1 DLL was removed - repeat from the begin (could happen that some DLL could not be removed before due to being indirect dependency, but can be removed now). Was it done like that? Or just one "run through all DLLs" iteration?
> 
> Actually, Claude ran through all the tests and then called [`AppDomain.GetAssemblies`](https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.getassemblies?view=net-10.0) and compared against the inst/lib folder. This 'audit' was done for each platform and the list of non-used assemblies was the same for all platforms. Then we removed them in batches and re-ran the tests after each batch.
> 
> So, we would not need to recursively test because GetAssemblies is telling us all the assemblies that were loaded.

ok, thx. let's see how it works in the future :)

Yuri05 (approved):

PR #1927, opened 2026-05-03
fix: replace non-ASCII µ literals with unicode escapes to resolve R CMD check warning

### LINKED ISSUE #1894: non-ASCII characters
From R CMD Check:

```r
Warning: Found the following files with non-ASCII characters:
  R/utilities-quantity.R
  R/utilities-units.R
```

veerlakshay:
R CMD check warns about non-ASCII characters in source files because `µ` (U+00B5) was used as a literal byte in strings and doc comments referencing the `kg/µmol` molecular weight unit.

The fix replaces all literal `µ` characters with `µ` R unicode escapes. R parses these to the identical U+00B5 byte at load time, so runtime behavior and unit string matching against the XML-loaded `ospUnits` keys are unchanged.

Closes #1894

## Summary by CodeRabbit

* **Documentation**
  * Updated utility function documentation examples and descriptions for consistency.

* **Refactor**
  * Improved code patterns in internal utility functions to enhance maintainability.


PavelBal (commented):
The rendered documentation will still show "µ", right? So the user will not be exposed to unicode escape?

veerlakshay:
Yes, the rendered documentation will still show "µ", R resolves the unicode escape at parse time, so users won't see any difference.

PavelBal (approved):

PR #1933, opened 2026-05-06
Do not set `ggplot2::theme_update(legend.title = ggplot2::element_blank())` in tests

PavelBal:
As discussed, I think test snapshots should show the default behavior of the package.

In this PR, I removed `ggplot2::theme_update(legend.title = ggplot2::element_blank())` from the tests.

I think that the plots look better with `ggplot2::theme_update(legend.title = ggplot2::element_blank())`. We have different options:

1. Add `ggplot2::theme_update(legend.title = ggplot2::element_blank())` to package defaults (either`ospsuite` or `ospsuite.plots`)
2. Do not add to defaults, but describe in the vignette as "best practices".

I am not sure what is better. For "cleanlincess", maybe 2? If we choose 2., then @pchelle I would set `ggplot2::theme_update(legend.title = ggplot2::element_blank())` in esqlabsR defaults.

## Summary by CodeRabbit

* **Tests**
  * Updated multiple plotting snapshot tests to stop applying custom default plot styling overrides, including legend title suppression and legend position/justification settings.
  * Adjusted the population time profile test setup by removing the prior default percentiles configuration.
  * Refreshed legend-related snapshot expectations so plots now match the standard styling behavior instead of test-specific legend overrides.


Felixmil:
I think it is a good opportunity to refactor all these theme_update calls into a helper in [tests/testthat/helper-for-tests.R](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/f9e913fdf0031fb10c368117a6283120791cbf2d/tests/testthat/helper-for-tests.R)

PavelBal:
> I think it is a good opportunity to refactor all these theme_update calls into a helper in [tests/testthat/helper-for-tests.R](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/f9e913fdf0031fb10c368117a6283120791cbf2d/tests/testthat/helper-for-tests.R)

Or remove them completely from tests ;)

Felixmil:
Yes, I would run test with default theme, unless tests are especially about testing theme features.

pchelle:
I agree. 

I think the first question is what is our purpose in these snapshot tests ?
If assessing the theme, indeed, we should make specific tests about how setting defaults theme affect the plot(Testing both adding a `theme` layer and using default `theme_update()`)

If not, we can indeed leave default theming as is and evaluate what is within the plots with snapshot tests.

And a vignette about good practices and __how to__ sounds a good idea

PavelBal:
> I think the first question is what is our purpose in these snapshot tests ?

I think - to test how the package behaves when loaded, no additional edits. If the plots do not look satisfactory, then the defaults must be adjusted.

PavelBal:
So, what is the better approach? Set this in defaults (on package load), or leave as is and describe in vignette?

@rengelke @Felixmil @pchelle @KatrinCoboeken I would like to hear your opinion.

Felixmil:
The snapshot tests should reflect what users see when they load the package with no extra configuration, so running them against the true default theme is the right approach.

On the "set it in defaults vs. vignette" question: the right answer depends on the scope of the side effect. `ggplot2::theme_update()` on package load modifies the global ggplot2 theme for the entire R session, so any plot the user creates after `library(ospsuite)` would be affected, not just ospsuite plots. That is probably not what we want.

The cleanest fix is to apply `legend.title = element_blank()` (and any other settings that improve the look) inside each ospsuite plotting function directly on the returned ggplot2 object. Since the functions already build on a ggplot2 object and apply additional layers via `+`, adding the theme tweak is a one-liner:

```r
# at the end of e.g. plotIndividualTimeProfile(), before returning:
plot <- plot + ggplot2::theme(legend.title = ggplot2::element_blank())
return(plot)
```

This keeps the effect local to the returned plot object and has no impact on any other ggplot2 plots in the user's session.

On the test side, the remaining `theme_update` calls for `legend.position` and `legend.justification` should either be moved into the plotting functions themselves (same rationale as above), or wrapped in `withr::local_theme()` if they are truly test-specific, so they do not leak global state between tests.

**Action plan if we go this way:**

1. For each ospsuite plotting function, append the relevant `ggplot2::theme()` settings before returning the plot object.
2. Remove all `theme_update` calls from the test files. If any are test-specific, replace them with `withr::local_theme()`.
3. Regenerate the snapshots so they reflect the updated function defaults.

KatrinCoboeken:
I deleted the theme_update so all figures are created with defaults. Which looks fine for me.

PavelBal:
I think we can then merge this PR.

@KatrinCoboeken what do you think about Felix' proposal? I think that the plots look better with this adjustment and would prefer to have them by default. If you agree, we can open a new PR.

However if you would prefer to keep the plotting functions without further theme updates, we can implement it separately for esqlabsR, no problem.

tests/testthat/_snaps/plot-with-ospsuite-plots-individualTimeProfiles/with-secaxis.svg
Felixmil, line 1: `legend.title = ggplot2::element_blank()` was removing the uncessary legend titles: `name`, `group`, `name`.

I suggest we move them to a ggplot2::theme() layer.

KatrinCoboeken:
I agree that the plots looks better with legend.title = ggplot2::element_blank(), and think the best place is as Felix suggested in the plotting functions. To be clean I would do this in another PR.  May be it would be good to add  an example how to switch it on the eplotting vignette.

rengelke (approved):

Issue #1916, opened 2026-04-27
macOS CI snapshot tests fail due to wrong R version


Felixmil:
The macOS runner is silently installing R 4.5.3 instead of R 4.6.0 (Linux and Windows install 4.6.0 as expected), causing vdiffr plot snapshot tests to fail because plots render slightly differently between R versions.

Tracking issue upstream: https://github.com/r-lib/actions/issues/1053

PR #1955, opened 2026-06-02
Fixes #1834 refactor DataCombined plot examples

pchelle:
This PR exports the `DataCombined` object created in the corresponding vignette so that users and us can re-use it directly as is for documentation and tests without having to load simulation results and obs data every time.

Not sure if this is the best way to tackle the issue. 
So, let me know if this makes sense on your side as well.

I have included the some snapshots of examples from pkgdown documentation for the exported dataset and plot functions

| Exported dataset | tlf Obs vs Pred | osp plot Obs vs Pred |
|--------------------|-----------------|-------------------------|
| dcAciclovir | image |  image |

## Summary by CodeRabbit

* **New Features**
  * Added a built-in `dataCombinedAciclovir` prebuilt `DataCombined` object combining simulated and observed data.
* **Documentation**
  * Updated many plot/residual examples to use `dataCombinedAciclovir`.
  * Simplified examples and removed non-runnable wrappers so they render and run more smoothly.
  * Updated the vignette and manual pages to reference the new dataset.
* **Refactor**
  * Modernized internal data/plot preparation code using native piping, preserving intended behavior.
* **Chores**
  * Included the new dataset documentation in the package build sequence.


PavelBal:
I like the fact that we get rid of the  `dontrun` examples.

But I don't understand where this DataCombined object is loaded?

Also please resolve the conflicts.

pchelle:
> I like the fact that we get rid of the `dontrun` examples.
> 
> But I don't understand where this DataCombined object is loaded?

The data needs to be saved as a .rda file in `/data/<dataset-name>.rda` to be loaded (following what's advised [here](https://r-pkgs.org/data.html#sec-data-data)).

In case, I included the origin story of the data in the package documentation (`data.R`), so it it reproducible and can be easilt updated if `DataCombined` objects evolve.

DESCRIPTION
Felixmil, line 1: remotes should target developement dependencies (no `@*release`) while in development.

\+ It is automatically handled by github workflows now

data/dataCombinedAciclovir.rda
Felixmil, line 1: Mention this exists in related vignettes ?

PavelBal (dismissed):

vignettes/data-combined.Rmd
```diff
@@ -85,6 +85,13 @@ myDataCombined$addDataSets(
 )
 ```
 
+The package exports `dataCombinedAciclovir` — a `DataCombined` object equivalent to the example built above — which 
+can be used directly to illustrate and test how `DataCombined` objects work:
```
Felixmil: weird hard linebreak

Felixmil (approved):

PavelBal (dismissed):

Felixmil (changes requested):
Actually, since it is a user facing change, we need a NEWS entry

Felixmil (dismissed):

pchelle:
I was not expecting the snapshot would lead to different legend order in the ubuntu tests:

| Reference | Ubuntu test |
|------------|---------------|
|  both-custom-separate-legend | both-custom-separate-legend new |

Felixmil:
# CI snapshot failures on PR #1955: investigation report

Two vdiffr snapshot tests fail on CI (`test-1-plot-individual-time-profile.R:87` and `:98`), both for `plotIndividualTimeProfile(oneObsSimDC(), showLegendPerDataset = TRUE)`. The failures are not caused by this PR (its only plotting changes are `@examples` docs) and reproduce neither locally nor on Windows. The difference is the vertical stacking order of two legends, not fonts and not locale.

## 1. Each runner ran a different R version

All three R-CMD-check runners requested `r-version: release`, but `release` resolved to a different build per platform on the run date (2026-06-26):

| Runner | R version | Snapshot result |
|---|---|---|
| Windows | 4.6.0 (2026-04-24) | PASS (FAIL 0 \| PASS 1638) |
| Linux (ubuntu) | 4.6.1 (2026-06-24) | FAIL (2 tests) |
| macOS | 4.6.1 (2026-06-24) | FAIL (2 tests) |
| Local (macOS) | 4.6.0 | PASS |

ggplot2 was 4.0.3 on every runner, so this is not a package-version difference. The only variable that tracks pass/fail is the R version: 4.6.0 passes, 4.6.1 fails.

Why the split: `r-lib/actions/setup-r@v2` installs R from CRAN's per-OS binaries, and those are published at slightly different times. R 4.6.1 source released 2026-06-24; by 2026-06-26 the Linux and macOS binaries existed (so `release` resolved to 4.6.1), but the Windows installer had not been published yet (so `release` fell back to 4.6.0). Nothing in the repo is misconfigured; this is transient publication-timing skew during the 4.6.0 to 4.6.1 rollout.

## 2. The impactful R change

R 4.6.1 changed `abbreviate()` for non-ASCII / "wide" characters (now uses `iswlower()`/`*upper()`). That is the one 4.6.1 change in the path that touches plot rendering.

## 3. Why that change flips the snapshot

For `showLegendPerDataset = TRUE` with one observed and one simulated dataset, the plot builds three guides:

- `colour` -> `order = 1` (explicit), always on top
- `shape` (observed label `Vergin 1995.Iv`) -> `order = 0`
- `linetype` (simulated label `Organism|...Plasma`) -> `order = 0`

The shape and linetype guides have equal, unset `order` (0). When `order` ties, ggplot2 breaks the tie by each guide's `hash`. That hash is computed over the guide's processed key contents, including the label strings, which pass through R's `abbreviate()`/character-normalization path. R 4.6.1's `abbreviate()` change alters the hash, which flips the sort order of the two `order = 0` guides, so the two legends swap position and the SVG snapshot no longer matches.

This latent instability was previously masked: PR #1977 removed the on-load `ospsuite.plots::setDefaults()`, which had set ggplot2 defaults that hid the raw hash tie-break. With it gone, the R 4.6.0 to 4.6.1 `abbreviate()` change is enough to tip the order over. This is the same R-version-dependent rendering class tracked in issue #1916.

PavelBal (approved):

Felixmil (approved):

PR #1958, opened 2026-06-08
V13

PavelBal:
This PR merges the **v13 development line into `main`**, making the ospsuite v13 codebase (currently `13.0.0.9001`) the default WIP branch. The individual changes were developed and reviewed in the constituent PRs (#1925–#2008); this PR is the integration step only.

### Major changes
- **MoBi project support**: load `.mbp3` projects, query modules, individuals, expression profiles, and simulations, and assemble simulations from project building blocks. New classes `MoBiProject`, `MoBiModule`, `SimulationConfiguration`, `MoleculesBuildingBlock`, plus helpers for creating and saving Initial Conditions, Parameter Values, Individual, and Expression Profile building blocks. See `vignette("mobi-projects")`.
- **PK-Sim snapshot support**: `loadSimulationsFromSnapshot()` loads simulations stored in a snapshot file as `Simulation` objects (#1929, #1972).
- Population simulations via `simulation$population`; `runSimulations()` can run mixed individual + population simulations in a single call (#1987).

### Breaking changes
- **.NET 10 runtime is now required** (previously .NET 8). Bundled assemblies target `net10.0`.
- `createIndividual()` / `createPopulation()` do not work with models created before v13 (refined absorption model). Re-create models from snapshot with the latest PK-Sim, or stay on ospsuite v12.
- `SimulationRunOptions$checkForNegativeValues` moved to `SolverSettings` (`simulation$solver$checkForNegativeValues`); the old constructor argument is deprecated and has no effect.

See [NEWS.md](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/v13/NEWS.md) for the full list, including minor improvements and bug fixes.

## Merge plan
- v13 is fully synced with `main` (0 commits behind), no conflicts.
- Merge with a **merge commit** (not squash/rebase) to preserve traceability to the constituent PRs.
- After the merge, **`main` remains the default WIP branch** and carries the v13 line; the `v13` branch is auto-deleted on merge. All release machinery (`merge-to-main.yaml`, `publish.yaml`, dev-version bump) stays keyed to `main` and needs no changes.
- Before merging, a **`v12` maintenance branch** is cut from current `main` (last v12 dev state, `12.4.4.9004`) so v12 patches remain possible.

## Follow-ups after merge
- Open PRs targeting `main` need a re-check against the v13 content: #2006, #1510.
- Doc pass: `AGENTS.md` (v13 worktree instructions) and `.github/copilot-instructions.md` (branch layout description) reference the pre-merge two-branch setup.

Issue #347, opened 2020-03-23
createIndividual: error message truncated


msevestre:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/345

msevestre:
@Yuri05 @PavelBal  This is an issue in general with rCLR I believe. Error messages are being truncated.
For the next version, we should come up with a better way to show managed error messages (e.g exception coming from PKSIm or Core that are thrown by us)

I don't want to implement something quickly that does not make sense.


Example of truncated message

```
Exception
Message: Could not create individuals with given constraint:

Method:  Void optimizeVolumes(PKSim.Core.Model.Individual, OSPSuite.Core.Maths.Random.RandomGenerator)
Stack trace:
   at PKSim.Core.Services.CreateIndividualAlgorithm.optimizeVolumes(Individual individual, RandomGenerator randomGenerator)
   at PKSim.Core.Services.CreateIndividualAlgorithm.distributeParameterFor(Individual individual, RandomGenerator randomGenerator, Action`2 action)
   at PKSim.Core.Services.CreateIndividualAlgorithm.Optimize(Individual individual)
   at PKSim.Core.Model.IndividualFactory.CreateAndOptimizeFor(OriginData originData, Nullable`1 seed)
   at PKSim.R.Services.IndividualFactory.CreateIndividual(IndividualCharacteristics individualCharacteristics)
```

msevestre:
Maybe the signature of the R API could change to return an structure: e.g. instead of returning the data, we would have a result object: For ex
{
  data: individual,
  error: xxxx
}

if error is not NULL then there is an error 

Let's think about it

PavelBal:
I do not really like such "nester" output structures. Also, having always to check whether there is an error or not before using the output. However, such a check could be performed before returning the output to the user.

msevestre:
@PavelBal Absolutely. This would just be for the direct R API calls to .NET and should not be visible to the user of the lib

PR #1961, opened 2026-06-13
Backport v12-safe R cleanups from v13, round 2 (#1958)

PavelBal:
Round 2 of extracting v12-applicable changes from the WIP v13 PR #1958, keeping `main` and `v13` close. **No new functionality, no new .NET library requirement** — every assembly stays the same (`OSPSuite.Core` / `OSPSuite.R` / `PKSim.R` / `System.Runtime`); none is switched (e.g. never `OSPSuite.Core` → PK-Sim). All assembly-qualified strings were verified to resolve at runtime under the current v12 rSharp/.NET.

## Changes

**Assembly-qualified type strings** (qualify the existing assembly, do not change it)
- `data-column.R`: `WithDimensionExtensions`, `WithDisplayUnitExtensions` → append `, OSPSuite.Core`
- `simulation.R`: `MODEL_CORE_SIMULATION_EXTENSIONS` → append `, OSPSuite.Core` (**only** this constant — the v13 `configuration` / `calculationMethodFor` / `SimulationConfiguration` feature work is deliberately excluded)
- `utilities.R`: `System.Enum` → `, System.Runtime`; `ForceGC` → `, OSPSuite.R`
- `utilities-pksim.R`, `utilities-individual.R`, `utilities-population.R`: `"PKSim.R.Api"` → `"PKSim.R.Api, PKSim.R"`

**Other**
- `utilities-quantity.R`: `getMolWeightFor()` uses cached `.getCoreTaskFromCache("ContainerTask")` (matches the rest of the file)
- `utilities-plotting.R`: drop the stale `# TODO 2DO: TLF #347` comment above the commented-out `capSize` line. **capSize stays commented** — `tlf::ThemeAestheticSelections$new()` (tlf 1.6.2.9001) has no `capSize` argument, so uncommenting it would throw `unused argument`. Cap size is already applied via `tlf::setDefaultErrorbarCapSize()` in the plot-*.R functions.
- `utilities-individual.R`: doc wording + `@param` order fixes.

**New exported helpers** (+ NAMESPACE + Rd)
- `isSupportedUnit(unit)` — `utilities-units.R`
- `validateIsNamedList(x, varName)` — `utilities.R`

**New internal helpers** (`@noRd`, no caller in v12 yet — used by v13-only building-block code; shipped to match v13)
- `.getParentContainerPath(path)` — `utilities-path.R`
- `.validateScalarStringOrNull(x, varName)` — `utilities.R`

## On the `capSize` request
The ask was to uncomment `capSize = generalPlotConfiguration$errorbarsCapSize`. That is **not safe** with the current `tlf` dependency (no `capSize` arg → runtime error), and would be redundant — cap size already flows through `setDefaultErrorbarCapSize()`. Left commented, matching v13.

## Related finding (v13, not in this PR)
`loadDataSetsFromExcel()` in v13 `utilities-data-set.R` has `dataImporterTask <- .getCoreTaskFromCache("DataImporterTask")` at ~line 284 that is **dead** — it is reassigned with the identical call right before first use (~line 310) and never referenced in between. `main` does not have this redundant line. Recommend deleting it from v13.

## Verification
- `pkgload::load_all()` clean.
- Runtime smoke tests (all pass): `System.Enum, System.Runtime` enum-name resolution via `DataSet$yErrorType`; `data-column` assembly-qualified extensions; `isSupportedUnit`; `validateIsNamedList`; `clearMemory`; `initPKSim()` + `createIndividual()` (confirms `PKSim.R.Api, PKSim.R`).
- Affected testthat files pass under the real suite.

## Summary by CodeRabbit

* **New Features**
  * `isSupportedUnit()`: Check whether a unit is supported by the platform
  * `validateIsNamedList()`: Validate that input is a properly named list with non-empty, non-missing names
* **Documentation**
  * Added help pages for the new validation functions
  * Refreshed wording for individual creation and related argument descriptions
* **Bug Fixes**
  * Improved platform/runtime resolution for several .NET integration calls (unit/extension calls, simulation resolution, PK-Sim initialization, population creation, forced garbage collection, enum lookup, and container task retrieval)
* **Chores**
  * Added an internal path helper and cleaned up minor plotting documentation comments


rengelke (dismissed):

PavelBal:
@copilot add a news entry for the new exported functions

PavelBal:
@rengelke pls re-approve, added an entry to NEWS

sorinvoicu:
isSupportedUnit() and validateIsNamedList() are now public API, but not covered by testthat tests

sorinvoicu:
- isSupportedUnit isn't scalar-safe.

  isSupportedUnit <- function(unit) {
    validateIsString(unit)
    return(unit %in% unlist(ospUnits))
  }

If validateIsString permits a length>1 vector (it does in ospsuite.utils), %in% returns a vector, so the function silently returns multiple logicals despite docs/examples implying a single TRUE/FALSE.

R/utilities-path.R
```diff
@@ -32,3 +32,23 @@ toPathString <- function(...) {
...
+#' @examples
+#' .getParentContainerPath("Organism|Organ|Liver") # "Organism|Organ"
+#' @noRd
+.getParentContainerPath <- function(path) {
+  validateIsString(path)
+  pathEntries <- toPathArray(path)
+  if (length(pathEntries) <= 1) {
+    return(NULL)
+  }
+  parentPathEntries <- head(pathEntries, -1)
```
nautilus69: The package imports only tail (NAMESPACE:180) and elsewhere uses the explicit utils::head convention (R/ospsuite-env.R:4). This new bare head and should become: 

 utils::head(pathEntries, -1) (matches existing convention) or pathEntries[-length(pathEntries)] (no dependency).

R/utilities-individual.R
nautilus69: NOT SPECIFICALLY FOR THIS PR BUT IS AFFECTED:

The function returns an IndividualCharacteristics object (return(individualCharacteristics), line 172), but:
- @return (line 107) says "An array of ParameterValue containing the value of each individual parameter" which is wrong object entirely.
- Title (line 86) says "Creates an individual using the PK-Sim Database." is misleading (it builds characteristics, not an individual).

R/utilities.R
```diff
@@ -139,3 +139,56 @@ clearMemory <- function(clearSimulationsCache = FALSE) {
...
+#' @param x Object to validate as a named list.
+#' @param varName Name of the variable being validated (used in error message).
+#'
+#' @returns invisible TRUE if the validation passed, otherwise an error is thrown.
+#'
+#' @export
+#' @examples
+#' validateIsNamedList(list(a = 1, b = 2), "myVar") # passes
+#' # validateIsNamedList(list(1, 2), "myVar") # throws an error
+validateIsNamedList <- function(x, varName) {
```
nautilus69, line 178: the family auto-derives the reported variable name via deparse(substitute(object)) (validation-type.R:76-88). 
The PR's validateIsNamedList(x, varName) instead requires the caller to hand-type varName, which can drift from the real symbol and is inconsistent with every other validateIs*.

This is necessary to decide before merge specifically because validateIsNamedList is being exported BECAUSE once released, the (x, varName) signature is locked. 
Either drop varName and use deparse(substitute(x)) to match the family, or consciously keep the explicit form. (Applies equally to the internal .validateScalarStringOrNull, though that one isn't exported, so it can change later.)

nautilus69 (changes requested):

R/utilities.R
PavelBal: Thanks — we considered switching to `deparse(substitute(x))` to match the `validateIs*` family, but decided to **consciously keep the explicit `(x, varName)` form**.

The v13 callers pass `varName` from inside R6 active-binding setters, e.g. `validateIsNamedList(value, "selectedInitialConditions")` in `SimulationConfiguration.R`. There the bound symbol is `value`, so `deparse(substitute(x))` would report the unhelpful name `"value"` rather than the field name. The explicit argument yields the meaningful message there.

Keeping the signature also avoids diverging from v13 (6 call sites) where this helper is already used. The internal `.validateScalarStringOrNull(x, varName)` mirrors it for consistency (and, being unexported, can still change later if needed).

R/utilities-individual.R
```diff
@@ -91,10 +97,10 @@ createDistributions <- function(individualCharacteristics) {
 #' @param gestationalAge Gestational age of the created individual (for human species only using the Preterm population). Default is 40 Weeks
 #' @param gestationalAgeUnit Unit in which the gestational age value is defined. Default is week(s)
 #' @param moleculeOntogenies Optional list of `MoleculeOntogeny` that will be used to retrieve ontogeny information for molecules.
-#' @param seed Optional seed parameter to use to generate start values for the created individual algorithm.
 #' A `MoleculeOntogeny` is an object with the name a `molecule` property (e.g the name of the molecule as defined in your simulation)
 #' and an `ontogeny` property (e.g. the name of the predefined ontogeny to use for this molecule). The list of all available ontogenies
 #' can be accessed programmatically using the enum `StandardOntogeny`
```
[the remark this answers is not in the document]
PavelBal: Fixed in 33d8b55 — corrected to "an object with a `molecule` property".

R/utilities-units.R
```diff
@@ -46,6 +46,16 @@ validateUnit <- function(unit, dimension) {
...
 
+#' Check if unit is supported in the OSPSuite platform
+#'
+#' @param unit String name of the unit
+#' @details Returns `TRUE` if the provided unit is supported in the OSPSuite platform, otherwise `FALSE`
+#' @export
+isSupportedUnit <- function(unit) {
+  validateIsString(unit)
+  return(unit %in% unlist(ospUnits))
+}
```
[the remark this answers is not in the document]
PavelBal: Fixed in 33d8b55 — added `@return` (logical, supported or not) and `@examples` to `isSupportedUnit`.

R/utilities-path.R
PavelBal: Fixed in 31086111 — now `utils::head(pathEntries, -1)`, matching the package convention (NAMESPACE imports only `tail`; `utils::head` is used in `ospsuite-env.R`).

R/utilities-individual.R
PavelBal: Fixed in 31086111 — `createIndividualCharacteristics` now reads:
- title: "Create the characteristics of an individual using the PK-Sim Database."
- `@return`: "An `IndividualCharacteristics` object." (was incorrectly "array of `ParameterValue`")

`man/createIndividualCharacteristics.Rd` updated to match.

PavelBal:
Addressed the `isSupportedUnit` scalar-safety point (raised by Sorin) in 1e34f541.

`validateIsString` permits length>1 vectors, so `unit %in% unlist(ospUnits)` could return a logical vector, contradicting the documented single `TRUE`/`FALSE`. Added a scalar guard:

```r
isSupportedUnit <- function(unit) {
  validateIsString(unit)
  validateIsOfLength(unit, 1)
  return(unit %in% unlist(ospUnits))
}
```

Now a vector input errors instead of silently returning multiple logicals; scalar input still returns a single `TRUE`/`FALSE`. (Same code exists in v13 — will mirror the fix there.)

PavelBal:
@sorinvoicu @nautilus69 addressed your comments

nautilus69 (approved):

PR #1972, opened 2026-06-16
Add loadSimulationsFromSnapshot() R API

PavelBal:
## Summary

Implements the R API around the `SnapshotTask.LoadSimulationsFromSnapshot` binaries added in #1950 (which shipped the binaries but no R-side API).

- `loadSimulationsFromSnapshot(snapshotFile, simulationNames = NULL)` in `utilities-snapshots.R`: loads simulations stored in a snapshot file and returns them as `Simulation` objects, optionally filtering by (case-sensitive) simulation name. They behave like simulations loaded from a `.pkml`, so they can be inspected, modified, and run.
- New **"Working with snapshots"** vignette, also documenting the previously undescribed `runSimulationsFromSnapshot()` and `convertSnapshot()`.
- NEWS updated.

## Core dependency (resolved)

Snapshot-loaded simulations initially came back with a `NULL` `SimulationConfiguration`, so `Simulation$new()` failed in `CreateSimulationBuilderFor`. Fixed on the computational core (PK-Sim#3560); bundled assemblies bumped to **MoBi.R 13.0.121 / PKSim.R 13.0.81**, which attach the configuration.

## Test plan

- [x] `tests/testthat/test-snapshot-task.R`: load-all → 2, name filter → 1, missing → 0, run a snapshot-loaded simulation, argument validation. **9/9 pass.**

## Release Notes

* **New Features**
  * Added exported `loadSimulationsFromSnapshot()` to load simulations from snapshot `.json` files as `Simulation` objects, with optional case-sensitive `simulationNames` filtering and `ignoreIfNotFound` behavior (error when missing, or `NULL` entries while preserving requested order).

* **Documentation**
  * Updated `NEWS.md`, API reference, and the “Working with simulations” snapshots vignette; refreshed `runSimulationsFromSnapshot()` export option wording. Updated site navigation to include snapshots.

* **Tests**
  * Reworked snapshot-related tests to cover loading, filtering, missing-name handling, argument validation, and running loaded simulations.


R/utilities-snapshots.R
```diff
@@ -346,6 +346,73 @@ convertSnapshot <- function(..., format, output = ".", runSimulations = FALSE) {
...
+    netSimulations,
+    function(netSimulation) Simulation$new(netSimulation)
+  )
+  names(simulations) <- vapply(
+    simulations,
+    function(simulation) simulation$name,
+    FUN.VALUE = character(1)
+  )
+
+  simulations
```
Felixmil, line 447: Wrap in a trycatch to capture errors ? Also if returned simulations object is "empty", surface to the user that the provided simulationNames is probably wrong.

PavelBal: Errors will be thrown by .NET - better to show that message, than some generic if we wrap in trycatch, don't you think?

I also changed the behavior when a simulation with a specified name is not found. I think it should be an error insteand of a warning, with an option "ignoreIfNotFound" to ignore.

tests/testthat/test-1-utilities-snapshots.R
```diff
@@ -160,3 +160,99 @@ test_that("gather files handles files and directories", {
...
+
+  simulations <- loadSimulationsFromSnapshot(
+    snapshotFile,
+    simulationNames = simName
+  )
+
+  expect_length(simulations, 1)
+  expect_equal(simulations[[1]]$name, simName)
+})
+
```
nautilus69, line 204: This is not a bug just for the sake of testcoverage, this test was dropped during consolidation: 

A multi-name filtering regression test for loadSimulationsFromSnapshot()


test_that("loadSimulationsFromSnapshot accepts several names", {
    allNames    <- Names(loadSimulationsFromSnapshot(snapshotFile))
    simulations <- loadSimulationsFromSnapshot(snapshotFile, simulationNames = allNames)

    expect_length(simulations, length(allNames))
    expect_setequal(names(simulations), allNames)
  })

nautilus69 (approved):
With minor test-coverage request for multi-name filtering.

PR #1967, opened 2026-06-15
Blank legend title inside plotting functions (closes #1966)

PavelBal:
Implements the approach agreed on #1933 (proposed by @Felixmil) and tracked in #1966.

## Problem
After #1933 removed the `ggplot2::theme_update(legend.title = element_blank())` calls from the tests, the default plots show the redundant legend titles (`name` / `group`). Setting `theme_update()` on package load was rejected because it mutates the **global** ggplot2 theme for the whole session (affects every plot after `library(ospsuite)`, not just ospsuite plots).

## Change
Apply the tweak **locally on the returned ggplot object** in each ospsuite.plots plotting function, via a small internal helper:

```r
.applyOspsuitePlotTheme <- function(plotObject) {
  plotObject + ggplot2::theme(legend.title = ggplot2::element_blank())
}
```

Wrapped the final `return()` in: `plotTimeProfile`, `plotPredictedVsObserved`, `plotResidualsVsCovariate`, `plotResidualsAsHistogram`, `plotQuantileQuantilePlot`. No global state is touched.

- Regenerated the `plot-with-ospsuite-plots-*` snapshots (legend titles now gone). All 4 plot test files pass locally.
- Vignette `plotting-with-ospsuite-plots.Rmd`: added a short "Legend title" example showing how to re-enable it (`theme(legend.title = element_text())`).

## Scope
- Only the `ospsuite.plots` family (`R/plot-with-ospsuite-plots.R`). The TLF-based plotting functions are intentionally left unchanged (per discussion).
- Only `legend.title` is baked into the functions. The `legend.position` / `legend.justification` settings that were in the old tests were variant-specific, not global defaults, so they are not added here.

Closes #1966

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Style**
  * Updated plot rendering to improve legend readability by default, including suppressing legend titles for a cleaner visual appearance.

* **Documentation**
  * Enhanced the plotting vignette with a new “Legend title” section explaining the default legend-title behavior and how to restore or customize it using a ggplot2 theme layer.
  * Added a “Legend position” section with examples, including guidance for placing the legend inside the plot.

Issue #1966, opened 2026-06-15
Apply legend.title = element_blank() inside plotting functions (not via theme_update)


PavelBal:
## Background

PR #1933 removed the `ggplot2::theme_update(legend.title = ggplot2::element_blank())` (and `legend.position` / `legend.justification`) calls from the plot snapshot tests, so snapshots now reflect the package's true default behavior. Consequence: legends again show the (unnecessary) titles `name` / `group` / `name`, which looked cleaner when blanked.

Discussion on #1933 (Felix, Katrin, pchelle, Pavel) converged on **applying the look improvements inside the plotting functions** rather than via global theme state.

## Why not `theme_update()` on package load

`ggplot2::theme_update()` mutates the **global** ggplot2 theme for the whole R session. After `library(ospsuite)` it would affect *every* ggplot the user creates, not just ospsuite plots — an unwanted side effect.

## Proposed approach (Felix)

Apply the theme tweak **locally on the returned ggplot object** in each ospsuite plotting function. Since the functions already build a ggplot and add layers via `+`, it is a one-liner before returning:

```r
# before returning, e.g. in plotTimeProfile():
plotObject <- plotObject + ggplot2::theme(legend.title = ggplot2::element_blank())
return(plotObject)
```

This keeps the effect local to the returned plot and leaks no global state.

## Action plan

1. In each ospsuite plotting function in `R/plot-with-ospsuite-plots.R` (`plotTimeProfile`, `plotPredictedVsObserved`, `plotResidualsVsCovariate`, `plotResidualsAsHistogram`, `plotQuantileQuantilePlot`), append the `ggplot2::theme(legend.title = ggplot2::element_blank())` layer before returning the plot object. (Best via a small internal helper to stay DRY.)
2. Remove remaining `theme_update` calls from the test files. Any genuinely test-specific theme settings (e.g. `legend.position` / `legend.justification` for specific snapshot variants) → wrap in `withr::local_theme()` so they don't leak global state between tests.
3. Regenerate affected snapshots so they reflect the new function defaults.
4. Add a short vignette example showing how to switch the legend title on/off.

## Scope notes

- `legend.title = element_blank()` is the agreed default. `legend.position`/`legend.justification` in the tests were variant-specific, not global defaults — keep those test-local (`withr::local_theme()`), don't bake into the functions.
[...]

PR #1977, opened 2026-06-22
Stop mutating global ggplot2 state on package load

### LINKED ISSUE #1968: library(ospsuite) mutates the global ggplot2 theme and geom defaults on load
## Summary

Loading the package mutates global `ggplot2` state. In `R/zzz.R`, `.onLoad()` calls `ospsuite.plots::setDefaults()`, which internally runs `ggplot2::theme_set()` and several `ggplot2::update_geom_defaults()` calls. As a result, simply running `library(ospsuite)` changes the default theme and default geom aesthetics for the **entire** R session, affecting every `ggplot2` plot the user builds afterward, including plots that have nothing to do with ospsuite.

This is the same class of concern that led us to reject `ggplot2::theme_update()` on load in #1933 and to apply legend tweaks locally in the returned plot object in #1966 / #1967. The on-load global mutation predates those and is broader (theme plus geom defaults plus options).

## How to reproduce

```r
library(ggplot2)
theme_get()$panel.background$fill
#> "#EBEBEBFF"   (grey, the theme_grey default)

library(ospsuite)
theme_get()$panel.background$fill
#> "white"       (theme_bw based ospsuite.plots theme, now the session default)

# a completely unrelated plot is now restyled
ggplot(mtcars, aes(wt, mpg)) + geom_point()
#> white panel, no minor grid, legend on the right, shape = 1 open points, etc.
```

## Where

* `R/zzz.R`: `.onLoad()` calls `ospsuite.plots::setDefaults()`, which calls `ggplot2::theme_set()` and `ggplot2::update_geom_defaults()` (global side effects).

## Why this is a problem

The [ggplot2 "ggplot2 in packages" guidance](https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html#creating-a-new-theme) recommends that a package define its theme as a **function** that is applied to the relevant plots (`plot + theme_custom()`), calculated at call time, rather than mutating the global theme. Mutating `theme_set()` on load:

1. Surprises users: `library(ospsuite)` silently restyles all of their plots, not only ospsuite plots.
2. Is order dependent: the final appearance depends on attach order versus any theme the user set themselves.
3. Is hard to scope or undo: nothing restores the previous state on unload (no `.onUnload`), even though `setDefaults()` returns the old state.

## Proposed direction

* Stop calling `setDefaults()` (the global-mutating path) from `.onLoad()`.
* Apply the ospsuite styling locally to the plot objects returned by the ospsuite plotting functions, via a theme function calculated at call time, as recommended by the ggplot2 docs. This depends on OSPSuite.Plots exposing a non-mutating way to obtain its theme and geom defaults (tracked separately in the companion OSPSuite.Plots issue).

This pairs with a companion issue in OSPSuite.Plots: https://github.com/Open-Systems-Pharmacology/OSPSuite.Plots/issues/130

Felixmil:
Loading the package used to mutate global `ggplot2` state for the whole R session: `.onLoad()` called `ospsuite.plots::setDefaults()`, which runs `theme_set()`, several `update_geom_defaults()` calls, and `options()` changes. As a result, `library(ospsuite)` silently restyled every plot a user built afterward, even plots unrelated to ospsuite. This is the same class of session contamination rejected in #1933 and worked around locally in #1966 / #1967. This change removes the on-load mutation and relies on the per-plot styling now shipped by `{ospsuite.plots}`.

## On-load behavior

`.onLoad()` no longer calls `setDefaults()`. The `{ospsuite.plots}`-based plotting functions (`plotTimeProfile()`, `plotPredictedVsObserved()`, `plotResidualsVsCovariate()`, `plotResidualsAsHistogram()`, `plotQuantileQuantilePlot()`) keep their full OSP styling because the current `{ospsuite.plots}` styles each plot at build time (a `theme_osp()` plus an `osp_ggplot` class that applies the OSP colour/fill/shape scales), with no global side effects. After `library(ospsuite)`, the user's global theme and geom defaults are left exactly as they were.

## Dependency and lockfile

The per-plot styling requires the newer `{ospsuite.plots}`, so the `DESCRIPTION` floor is raised to `(>= 1.2.0.9003)` and `renv.lock` is synced to that build.

## Why the `{tlf}`-based plots are also affected

This is the non-obvious part of the change. The deprecated `{tlf}`-based functions (`plotIndividualTimeProfile()`, `plotPopulationTimeProfile()`, `plotObservedVsSimulated()`, `plotResidualsVsTime()`, `plotResidualsVsSimulated()`) were never styled per plot. They render against whatever the global `ggplot2` theme is at draw time, and they were silently relying on the theme that `setDefaults()` installed on load. Once that on-load mutation is gone, those plots no longer inherit it, so their appearance changes.

I deliberately did not re-apply a theme to these plots, for two reasons:

1. It is not faithfully reproducible. `{tlf}` resolves its theme from the global theme at print time, so there is no way to restore the exact previous look without re-introducing the global mutation this PR removes.
2. More importantly, appending a theme after `{tlf}` builds the plot would override whatever the user configured through `DefaultPlotConfiguration` (legend position, title alignment, grid, fonts, colours). The bulk of the OSP look for these plots (white panel, no grid, left-aligned titles) already comes from `{tlf}`'s own plot configuration, not from the global theme, so the practical visual change is small and user `DefaultPlotConfiguration` settings are now fully respected.

These functions are soft-deprecated and scheduled for removal in version 14.0.

### The visible change is a single panel border, and removing it restores the original look

The one visible difference in the `{tlf}` snapshots is the **panel border** (the box outline around the plotting area). Tracing the history of one snapshot, `_snaps/.../plot-observed-vs-simulated/defaults.svg`, shows the border was never part of the original design; it appeared only during the `setDefaults()` era and this PR removes it:

| State | Commit | Panel | Snapshot |
|---|---|---|---|
| 1. Original `plotObservedVsSimulated()` | [`fd9f0198`](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/fd9f0198/tests/testthat/_snaps/plot-observed-vs-simulated/default-plot.svg) (#945) | No border (open L-shaped axes) | [view](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/fd9f0198/tests/testthat/_snaps/plot-observed-vs-simulated/default-plot.svg) |
| 2. Just before #1911 | [`bf06d879`](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/bf06d879/tests/testthat/_snaps/plot-observed-vs-simulated/defaults.svg) | No border (still L-shaped) | [view](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/bf06d879/tests/testthat/_snaps/plot-observed-vs-simulated/defaults.svg) |
| 3. #1911 "Do not set settings for tlf based tests" | [`da542418`](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/da542418/tests/testthat/_snaps/plot-observed-vs-simulated/defaults.svg) | **Full box border appears** | [view](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/da542418/tests/testthat/_snaps/plot-observed-vs-simulated/defaults.svg) |
| 4. This PR | [`d24c6840`](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/d24c6840/tests/testthat/_snaps/1-plot-observed-vs-simulated/defaults.svg) | No border (back to L-shaped) | [view](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/d24c6840/tests/testthat/_snaps/1-plot-observed-vs-simulated/defaults.svg) |

What happened in between: when `ospsuite.plots::setDefaults()` was added to `.onLoad()`, it set the global theme to `theme_osp()`, which is `theme_bw()`-based and therefore carries a `panel.border`. The `{tlf}` tests originally compensated for this with a local workaround, `ggplot2::theme_update(panel.border = ggplot2::element_blank())`, so the snapshots stayed borderless. PR #1911 removed that workaround, and from that point the snapshots rendered **with** the inherited border (state 3). I confirmed via `git log --follow -p` that the bordering rect (`<rect ... stroke: #333333>`) is added in exactly one commit (#1911) and removed in exactly one (this PR).

So the `{tlf}` snapshot changes in this PR are not a regression: they revert an artifact of the global on-load mutation and restore the plots to their **original, pre-`{ospsuite.plots}` appearance** (states 1 and 2 match state 4).

## Documentation

The vignettes no longer instruct users to call `setDefaults()` or to restore `ggplotDefaults`; plots are styled automatically and the only optional setup is the watermark option.

Closes #1968

## Summary by CodeRabbit

* **Chores**
  * Updated `ospsuite.plots` dependency to version 1.2.0.9003 or higher

* **Documentation**
  * Added release notes noting that loading ospsuite no longer modifies the session's global ggplot2 theme; plots are now styled individually


pchelle (approved):
If my understanding is correct, this may break some snapshot tests down the road on packages that were using ospsuite-r plots

Felixmil:
Yes, just like [da542418](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/da542418/tests/testthat/_snaps/plot-observed-vs-simulated/defaults.svg)

Yuri05 (approved):

PR #1982, opened 2026-07-01
Treat empty sheets vector as NULL in loadDataSetsFromExcel()

PavelBal:
## Summary

An empty character vector (`character(0)`) passed to the `sheets` argument of `loadDataSetsFromExcel()` was non-`NULL`, so it took the override branch and set `importerConfiguration$sheets <- character(0)` — importing **no** sheets instead of falling back to the configuration/all-sheets logic.

This PR normalizes `character(0)` to `NULL` before the branch logic, so an empty vector behaves the same as `sheets = NULL`.

## Changes

- **`R/utilities-data-set.R`** — normalize an empty `sheets` vector to `NULL` before the sheet-selection branch logic.
- **Roxygen docs** — document that `character(0)` is treated the same as `NULL`.
- **`tests/testthat/test-utilities-data-set.R`** — regression test verifying an empty vector falls back to the configuration sheet.
- **`NEWS.md`** — added a bullet under the development version.

## Decision-tree context

The `sheets` / `importAllSheets` priority chain is `importAllSheets` > `sheets` > configuration. Previously an empty vector incorrectly reached the "import listed sheets" branch (with an empty list); it now joins the `sheets = NULL` path (configuration sheets, or all sheets when the configuration defines none).

## Testing

`test-utilities-data-set.R` passes locally via `pkgload::load_all()` + `testthat` (the single skip is the pre-existing "On CRAN" deprecation snapshot).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spreadsheet import behavior so providing an empty sheet list is treated the same as leaving sheets unspecified.
  * Ensures imports fall back to the configured sheets (or all sheets) instead of returning no data.
* **Tests**
  * Added coverage verifying results are non-empty and dataset names match when using `sheets = character(0)` vs `sheets = NULL`.


PavelBal:
Thanks @coderabbitai. Addressed the nitpick in 10d63d1b.

Rather than `expect_equal(importConfig$sheets, "TestSheet_1")` — which would pass on the *buggy* code too, since the old override branch restored `importConfig$sheets` afterward and so never distinguished the two paths — the test now asserts the `character(0)` result is **identical to the `sheets = NULL` result** (`expect_equal(names(fromEmpty), names(fromNull))`). This directly validates the documented equivalence and fails on the pre-fix behavior.

@coderabbitai review

Yuri05 (dismissed):

nautilus69 (approved):

PR #1983, opened 2026-07-02
Ship Enums.NET.dll (fixes loadDataSetsFromExcel on v13)

PavelBal:
## Problem

On the `v13` branch, `loadDataSetsFromExcel()` fails for Excel workbooks that
contain a sort/filter (an `AutoFilter` / `SortState`) on any sheet, throwing:

```
OSPSuite.Infrastructure.Import.Core.InvalidObservedDataFileException
Message: An error occurred while reading the file. Please check the content.
```

The same file loads fine on `main` (v12).

## Root cause

The real (swallowed) exception is:

```
System.IO.FileNotFoundException: Could not load file or assembly
'Enums.NET, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7ea1c1650d506225'
   at NPOI.OpenXmlFormats.Spreadsheet.CT_SortCondition.Parse(...)
```

- The bundled Excel reader **NPOI 2.8.0** references **Enums.NET 5.0.0.0** and needs
  it at runtime to parse `SortState`/`SortCondition` elements.
- `Enums.NET` was **not** on the DependencyManager allow-list
  (`@(AllowedAssembly)` in `DependencyManager.csproj`), so the PostBuild sweep
  **deleted** `inst/lib/Enums.NET.dll` after copying it (the allow-list is
  intentional per #1587, but NPOI genuinely needs this DLL).
- `OSPSuite.Core`'s `ExcelDataSourceFile.DoLoadWork` catches the
  `FileNotFoundException` and rethrows the generic `InvalidObservedDataFileException`
  (logging the real message only to a logger that is silent in the R/CLI setup),
  which is why the cause was hidden.

`main` still ships `Enums.NET.dll` (its allow-list / older NPOI included it), so v12
is unaffected.

## Fix

- Add `Enums.NET` to `@(AllowedAssembly)` in `DependencyManager.csproj` so the
  PostBuild sweep keeps it.
- Add `inst/lib/Enums.NET.dll` — the **5.0.0** package, **net7.0** build
  (AssemblyVersion 5.0.0.0), which is what a `net10.0` DependencyManager build
  resolves for NPOI's `Enums.NET >= 5.0.0` dependency.

## Verification

With both changes, the previously-failing workbook loads on v13:

```
Loaded 37 data set(s).
```

(Reproduced via `devtools::load_all()` on the `v13` tree with renv active, i.e. using
this branch's `inst/lib` and rSharp `V13` build.)

## Note

`.github/workflows/build-libraries.yaml` regenerates and commits `inst/lib` when
`DependencyManager.csproj` changes, but it is a `workflow_call` reusable workflow and
is not currently wired into a triggering workflow — so the DLL is committed here
directly. If/when that workflow runs, it will reproduce this same `Enums.NET.dll`.


rwmcintosh (approved):

rwmcintosh:
I would add a test to make sure there's not going to be a regression

rwmcintosh (changes requested):
I'd add a test to make sure no regression is possible

PR #1984, opened 2026-07-02
Add CLAUDE.md contributor/agent guide

PavelBal:
Adds a `CLAUDE.md` guide describing the R ↔ .NET (rSharp) architecture and the practical workflow for working across the `main` (v12) and `v13` release lines.

## What it documents
- **Where the .NET implementation lives** — the OSPSuite.Core / MoBi / PK-Sim repos, and the branch mapping (this package's `main` = Core `develop`/v12, `v13` = Core `V13`). Read the matching branch when tracing what an R method does.
- **How `inst/lib` is populated on each branch** — v13's DependencyManager + allow-list vs. main's `update_core_files.R` (downloads PK-Sim build artifacts).
- **Branch ↔ dependency coupling** — per-branch rSharp builds and `inst/lib` DLLs; why switching branches in place requires `renv::restore()` + an R restart (loaded DLLs are locked); and the recommended git-worktree-per-branch workflow.
- **Running gotchas** — avoid `Rscript --vanilla` (skips renv/`.Rprofile`), `load_all()` is the true per-branch test, and how to surface wrapped/truncated .NET exceptions from rSharp.

Docs only — no code or dependency changes.

## Summary by CodeRabbit

* **Documentation**
  * Added a contributor and agent guide with setup, debugging, branch-switching, and troubleshooting tips.
  * Simplified the main guidance note to point to the consolidated contributor instructions.

* **Chores**
  * Updated package build settings to exclude additional top-level documentation files from build artifacts.


CLAUDE.md
```diff
@@ -0,0 +1,119 @@
...
+
+```bash
+git worktree add ../OSPSuite-R-main main
+git worktree add ../OSPSuite-R-v13  v13
+# then in each: R -> renv::restore()
+```
+
+## Running things
+
+- The system R here is `C:\Program Files\R\R-4.6.0\bin\Rscript.exe` (not on PATH).
```
Felixmil: machine dependant

CLAUDE.md
```diff
@@ -0,0 +1,119 @@
...
+
+Run `renv::restore()` once in each worktree. After that you can keep an R session
+open per worktree and compare `main` vs `v13` behavior without any restore/restart
+thrash. Example:
+
+```bash
+git worktree add ../OSPSuite-R-main main
+git worktree add ../OSPSuite-R-v13  v13
+# then in each: R -> renv::restore()
+```
```
Felixmil: I agree with this and this is what I use but it feels more like a user preference than project specifics.
renv handling for worktrees is a good addition but I'm afraid what is proposed here is not the best way to handle it.

CLAUDE.md
```diff
@@ -0,0 +1,119 @@
...
+repositories:
+
+- **OSPSuite.Core** — https://github.com/Open-Systems-Pharmacology/OSPSuite.Core
+  (the core domain/import/simulation logic, e.g. `ExcelDataSourceFile`, `DataImporter`).
+- **MoBi** — https://github.com/Open-Systems-Pharmacology/MoBi
+- **PK-Sim** — https://github.com/Open-Systems-Pharmacology/PK-Sim
+  (`PKSim.R` is the entry assembly the R package initializes).
+
+**Branch mapping matters:** this R package's `main` ships Core **v12** = the Core
+repo's `develop` branch; this package's `v13` ships Core **v13** = the Core repo's
```
Felixmil: `develop` branch ? I think it has been decomissioned for long time now, better not even mention it.

CLAUDE.md
```diff
@@ -0,0 +1,119 @@
...
+# then in each: R -> renv::restore()
+```
+
+## Running things
+
+- The system R here is `C:\Program Files\R\R-4.6.0\bin\Rscript.exe` (not on PATH).
+- **Do NOT run with `Rscript --vanilla`** — `--vanilla` skips `.Rprofile`, so
+  `renv/activate.R` never runs and you silently fall back to the *global* user
+  library (wrong rSharp / missing dev deps). Run without `--vanilla` so renv activates.
+- `devtools::load_all()` loads the **current source + this tree's `inst/lib` DLLs**
```
Felixmil: regroup all renv specific entries in a renv dedicated section ?

Yuri05:
@PavelBal Please add claude.md to **.Rbuildignore**.

Felixmil (changes requested):
Mixes user/machine preferences with actual project context

Yuri05:
@coderabbitai full review

CLAUDE.md
PavelBal: No, not in Core/PKSim/MoBi.

CLAUDE.md
PavelBal: Hmmm what would be the best way?

The problem is that when switching between v13 and v12 branches, the R session must be reloaded, otherwise the DLLs are not updated (because they are still loaded). Without the worktrees, Claude might switch in between without knowing it (or the user does).

Also, v12 and v13 branches required different versions of rSharp.

Yuri05:
<details><summary><b>PR review by Claude Fable :)</b></summary>

I've read the full diff (119-line new CLAUDE.md) plus the existing review comments. Here's my assessment.

**What's good.** The core content is genuinely valuable and non-obvious: the branch↔dependency coupling table (same rSharp version string `1.2.0.9000` on both branches but different git SHAs), the DLL-locking explanation for why an R restart is needed before `renv::restore()`, the v13 DependencyManager allow-list mechanism (with the pointer to issue #1587), and the .NET exception truncation/rewrapping gotcha with the `rSharp::newObjectFromName()` workaround. This is exactly the tribal knowledge that costs an agent (or new contributor) hours to rediscover. Docs-only, no code risk.

**Must fix before merge:**
1. Add CLAUDE.md to .Rbuildignore [github](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1984) (Yuri05's point) — otherwise `R CMD check` flags it as a non-standard top-level file.
2. The hardcoded path `C:\Program Files\R\R-4.6.0\bin\Rscript.exe` (not on PATH) [github](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1984/files/e3c408585b8a2e5d920778e374ba68f9098ecded) is machine-specific, as Felixmil noted. Replace with something environment-neutral ("locate Rscript; it may not be on PATH on Windows") or drop it.
3. Verifiability-in-time claims like "verified: two hash dirs exist under `.../cache/.../rSharp/...`" are a snapshot of one machine's state, not a project fact. Rephrase as behavior ("renv keys cache entries by content hash, so both builds coexist").

**Should fix:**
4. The `develop` branch question: Felixmil says it's decommissioned, Pavel says not in Core/PKSim/MoBi. Whoever is right, this is exactly the kind of statement that rots — I'd phrase it as "main ships Core v12; check renv.lock / the Core repo for the current branch mapping" rather than hardcoding branch names twice.
5. Branch drift: this file lands on `main` but half of it documents v13 mechanics. Either the v13 branch needs its own copy kept in sync, or the file should clearly state it describes both lines and which branch is authoritative. Otherwise it will silently diverge.
6. Felixmil's structural suggestion is sound: consolidate the renv material (restore-after-switch, `--vanilla` warning, dev-dep check in `load_all()`) into one section — it's currently scattered across three.
7. The worktree section reads as a personal workflow recommendation. I'd keep the *facts* (per-worktree project library keyed by path hash, shared cache separates the rSharp builds) but soften "Preferred workflow" to "One option: ...", which also addresses Felixmil's objection.

**Minor:** the `workflow_call` caveat about build-libraries.yaml ("must be wired into a triggering workflow") is useful today but likely to go stale — consider linking the workflow file instead of describing its wiring.

**Verdict:** worth merging after revision. The requested changes are all about separating durable project facts from machine/session/preference details — the underlying technical content is accurate and high-value. If the team also uses other agents (Copilot, CodeRabbit is already active), naming it `AGENTS.md` with a `CLAUDE.md` symlink or one-line pointer would make it tool-neutral.
</details>

CLAUDE.md
Yuri05: I think I would create 2 slightly different versions of Claude.md in **main** and in **V13**; in each branch the claude.md should only be specific for this branch and do not even mention things in another branch(es)

PavelBal:
@Yuri05 @Felixmil addressed

Yuri05 (dismissed):

CLAUDE.md
```diff
@@ -1,119 +1 @@
...
-  cut off before any InnerException detail.
-- Several Core readers **catch the real exception and rethrow a generic one**
-  (e.g. `ExcelDataSourceFile.DoLoadWork` catches everything and throws
-  `InvalidObservedDataFileException("An error occurred while reading the file…")`,
-  logging the real message only to a logger that is silent in the R/CLI setup).
-
-To get the real cause, **replicate the failing .NET call path at a low level from R**
-via `rSharp::newObjectFromName()` / `$call()` / `$get()`, wrapping each step in
-`tryCatch`, so the raw unwrapped exception surfaces at the exact failing operation.
+See [AGENTS.md](AGENTS.md) for agent & contributor guidance for this repository.
```
Felixmil: should use `@` for automatic context injection.

cf: https://code.claude.com/docs/en/memory#import-additional-files

Felixmil: otherwise it's not 100% sure it's loaded into context.

Felixmil (changes requested):

Felixmil (approved):

PR #1985, opened 2026-07-03
Add CLAUDE.md contributor/agent guide (v13)

PavelBal:
Adds a **v13-specific** `CLAUDE.md` describing the R ↔ .NET (rSharp) architecture and the workflow for this line. Companion to the main-branch guide (#1984); per review feedback (Yuri05), each branch's `CLAUDE.md` is scoped to that branch and does not describe other branches.

## What it documents
- **Where the .NET implementation lives** — OSPSuite.Core / MoBi / PK-Sim, on their `V13`/`v13` branches — and to read the matching branch when tracing what an R method does.
- **How `inst/lib` is populated on v13** — the `shared/DependencyManager/` project's allow-list (`@(AllowedAssembly)`): DLLs not on the list are deleted at build time (issue #1587). Documents how to ship a new DLL, using the recent `Enums.NET` fix as the concrete example.
- **renv / branch switching / worktrees** — why switching branches needs `renv::restore()` + an R restart (loaded DLLs are locked), and the git-worktree-per-branch workflow.
- **Running & debugging** — `load_all()` as the true per-branch test, and how to surface wrapped/truncated .NET exceptions from rSharp.

Docs only — no code or dependency changes.

## Summary by CodeRabbit

* **Documentation**
  * Updated contributor/agent guidance to follow **v13** conventions, including worktree setup and updated instructions for keeping required packaged .NET binaries available at runtime (including handling of missing/transitive dependencies).

* **CI & Automation**
  * Improved library build automation to create and open PRs using an app-based token, always rebuilding libraries and committing updated outputs when changes occur.
  * Removed/emptied the previous main workflow and the core-files update workflow (no replacement included in this change set).


AGENTS.md
```diff
@@ -0,0 +1,100 @@
...
+committed into **`inst/lib/`** (OSPSuite.Core.dll, NPOI, SQLite, etc.) and are
+loaded at runtime by rSharp when the package initializes.
+
+## Where the .NET implementation lives (look here, not just in R)
+
+The R functions are thin wrappers over .NET. When you need to understand what a
+method *actually does* — or why it throws — read the .NET source in the upstream
+repositories, on the branch this line builds against:
+
+- **OSPSuite.Core** (`V13` branch) — https://github.com/Open-Systems-Pharmacology/OSPSuite.Core
```
Yuri05: here and everywhere: point directly to the V13 branch of the PK-Sim/MoBi/Core/ - repo? So here:
**OSPSuite.Core** (`V13` branch) - https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/tree/V13

AGENTS.md
```diff
@@ -0,0 +1,100 @@
...
+  (core domain/import/simulation logic, e.g. `ExcelDataSourceFile`, `DataImporter`).
+- **MoBi** (`v13` branch) — https://github.com/Open-Systems-Pharmacology/MoBi
```
Yuri05: Add that MoBi.R is the entry project (same as for PK-Sim below)

AGENTS.md
```diff
@@ -0,0 +1,100 @@
...
+
+**To ship a new DLL, add its name (without extension) to `@(AllowedAssembly)`** in
+`DependencyManager.csproj`. (This is exactly how a missing transitive dependency bites:
+NPOI needs `Enums.NET` at runtime to parse sheets with a sort/filter, but it was not
+on the allow-list, so `loadDataSetsFromExcel()` failed until `Enums.NET` was added.)
+
+`.github/workflows/build-libraries.yaml` rebuilds the solution and auto-commits the
+regenerated `inst/lib` when `DependencyManager.csproj` changes — but note it is a
+`workflow_call` reusable workflow and must be wired into a triggering workflow to run.
+If it isn't invoked on your PR, commit the DLL manually alongside the `.csproj` change.
```
Yuri05: I don't get the last sentence.
Actually is `build-libraries.yaml` called at all? From where? I could not find any calling workflow in the V13 branch, hmm

AGENTS.md
PavelBal: Done — all three links now point to their v13 branches (Core/PK-Sim `/tree/V13`, MoBi `/tree/v13`).

AGENTS.md
PavelBal: Done — added that `MoBi.R` is an entry assembly the R package initializes (`loadAssembly` + `MoBi.R.Api InitializeOnce` in init-package.R), same as PKSim.R.

AGENTS.md
PavelBal: Confirmed — nothing on this branch invokes `build-libraries.yaml`, so it doesn't run automatically. Reworded to state exactly that, and that you build the DependencyManager locally and commit `inst/lib` yourself.

Yuri05: should we than remove `build-libraries.yaml`? @rwmcintosh can you confirm that it is not used anymore?

Yuri05: > should we than remove `build-libraries.yaml`? @rwmcintosh can you confirm that it is not used anymore?

Or change the workflow trigger from `workflow_call` to `workflow_dispatch` ?

rwmcintosh: I guess we can keep the build-libraries.yaml as-is. IDK. It seems to me that if the dependencyManager.csproj does not contain an update, then the build would execute, but would not make any changes because binaries would be identical to the existing ones.

If the project contains an update then it would replace them.

If someone without dotnet SDK wanted to update the versions, then this would be helpful because they just have to commit the change to csproj.

As long as a trigger does not create an empty commit all the time and retrigger builds etc. then it's probably ok to use it on all push, or PR or whatever.

When I perform the update, I commit the updated binaries, but an R dev might not be able to do that so easily.

Yuri05: Well, then I would just convert the trigger from `workflow_call `to `workflow_dispatch` so that the workflow can be triggered manually on demand (and adjust the description in AGENTS.md accordingly)
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/7dcc017c4db5ed741065e56a015ef2342a8e41ac/.github/workflows/build-libraries.yaml#L2

rwmcintosh: yeah, could do that too. As long as it can be done by someone other than u and i

Yuri05: > yeah, could do that too. As long as it can be done by someone other than u and i

well, then @PavelBal 😉

PavelBal: @Felixmil can you comment on this? I don't have sufficient knowledge about the workflows.

Felixmil: Workflows changed significantly on main, and not all changes were aligned in v13 yet.

Also, it needs some fixes:
- The workflow cannot be triggered (nothing calls it and not dispatchable),
- Commits directly to target branch: will fail once v13 hits main (because of branch protection rules).
- update-core-files workflow is outdated and should be deleted entirely.

I will align v13 with main regarding workflow, implement the fixes and update `AGENTS.md` accordingly in a separate PR so this one is not blocked and stays tightly scoped.

Felixmil: Done here: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1990/

PavelBal: @Yuri05 Do you agree?

Yuri05: > @Yuri05 Do you agree?

yes

Yuri05 (approved):

PR #1987, opened 2026-07-04
Consolidate runSimulations to run mixed individual + population simulations

### LINKED ISSUE #1981: Consolidate runSimulations to run a mixed list of individual + population simulations
## Background

Part of resolving Open-Systems-Pharmacology/PK-Sim#3592. Once a simulation carries its own population + aging data (OSPSuite.Core upstream), `runSimulations` should accept a **heterogeneous list of individual and population simulations** and run them correctly — removing the current restriction that forbids populations on the multi-simulation path.

## Current behavior (`R/utilities-simulation.R`)
- `runSimulations` branches on the **count** of simulations: a single simulation may take a `population` (and `agingData`); multiple simulations with a non-null `population` is rejected with `errorMultipleSimulationsCannotBeUsedWithPopulation`.
- The reason is structural: `population` is one shared positional argument with no per-simulation slot, and the concurrent path (`.runSimulationsConcurrently`) adds bare simulations to `ConcurrentSimulationRunner` and never forwards a population.

## Changes
- Point the multi-simulation path at the new `SimulationRunner.RunSimulations(list, options)` (OSPSuite.Core), which partitions individuals (parallel) vs populations (sequential) and reads each simulation's own population/aging. `ConcurrentSimulationRunner`'s simulation-list role is superseded; its `SimulationBatch` role is untouched.
- **Remove** the multi-population restriction.
- **Add R setters** to bind a population/aging to a `Simulation` for the manual-construction workflow (e.g. `simulation$population <- pop`, `simulation$agingData <- aging`). `loadPopulation` currently returns an unbound `Population`; there is no API today to attach it to a specific simulation.
- **Precedence / single route:** decide how the legacy positional `population`/`agingData` args (single-sim path) relate to the on-object data. Recommended: on-object is the canonical route; error on conflict; deprecate the positional args. This is also what makes the snapshot case work, since a loaded simulation can only return data *on the object*.
- Result decoding is unchanged — `SimulationResults` is already one unified type for individual and population runs, decoded by Id, so a mixed result set needs no new decoder.

## End-to-end test (the seam)
A population snapshot → `loadSimulationsFromSnapshot` → `runSimulations` on a **mixed list** → returns population results (`Count > 1`) with aging applied. This is the test that proves the whole cross-repo chain; none of the per-repo tests cover the seam.

## Blocking order
Blocked by:
- Open-Systems-Pharmacology/OSPSuite.Core#2897 (`RunSimulations` — the method this repoints to; itself blocked by the wrapper-properties issue Open-Systems-Pharmacology/OSPSuite.Core#2896)
- Open-Systems-Pharmacology/PK-Sim#3592 (snapshot populates the wrapper properties)

rwmcintosh:
Fixes #1981.

`runSimulations()` now runs a single mixed list of individual and population simulations through the stateless `SimulationRunner.RunSimulations` in OSPSuite.Core, instead of routing population runs through a separate path. Adds a snapshot seam test (`ind_and_pop_snapshot.json`) that loads a project with one individual and one population simulation and runs them together.

Bundled binaries bumped to PKSim.R 13.0.84 / MoBi.R 13.0.124 / OSPSuite.R 13.0.131, which carry the `Simulation` wrapper properties and `RunSimulations` overload this depends on.

## Summary by CodeRabbit

* **New Features**
  * `runSimulations()` can run one or multiple simulations per call, including mixed individual and population simulations.
  * Added `Simulation$population` (set to a `Population` or `NULL`) and read-only `Simulation$isPopulation` to switch and detect population mode.
* **Documentation**
  * Updated `runSimulations()` documentation for mixed scheduling and clarified that `population`/`agingData` are single-simulation conveniences.
  * Added a vignette on assigning populations to simulations.
* **Bug Fixes**
  * Improved validation and output handling for population-specific runs.
* **Tests**
  * Added end-to-end coverage for the individual-and-population snapshot seam path.


R/utilities-simulation.R
```diff
@@ -411,55 +417,44 @@ runSimulations <- function(
...
+  names(resultsIdSimulationIdMap) <- names(simulationIdSimulationMap)
 
+  .getConcurrentSimulationRunnerResults(
+    results = results,
+    resultsIdSimulationIdMap = resultsIdSimulationIdMap,
+    simulationIdSimulationMap = simulationIdSimulationMap,
+    silentMode = silentMode,
+    stopIfFails = stopIfFails
+  )
+}
```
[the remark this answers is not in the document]
rwmcintosh: that's incorrect. The signature in OSPSuite.R is 
```
ConcurrencyManagerResult<SimulationResults>[] RunSimulations(SimulationRunOptions options, params RSimulation[] simulations);
```

Yuri05 (approved):

R/utilities-simulation.R
```diff
@@ -267,11 +267,19 @@ runSimulation <- function(
...
+#' A list of simulations may mix individual and population simulations. A population simulation is
+#' one that carries a population (an `IndividualValuesCache`) on the underlying object, e.g. as
+#' produced by [loadSimulationsFromSnapshot()]; such a simulation is run as a population. Individual
+#' simulations are run in parallel and population simulations sequentially - the scheduling is
+#' handled by `OSPSuite.Core`, which is handed the full list.
+#'
+#' The `population` (and optional `agingData`) arguments are a convenience for the **single
+#' simulation** case only: they bind a population to that one simulation for this run. They cannot be
+#' used with more than one simulation (there is no per-simulation slot for them); to run several
+#' population simulations, carry the population on each simulation object instead.
```
Felixmil: It looks like an actionable advice while there are no way to do this outside of loading simulation from snapshot.

Felixmil (changes requested):
My main concern here is the API shape, I think it can confuse users.

`runSimulations()` now has two ways to run a population sim, and which one applies depends on how the simulation was created rather than anything visible at the call site:

- pkml sims (`loadSimulation`) use the `population` arg, but only for a single sim
- snapshot sims carry the population on the object, so you just pass the list

So `runSimulations(list(a, b))` silently means individual or population per element based on state the user has no documented way to inspect (the `Simulation` class exposes no `population` field or accessor, only the generic `sim$get()` escape hatch), and `runSimulations(list(a, b), population = pop)` is a hard error. Someone who learned the single-sim case will try to scale it up and hit a wall with no obvious way to the thing that does work.

The `@details` also tells the user to "carry the population on each simulation object instead", but there's no R API to do that. The only way to get such a sim is `loadSimulationsFromSnapshot()`, so the advice points at a capability we don't expose.

Two directions I'd suggest:

**A (best imo): harmonize on the object.** 
Drop `population` from `runSimulations()` and add a real setter so the object is the single source of truth:

```r
sim <- loadSimulation(path)
sim$population <- pop
runSimulations(list(sim1, sim2))   # one path, pkml or snapshot alike
```

One mechanism, scaling from 1 to N is trivial, and the doc advice becomes true. Bigger change though: needs a setter writing the `IndividualValuesCache` onto the object (doesn't exist in R yet, and I'm not sure `OSPSuite.Core` exposes it through rSharp), plus deprecating the arg.

**B (cheaper but not great imo): split by function.** 
Keep the object-carried case in `runSimulations()` and pull the single-sim binding into its own named function:

```r
runSimulations(sims)                                 # object decides mode, 1..N
runPopulationSimulation(sim, population, agingData)  # explicit, single sim only
```

No invisible mode-switch inside one function, `agingData` (meaningless outside the population case) leaves the general signature, and it's non-breaking if we keep `runSimulations(sim, population=)` as a deprecated shim.

If we don't want to restructure at all here, the minimum is the doc fix: stop framing "carry the population on the object" as a user action and just say object-carried population sims come from `loadSimulationsFromSnapshot()`, while `population =` is a single-sim run-time binding.

NEWS.md
```diff
@@ -15,6 +15,7 @@ If no original PK-Sim project or snapshot are available, the user should use the
 
 ## Minor improvements and bug fixes
 
+- A `Simulation` object now has a `population` field: assign a `Population` with `simulation$population <- myPopulation` to make it a population simulation, and read it back with `simulation$population`. Assigning `NULL` switches the simulation back to an individual simulation. The read-only `simulation$isPopulation` field reports whether a simulation will be run for a population. This is an alternative to passing `population` to `runSimulations()` (which continues to work unchanged) and makes it possible to run several population simulations in a single `runSimulations()` call. See `vignette("create-run-population")` (#1988).
```
Felixmil:
```suggestion
- A `Simulation` object now has a `population` field: assign a `Population` with `simulation$population <- myPopulation` to make it a population simulation, and read it back with `simulation$population`. Assigning `NULL` switches the simulation back to an individual simulation. The read-only `simulation$isPopulation` field reports whether a simulation will be run for a population. This is an alternative to passing `population` to `runSimulations()` (which continues to work unchanged) and makes it possible to run several population simulations in a single `runSimulations()` call. See `vignette("create-run-population")` (#1987).
```

R/simulation.R
```diff
@@ -120,6 +120,36 @@ Simulation <- R6::R6Class(
...
+          return(NULL)
+        }
+        return(Population$new(netPopulation))
+      }
+      # Not using validateIsOfType() here: it inspects the calling frame to build
+      # its message, which fails when called directly from an R6 active binding.
+      if (!is.null(value) && !isOfType(value, "Population")) {
+        stop(messages$errorWrongType("population", class(value)[1], "Population"))
+      }
+      self$set("IndividualValuesCache", value)
```
Felixmil, line 155: Should we also clear Aging Data here ?

PavelBal: Good catch — yes. Done in 83814e2a.

My reasoning: aging data isn't standalone; it describes the aging trajectories of the individuals in a *specific* population. Once the population is removed or replaced, the previously attached aging data is orphaned/mismatched. Since a population run reads `sim.AgingData` directly, leaving it in place means stale aging data could be silently applied to a different population — or linger on a simulation that is now individual. And because there is no independent public setter for aging data, coupling its lifecycle to the population keeps the object in a consistent state.

So the `population` setter now clears any aging data whenever a population is assigned, replaced, or set to `NULL`. Added tests for both the remove-and-clear and the reassign-and-clear cases. (The set-then-restore path in `runSimulations()` is unaffected: it sets/restores population and aging data via the raw .NET `$set` rather than this active binding.)

Felixmil (approved):

PR #1989, opened 2026-07-06
Expose population getter/setter and isPopulation on Simulation (#1988)

### LINKED ISSUE #1988: Expose isPopulation and a population getter/setter on Simulation; consolidate runSimulations population handling
## Motivation

Since v13, a `Simulation` object can carry a population (an `IndividualValuesCache`) on the underlying .NET object. `runSimulations()` already dispatches on this: simulations that carry a population are run as population simulations, the rest as individual simulations, all through the single `SimulationRunner.RunSimulations` entry point (see #1981 / PR #1987).

However, from R there is currently **no way to inspect or change** whether a simulation carries a population. The only way to run a population simulation is to pass a `population` (and optional `agingData`) through the `runSimulations(..., population = ...)` convenience argument, which:

- is restricted to a **single** simulation (the argument has no per-simulation slot), so several population simulations cannot be run in one call; and
- routes through a separate internal code path (`.runSimulationWithPopulation` → `SimulationRunner.Run(SimulationRunArgs)`) instead of the consolidated `RunSimulations` path.

The .NET wrapper (`OSPSuite.R.Domain.Simulation`) already exposes everything needed for a cleaner design — `IsPopulation` (computed), and a settable `IndividualValuesCache` and `AgingData`.

## What we want to achieve

1. **Expose `isPopulation`** on the R `Simulation` object (read-only) — tells the user whether the simulation will run as a population simulation.
2. **Expose a `population` getter/setter** on the R `Simulation` object:
   - getter returns the assigned `Population` (or `NULL`);
   - setter assigns a `Population`, turning the simulation into a population simulation;
   - **assigning `NULL` switches the simulation back to an individual simulation.**
3. **Consolidate the run path.** The `population`/`agingData` convenience arguments of `runSimulations()` are kept for backward compatibility (no breaking change), but implemented via **set-then-restore**: the population/aging data is applied to the simulation, the run is performed through the unified `RunSimulations` path, and the simulation's original state is restored afterwards — so the caller's object is left unchanged and there is a single run code path. `.runSimulationWithPopulation` becomes obsolete and is removed.
4. As a direct consequence, **several population simulations can be run in one `runSimulations()` call** by assigning a population to each simulation beforehand.

This is offered as an **alternative** to the existing `population = ` argument, not a replacement.

## Notes / scope

- Aging data continues to be supplied through the `runSimulations(..., agingData = ...)` argument (applied via the same set-then-restore mechanism). Only `population` gets a public getter/setter, since aging data is meaningful only together with a population.
- Fixes a latent bug: `runSimulations()` referenced an undefined message (`messages$errorMultipleSimulationsCannotBeUsedWithPopulation`) when the `population` argument was combined with multiple simulations, producing an empty error. The message is now defined.
- Documentation: roxygen for the new fields (written for modelers), a new subsection in `vignette("create-run-population")`, and a `NEWS.md` entry.
- Tests cover the new fields, switching back to individual, running multiple population simulations in one call, and that the `population`/`agingData` arguments leave the simulation object unchanged (set-then-restore).

Follow-up to #1981 / PR #1987; PR targets the `1981-consolidate-runsimulations` branch.

PavelBal:
## What & why

Follow-up to the run-consolidation work on this branch (#1981 / #1987). It exposes the population that a `Simulation` already carries on the .NET side and unifies how `runSimulations()` handles the `population` argument.

Closes #1988. Targets `1981-consolidate-runsimulations`.

## Changes

**New `Simulation` fields** (`R/simulation.R`)
- `simulation$population` — get/set. Assigning a `Population` turns the simulation into a population simulation; reading returns the assigned `Population` (or `NULL`).
- `simulation$population <- NULL` — **switches the simulation back to an individual simulation**.
- `simulation$isPopulation` — read-only; whether the simulation will run as a population simulation.

**Consolidated `runSimulations()`** (`R/utilities-simulation.R`)
- The `population`/`agingData` convenience arguments are **kept for backward compatibility** (no breaking change), but are now implemented via **set-then-restore**: the population/aging data is applied to the simulation, the run goes through the same `.runSimulations()` path as every other run, and the simulation's original state is restored afterwards (via `on.exit`), so the caller's object is left unchanged.
- The separate `.runSimulationWithPopulation()` / `SimulationRunner.Run(SimulationRunArgs)` path is removed — all runs now dispatch through the single `RunSimulations` entry point.
- As a result, **several population simulations can now be run in one call** by assigning a population to each simulation beforehand.
- When a `population` argument is combined with an already-assigned population, the argument takes precedence **for that run only**.

## Design decision

For "single source of truth vs. mutation" we chose **set-then-restore** (option b) over deprecating the `population` argument, to avoid a large breaking change. The convenience argument keeps working exactly as before and never mutates the simulation object passed in — verified by test.

## Behaviour note

A single failing population simulation now surfaces via the standard `runSimulations()` warning/`stopIfFails`/`silentMode` handling (returning `NULL` for that entry) instead of throwing, consistent with individual and multi-simulation runs.

## Tests

- `Simulation` fields: individual by default; `isPopulation` read-only; assigning a population flips `isPopulation`; assigning `NULL` switches back; type validation.
- `runSimulations()`: runs a population when one is assigned to the simulation; runs several population simulations in one call; the `population` argument leaves the simulation unchanged (set-then-restore); the argument overrides an assigned population for the run only, restoring it afterwards.

All new tests pass locally via `devtools::load_all()` + `testthat` (0 failures/errors).

## Docs

- Roxygen for the new fields and updated `runSimulations()` details (written for modelers).
- New "Assigning a population to a simulation" section in `vignette("create-run-population")`, explicitly framed as an **alternative** to the existing `population` argument.
- `NEWS.md` entry.

## Summary by CodeRabbit

* **New Features**
  * Added a way to assign a population directly to a simulation and check whether it will run as a population simulation.
  * Documented two supported workflows: per-simulation population assignment and a one-off population override for a single run.
  * Added guidance for running multiple population simulations in one call.

* **Bug Fixes**
  * Improved population override handling so the simulation’s original state is restored after the run.
  * Updated validation so invalid population values are rejected with a clear error.


PavelBal:
@coderabbitai review

Yuri05 (approved):

PR #2002, opened 2026-07-29
1241 dataset order

pchelle:
## Summary by CodeRabbit

- **Bug Fixes**
  - Plot legends now follow dataset and group insertion order instead of alphabetical ordering.
  - Observed and predicted legends remain consistent.
  - Empty dataset levels no longer create incorrect legend entries.
  - Converted data now preserves dataset order, with dataset and group values represented consistently for plotting.

- **Documentation**
  - Clarified dataset and group ordering behavior in data conversion documentation and release notes.


R/data-combined.R
```diff
@@ -466,7 +467,26 @@ DataCombined <- R6::R6Class(
...
 
       # Apply data transformations
-      private$.dataTransform(private$.dataCombined)
+      data <- private$.dataTransform(private$.dataCombined)
+      
+      # Encode `name` as a factor whose levels follow the order in which
+      # datasets were added (i.e. `self$names`, which preserves insertion
+      # order). This guarantees that downstream ggplot2 legends display
+      # entries in the order datasets were added, rather than the default
+      # alphabetical ordering applied to character columns.
```
Felixmil: these comments are too verbose compared to the simple operations that are done here.

R/plot-individual-time-profile.R
```diff
@@ -170,7 +175,7 @@ plotIndividualTimeProfile <- function(
   )
 
   tlf::setDefaultErrorbarCapSize(defaultPlotConfiguration$errorbarsCapSize)
-
+
```
Felixmil, line 178:
```suggestion

```

Felixmil (commented):
Only minor comments (comments are too verbose or not really necessary imo).
Plots look good but the factor level swap will have a lot of side effects on depending packages, everyone should be aware of this

pchelle:
> Only minor comments (comments are too verbose or not really necessary imo). Plots look good but the factor level swap will have a lot of side effects on depending packages, everyone should be aware of this

I agree, this will affect a lot of plots downstream.
Alternative solution would be to create a global settings to define if legend should be ordered by alphabetical order (current) or dataset addition (PR).
@PavelBal thoughts ?

PavelBal:
I don't see a need for a setting. The new behavior is the "correct" one, downstream packages have to adapt.

PavelBal (approved):

PR #2008, opened 2026-08-13
Avoid eval() when building the y2axis aes mapping

### LINKED ISSUE #1976: R Magrittr Pipe Chain Leading to eval Code Injection
CWE-95

### Description
This rule detects cases where user-supplied input is passed through a pipe chain and ultimately evaluated by R's `eval()` function. When untrusted data reaches `eval()`, an attacker can craft malicious strings that get executed as arbitrary R code within the application's runtime environment.

This presents a critical security risk because `eval()` executes any valid R expression, allowing attackers to read or modify files, exfiltrate sensitive data, execute system commands, or completely compromise the system.

### Exploitability (AI-assessed): Non-Exploitable
The `eval()` call uses `bquote()` with a `.()` substitution operator, which safely interpolates the `y2UnitValue` variable into the expression before evaluation. `y2UnitValue` is extracted from an internal `metaData` structure (not user input), and `bquote()` prevents arbitrary code injection by only substituting the specified variable. This is a safe pattern for constructing ggplot2 aesthetic mappings and does not represent an exploitable code injection vulnerability.

### Remediation (if hardening anyway)
Avoid `eval()` on constructed expressions where practical. Prefer non-evaluating ggplot2 mechanisms (e.g. `aes(.data[[col]])` / tidy-eval) over building expressions for `eval()`. If `eval()` is required, keep substituting only known internal values via `bquote(.())` — never raw strings derived from external input.

PavelBal:
Fixes #1976

## What

Replaces the `eval(bquote(...))` construction of the secondary-y-axis aesthetic in `.getMappingForTimeprofiles()` with rlang `!!`-injection, which `ggplot2::aes()` supports natively:

```r
# before
eval(bquote(ggplot2::aes(y2axis = yUnit == .(y2UnitValue))))
# after
ggplot2::aes(y2axis = yUnit == !!y2UnitValue)
```

## Why

The static-analysis finding in #1976 (CWE-95) flagged the `eval()` call. It was already non-exploitable — `bquote(.())` inserts the unit string as a literal value, never parsed as code — but the `eval()` is unnecessary ceremony: `!!` inlines the value into the captured expression the same way, keeping the mapping independent of the enclosing environment, without any `eval()` for scanners to flag.

## Verification

- The old and new forms produce identical quosure expressions (`yUnit == "mg/l"`) and identical results when evaluated against data.
- `testthat::test_file("tests/testthat/test-plot-with-ospsuite-plots.R")` (which covers the y2 mapping path): `FAIL 0 | WARN 0 | PASS 93`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secondary-axis plot handling to ensure data is mapped correctly when units match the secondary axis.

### DIFF
--- a/R/plot-with-ospsuite-plots.R
+++ b/R/plot-with-ospsuite-plots.R
@@ -1275,7 +1275,7 @@ plotQuantileQuantilePlot <- function(
   if (any(names(metaData) %in% "y2")) {
     y2UnitValue <- metaData[["y2"]][["unit"]]
     mapping <- structure(
-      c(mapping, eval(bquote(ggplot2::aes(y2axis = yUnit == .(y2UnitValue))))),
+      c(mapping, ggplot2::aes(y2axis = yUnit == !!y2UnitValue)),
       class = "uneval"
     )
   }


PavelBal:
@coderabbit review

KatrinCoboeken (approved):

pchelle (approved):

PR #2015, opened 2026-08-17
Fix v13 vignette review findings (follow-up to #1958)

PavelBal:
Follow-up to the #1958 vignette review. Fixes the confirmed findings and closes the gaps (13 findings + 5 gaps identified; 3 items were rejected/rescoped/dropped by maintainer review and are intentionally absent).

**create-run-population.Rmd** — document that the population-assignment route clears aging data (aging runs must use the `agingData` argument of `runSimulations()`); explain that populations are stored by reference (two simulations assigned the same object share it); stop leaking the population-mutated simulation into the global simulation cache; show the named-list input form of `runSimulations()` and explain result naming.

**snapshots.Rmd** — add a note scoping the helpers to PK-Sim snapshots (MoBi snapshots not yet supported); state the `RunForAllOutputs` default; move the article to the "Working with MoBi and PK-Sim projects" pkgdown section.

**mobi-projects.Rmd** — remove invalid `execute = FALSE` chunk options (knitr silently ignored them); use exact `BuildingBlockTypes` enum keys in prose (`SpatialStructure`, `EventGroups`, ...); fix the Thyroid/`Thyroid_Generic` module-naming drift; remove a stray quote; add the missing trailing newline.

**Code fix** — `runSimulationsFromSnapshot()` now validates `exportJSON` like the other export flags; previously `exportJSON = "TRUE"` was silently treated as `FALSE` (NEWS bullet added).

**New pinning tests** — both vignette patterns under suspicion were investigated and confirmed safe, then pinned: `createMoBiModule()` does **not** detach project-owned building blocks from their source module, and `createSimulation()` genuinely errors when the initial-conditions BB is deselected (so the vignette's error demo is legitimate). No vignette rewrites were needed for either.

**Verification** — targeted test sweep across the five affected test files: `FAIL 0 | WARN 0 | SKIP 0 | PASS 309`; all three touched vignettes knit successfully against this branch.

**Optional follow-ups noted during final review (none blocking):**
- Add `regexp =` to the two new bare `expect_error()` pins so they can't pass on unrelated errors.
- Pass `output = temp_dir` in the exportJSON expectation so a future validation regression can't write CSV debris into `tests/testthat/`.
- Consider a follow-up test for whether `removeBuildingBlock()` on a module created from shared BBs propagates to the source module (adjacent to, but outside, the F9 scope).

## Summary by CodeRabbit

* **Bug Fixes**
  * `runSimulationsFromSnapshot()` now rejects non-logical values supplied for `exportJSON` instead of silently treating them as `FALSE`.
* **Documentation**
  * Improved guidance for assigning populations, simulation aging, shared population references, and named simulation results.
  * Clarified that snapshot helpers support PK-Sim snapshots only.
  * Documented the default value of `RunForAllOutputs`.
  * Reorganized snapshot documentation under “Working with projects.”
  * Corrected and expanded MoBi project examples and terminology.


PavelBal:
@coderabbit review

vignettes/create-run-population.Rmd
```diff
@@ -136,22 +136,27 @@ sim$population <- NULL
...
 
 ```{r multiplePopulations}
 # Load two separate simulations (here from the same file, for illustration)
 sim1 <- loadSimulation(simFilePath, loadFromCache = FALSE, addToCache = FALSE)
 sim2 <- loadSimulation(simFilePath, loadFromCache = FALSE, addToCache = FALSE)
 
-# Each simulation carries its own population
+# Assign a population to each simulation. The population is stored by
+# reference: here both simulations share the same `myPopulation` object, so
+# modifying it later (e.g. with `setParameterValues()`) would affect both.
```
nautilus69: A comment tells readers to modify a Population with setParameterValues(), but the exported top-level setParameterValues(parameters, values, units) 
(in R/utilities-parameter.R) expects Parameter object, not a Population. 
A reader calling  setParameterValues(myPopulation, ...) will error or do the wrong thing. The correct by-reference mutation is the method myPopulation$setParameterValues() which is at 
(R/population.R:52)

nautilus69 (dismissed):
Approved, With a minor adjustment.

nautilus69 (approved):

